PR #1, opened 2017-07-26
Add support for phpredis

themsaid:
These changes are needed to make sure Horizon runs using phpredis as well as predis, I'm currently working on another PR to laravel 5.5 that fixes the adaptation a little more.


taylorotwell:
Do we need to update the Horizon composer.json to use 5.5 components?

PR #8, opened 2017-07-26
Add provider step to readme file

ybaruchel:
Added instructions on how to add the `Laravel\Horizon\HorizonServiceProvider::class` to app.php providers array.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,10 @@ Horizon requires Laravel 5.5, which is currently in beta, and PHP 7.1+. You may 
 After installing Horizon, publish its assets using the `vendor:publish` Artisan command:
 
     php artisan vendor:publish
+    
+After publishing vendor assets, add the ServiceProvider to the providers array in config/app.php
+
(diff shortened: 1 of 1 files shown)


alexbowers:
Not needed. This is aiming at Laravel 5.5, which has auto-discovery.

sudiptpa:
@ybaruchel you may have to explain a bit more about Laravel 5.5 will have auto-discovery but if you want to install horizon on previous version of laravel on 5.x, like that with your changes.

thecrypticace:
This may be aimed at 5.5 but the composer.json file actually says it supports 5.4:
```json
"illuminate/contracts": "~5.4",
"illuminate/queue": "~5.4",
"illuminate/support": "~5.4",
```

My guess is that's a typo. Maybe it was originally developed on 5.4 but now requires 5.5?

taylorotwell:
Indeed it isn't necessary on 5.5.

PR #12, opened 2017-07-26
Switch to laravel-mix

cretueusebiu:
Switch to laravel-mix since Elixir is no longer supported and doesn't even work on the latest Node version. Now you can also version the js/css files.
I've also removed moment's locales and used Vue runtime so the js build is smaller.


taylorotwell:
I don't know if were able to get this working with Mix? @themsaid would have more information?

themsaid:
I believe David had issues with copying the compiled files while `watch` is on, will pull the config in the PR and see if that issue is resolved.

cretueusebiu:
Why not ? Works just fine. Mix allows you to set the path to your custom manifest file.

```php
mix('css/app.css', 'vendor/horizon')
```

will generate `/vendor/horizon/css/app.css`

themsaid:
Thank you @cretueusebiu it works now :)

PR #17, opened 2017-07-26
Update deprecated composer.lock version

lucasmichot:
Update deprecated composer.lock version


barryvdh:
Shouldn't you just delete the composer.lock?

themsaid:
@barryvdh removed it since it's ignored in all other laravel packages.

PR #22, opened 2017-07-26
use ::class notation everywhere

lucasmichot:
use ::class notation everywhere


taylorotwell:
But again this is actually longer in most cases 😬

lucasmichot:
But this allows to click on the class name to open the class in any IDE... ;-)

PR #38, opened 2017-07-27
Add Mail Notifications

ThibaudDauce:
Add the missing email notification channel for people without Nexmo subscription and without Slack.

### DIFF
--- a/src/Horizon.php
+++ b/src/Horizon.php
@@ -28,6 +28,13 @@ class Horizon
     public static $smsNumber;
 
     /**
+     * The email address for notifications.
+     *
+     * @var string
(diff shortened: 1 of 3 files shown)


alexbowers:
This is something we would want when it comes to using it after 5.5 comes out.

taylorotwell:
People will complain that URL in the action isn't customizable or a named route or something. I would remove it entirely.

ThibaudDauce:
@taylorotwell I think the Horizon prefix should be customizable in the service provider or in the config but we could change that afterwards? But for now, you're right, I will remove the action button :-)

PR #40, opened 2017-07-27
Use stricter comparisons

lucasmichot:
Use stricter comparisons


taylorotwell:
Not particularly necessary in some of these situations but will merge.

PR #41, opened 2017-07-27
Register resolving method in register instead of boot

barryvdh:
Otherwise this can lead to timing issues, where another ServiceProvider resolves the binding, before the resolving callback is registered. See https://github.com/barryvdh/laravel-debugbar/issues/682 for example.
//cc @fredlahde


fredlahde:
Tried this PR, and it solves the linked issue.

themsaid:
@jbrooksuk does it solve your issue?

jbrooksuk:
It looks like it should to be honest. I've not had chance to test yet.

barryvdh:
@jbrooksuk Were you using debugbar?

PR #48, opened 2017-07-27
Monitor tags by clicking on the tag within the Recent Jobs screen

tplaner:
Creates buttons out of each of the job tags listed on the Recent Jobs screen.

Upon clicking one of these tags it will begin monitoring that tag and redirect you to the Monitoring screen.


themsaid:
This might break the style when the number of tags is a bit large since each tag is now rendered as a button.

tplaner:
This is what it looks like:

image

themsaid:
@taylorotwell what do you think?

IMO if we're going to add this at least a confirmation modal should appear to the user so that knows whats going on.

taylorotwell:
Going to hold off for now.

PR #61, opened 2017-07-29
Wrap several commands in a single pipe

themsaid:
It also fixes an issue with predis not using the set prefix with `zadd` command, it's only used when it's called from within a pipe ¯\_(ツ)_/¯


taylorotwell:
Hmm, do we know the actual cause of it not using the prefix outside of a pipe. Could that be a bug on our end?

themsaid:
Yes it could be a problem with 5.5 phpredis connection, I'm re-visiting it again as we speak.

PR #63, opened 2017-07-30
Null check while adding a new Monitoring Tag

emir:
`                if (!this.name) {
                    this.$refs.tag.focus();
                    return;
                }`

I saw that code above at AddTagModal's saveTag() method for validate user provides a **tag** but I've successfully created an empty tag in the morning. (I forgot to investigate why and deleted all keys starts with **horizon** manually on redis-cli). Empty tag cause an issue while deleting on Horizon.

Not sure am I doing right with don't using validate() method, but it is quick solution for directly check $request->tag.

### DIFF
--- a/src/Http/Controllers/MonitoringController.php
+++ b/src/Http/Controllers/MonitoringController.php
@@ -94,11 +94,17 @@ class MonitoringController extends Controller
      * Start monitoring the given tag.
      *
      * @param  \Illuminate\Http\Request  $request
-     * @return void
+     * @return \Illuminate\Http\JsonResponse
      */
     public function store(Request $request)
     {
+        if (null === $request->tag) {
+            return response()->json(['success' => false], 422);
+        }
+
         dispatch(new MonitorTag($request->tag));
+
+        return response()->json([], 201);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


themsaid:
Just make sure you have the latest compiled javascript file, I can't create an empty tag currently using chrome, firefox, or safari.

src/Http/Controllers/MonitoringController.php
```diff
@@ -94,11 +94,17 @@ protected function getJobs($jobIds, $startingAt = 0)
      * Start monitoring the given tag.
      *
      * @param  \Illuminate\Http\Request  $request
-     * @return void
+     * @return \Illuminate\Http\JsonResponse
      */
     public function store(Request $request)
     {
+        if (null === $request->tag) {
```
lucasmichot: Maybe to call `\Illuminate\Http\Request::__get` twice, why not:
```php
if ($tag = $request->get('tag')) {
    dispatch(new MonitorTag($tag));

    return response()->json([], 201);
}

return response()->json(['success' => false], 422);
```

taylorotwell:
See @themsaid's comment. How do we recreate?

PR #69, opened 2017-08-03
Added Balance Options and Tags section

davidpiesse:
Added into Readme as these were the two things I needed to search elsewhere for to get info on to get Horizon up and running.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Horizon::auth(function ($request) {
 
 ### Balance Options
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -32,6 +32,29 @@ Horizon::auth(function ($request) {
 });
 ```
 
+### Balance Options
+
+Horizon allows you to choose from three balancing options. THe default is 'simple' which splits the jobs between the processes evenly:
```
dpde: Little typo here. "THe" instead of "The"

davidpiesse: Patched! :) Thanks!

PR #80, opened 2017-08-08
Vendor Publish only Laravel Horizon

dbpolito:
❤️


browner12:
Laravel 5.5 (which is required for horizon) has an awesome new interactive console when running the `vendor:publish` command, so this isn't *as* necessary anymore.

I'm okay either way, though.

PR #81, opened 2017-08-08
Change supervisor balance from false to off

dbpolito:
I think `off` is more verbose than `false`, also i see that on the class the default balance option is `off`: https://github.com/laravel/horizon/blob/master/src/SupervisorOptions.php#L35

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,11 +24,11 @@ After publishing Horizon's assets, its primary configuration file will be locate
 
 ### Balance Options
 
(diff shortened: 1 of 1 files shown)


themsaid:
Currently any value other than 'simple' & 'auto' would indicate balancing is off, so you can use `false` or `off` or `non`, etc...

taylorotwell:
I think it's fine as false, but @themsaid is correct.

PR #92, opened 2017-08-14
[WIP] Allow users to specify a custom middleware

phroggyy:
This PR allows for users to pass a custom middleware, instead of just a closure, to `Horizon::auth()`. This allows for greater flexibility as users can use existing admin middleware instead of having to specify a new closure just for Horizon.

This is a WIP, and before making this ready for merge, I will fix any code style that is wrong, and add support for middleware parameters to be passed. Support for parameters would allow users to easily make use of Laravel's permissions system for Horizon access.

### DIFF
--- a/src/Contracts/InvalidAuthenticationMethod.php
+++ b/src/Contracts/InvalidAuthenticationMethod.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace Laravel\Horizon\Contracts;
+
+use Exception;
+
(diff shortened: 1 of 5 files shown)


taylorotwell:
My suggestion is to simply abstract your admin check out of your middleware and have the middleware and Horizon::auth() call the same service class to determine if something is an admin.

phroggyy:
@taylorotwell what about the application of using the permissions system?

PR #93, opened 2017-08-14
Ability to register horizon routes and pass middleware

themsaid:
Instead of Authenticating the user using a closure, this PR allows publishing Horizon routes using `Horizon::routes()`, then you can pass middleware.

```
Horizon::routes('auth')
```

Or

```
Horizon::routes('auth.is_admin')
```

---


ConnorVG:
This makes no sense, why not just do _exactly what *Passport* does_? It provides so much more control and keeps the "Laravel package usage" more generic.

themsaid:
@ConnorVG what control do you need? I personally believe the existing `Horizon:: auth()` is enough, I only opened this PR to allow using middleware even tho I truly believe it's not necessary.

scrwdnet:
@themsaid you mentioned you don't believe middleware is necessary, curious as to how **you** would recommend getting the logged in user inside the `Horizon::auth()` closure? Example would be lovely 😍

themsaid:
@scrwdnet 

```
Horizon::auth(function ($request) {
    return auth()->check();
});
```

This will prevent a guest from accessing Horizon.


```
Horizon::auth(function ($request) {
    return auth()->user()->email == 'jon@targaryen.xo';
});
```

This will grant access to only `Jon`

scrwdnet:
@themsaid I have my `Horizon::auth()` call in `AppServiceProvider@boot` and `auth()->user()` always returns `null`.

#GoTSpoilers

themsaid:
@scrwdnet we fixed that in https://github.com/laravel/horizon/pull/68

Make sure you have the latest code

taylorotwell:
I think Horizon::auth is fine for now.

scrwdnet:
@themsaid no composer updates, has that been released yet? I'm at v0.1.0 - shall I move this to an issue?

themsaid:
@scrwdnet we'll issue a release soon.

scrwdnet:
Thanks @themsaid

taylorotwell:
Still prefer to keep it as-is. This exposes Horizon on production by default and forces writing of your own middleware. Open to other solutions if possible.

PR #94, opened 2017-08-14
Made view param optional

robinvdvleuten:
When you make the {view} param optional, you only need one "catch-all" route.


themsaid:
thanks

PR #125, opened 2017-09-08
Ability to pause / resume the queue from the web interface

jgill88:
There are sometimes situations where I would like to easily pause the queue. It would be great if you could do that from the web interface, so I made it :smile: 

The status icon is now clickable when the queue is either running or paused, and shows a loading indicator while the queue is changing states.

I also refactored the status component to be a bit cleaner in the process.

Added tests. passing.


jgill88:
Right after I submitted the pr, I noticed that it totally broke the status component so I closed it :cry: . I will submit another one.

PR #126, opened 2017-09-08
Ability to pause / resume the queue from the web interface

jgill88:
Ok... round 2. My last pull (#125) broke the status component. 

There are sometimes situations where I would like to easily pause the queue. It would be great if you could do that from the web interface, so I made it :smile:

The queue status icon is now clickable when the queue is either running or paused, and shows a loading indicator while the queue is changing states.

Added tests. passing.


taylorotwell:
No plans to add. I don't like the keys `please-pause` etc.

PR #138, opened 2017-09-13
panel-content overflow-x from scroll to auto

jericopulvera:
Do not show scrollbar when content does not exceed the panel maximum width.

### DIFF
--- a/resources/assets/less/_components.panels.less
+++ b/resources/assets/less/_components.panels.less
@@ -21,5 +21,5 @@
 
 .panel-content {
   background: #FAFAFA;
-  overflow-x: scroll;
+  overflow-x: auto;
 }


themsaid:
It only shows the scrollbar when there's an overflow.

Please share a screenshot and mention the browser you use.

jericopulvera:
I use Google Chrome.

overflow: scroll;
scrollbar

overflow: auto;
no-scrollbar

Tested in firefox and it still has scrollbars when it is overflow: scroll;

themsaid:
That's not what I see on different screen sizes, scrollbar doesn't appear at all.

jericopulvera:
[overflow scroll vs auto](https://stackoverflow.com/questions/6689412/difference-between-html-overflow-auto-and-overflow-scroll)

Maybe it looks more consistent to just show the bars without scroll when text does not exceed the content width.

taylorotwell:
What's the status on this?

themsaid:
@taylorotwell I think it's ok to merge.

taylorotwell:
Please re-send to 1.0 branch.

PR #144, opened 2017-09-15
panel-content overflow-x from scroll to auto

jericopulvera:


KKSzymanowski:
Can the same be done for `.stats`? Currently scroll bars are always visible:
image
Changing `overflow: scroll` to `overflow: auto` hides the scroll bar when it's not necessary and shows it for smaller screens.
image
image


Should I make a PR?

Issue #145, opened 2017-09-15
Behind an nginx TLS proxy, Horizon tries to load ajax calls from HTTP


mattiasgeniar:
I run Horizon behind an Nginx proxy that handles SSL/TLS, and Apache is acting as the webserver that only does HTTP + PHP handling.

When I try to load the recent jobs, Horizon gives this error in the console;

```
Mixed Content: The page at 'https://staging.xxx/horizon/recent-jobs' was loaded over HTTPS, but requested an insecure XMLHttpRequest endpoint 'http://staging.xxx/horizon/api/jobs/recent?starting_at=-1&limit=50'. This request has been blocked; the content must be served over HTTPS.
```

I'm not sure where to fix this in vuejs, it appears to come from `resources/assets/js/pages/RecentJobs/Jobs.vue` with this line;

```
return this.$http.get('/horizon/api/jobs/recent/' + '?starting_at=' + starting + '&limit=' +
```

It _looks_ like `this.$http.get` doesn't keep the `X-Forwarded-Proto` in mind, like [fideloper/TrustedProxy](https://github.com/fideloper/TrustedProxy)?

alexgarrett:
I'm having the same issue here.

mattiasgeniar:
I tracked this down to the [axios HTTP client](https://github.com/mzabriskie/axios#creating-an-instance) used by Vuejs, but it looks like a rabbithole to get it fixed. Would need a config to be passed on to the view to overwrite the `baseURL`.

The Horizon code defines the HTTP client as such;

```
resources/assets/js/app.js:Vue.prototype.$http = axios.create();
````
Currently without the ability to overwrite the options given to the `create()` call.

**Comment by alexgarrett** [comment]
I've actually found a solution my end, hope this helps you.

In the default .htaccess for Laravel, I removed this:

```
# Redirect Trailing Slashes If Not A Folder...
RewriteCond %{REQUEST_FILENAME} !-d
RewriteCond %{REQUEST_URI} (.+)/$
RewriteRule ^ %1 [L,R=301]
```

Probably not the best solution long-term, but I'm using a standalone app just for Horizon, so it doesn't affect anything else for me.

mattiasgeniar:
Interesting, that fixes it for me as well. Will look at that .htaccess file to determine the protocol better!

cucxabeng:
I have a same issue, I use forge and envoyer to manage env and deploy my source code 

I just want use https only (behind cloudeflare https)

Please help me..

[...]

PR #165, opened 2017-09-27
Remove trailing slashes on Ajax calls in Vuejs

mattiasgeniar:
If you run Laravel on Apache, it has the following .htaccess rule that gets used.

```
<IfModule mod_rewrite.c>
    [ ... snip ... ]

    # Redirect Trailing Slashes If Not A Folder...
    RewriteCond %{REQUEST_FILENAME} !-d
    RewriteRule ^(.*)/$ /$1 [L,R=301]

    [ ... snip ... ]
</IfModule>
```

This looks at all requests that have a trailing slash, and redirects them to the page _without_ the trailing slash. If you're behind an HTTPS proxy (like Nginx/Cloudflare/...) this redirects you to an HTTP page, instead of an HTTPS page.

This PR tries to avoid that redirect altogether, by modifying the Vuejs calls to be _without_ the trailing slash, which are also valid.

This will solve issue #145.

### DIFF
--- a/resources/assets/js/pages/Failed/Job.vue
+++ b/resources/assets/js/pages/Failed/Job.vue
@@ -51,7 +51,7 @@
             loadFailedJob(id) {
                 this.loadingJob = true;
 
-                this.$http.get('/horizon/api/jobs/failed/' + id)
+                this.$http.get('/horizon/api/jobs/failed' + id)
                         .then(response => {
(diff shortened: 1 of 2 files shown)


jstenquist:
Confirmed that this fixed my issue in AWS under Application load balancer with ssl termination

PR #166, opened 2017-09-28
Compile assets (for recent changes)

reinink:
We're running into the issue solved in this PR (https://github.com/laravel/horizon/commit/2353b252048fa5626d0b72cf74937165905b27b2), but since the assets have not been compiled since we cannot use them. Would obviously be nice to get this tagged as a release too. ❤️ 

These assets were compiled by running `yarn run production`.


ninjaparade:
@reinink approved.

Issue #133, opened 2017-09-11
Be able to see date for job lists


pdewit:
For the lists in `horizon/recent-jobs` and `horizon/failed`, I'd like to see the date at which they failed/were created, and not only the time. I have some jobs that have failed on different days and I'd like to see on which day they failed. Might be useful to add it to the `title` attribute so it can be seen on mouseover.

bbashy:
Yeah, very confusing!

themsaid:
This is currently added.

PR #175, opened 2017-10-02
Add date for older jobs

uber1337:
This is related to #133.

I added date  for jobs older than today in order to easy identify when the issue happened.


taylorotwell:
Has conflicts.

taylorotwell:
@themsaid can you fix this?

themsaid:
@taylorotwell yes but the reason why we don't show date is to make the table more compact.

pdewit:
@themsaid Yeah I agree, that's why I noted in issue #133 that it might be best to be able to see it on mouseover or something similar.

Issue #176, opened 2017-10-03
A supervisor cannot start more than 4 worker a second


halaei:
There is .25 second sleep in the code that makes supervisor slow on starting workers. I need to have more than 100 workers, so it is not good for my case:

```php
class WorkerProcess
{...
    protected function restart()
    {
        ...
        usleep(250 * 1000);
        ...
    }...
}
```

PR #177, opened 2017-10-03
no sleep

halaei:
Fix for #176
### Changes
after every start(), the process goes to a cool-down period. During the cool-down we should not attempt to restart.
The first cool-down is 1 second long. If the process is running after a cool-down we set `restartAgainAt` to null.
If the process is not running after any cool-down, another 1 minute cool-down begins and `UnableToLaunchProcess` event is fired.

### Suggestion
For the future releases, I think we can implement [Supervisord state machine](http://supervisord.org/subprocess.html#process-states). This pull request also simulates moving the process from 'STARTING' to 'RUNNING' state in Supervisord implementation. Please see `startsecs` config [here](http://supervisord.org/configuration.html#program-x-section-values) as well.

Supervisor Transitions

### DIFF
--- a/src/WorkerProcess.php
+++ b/src/WorkerProcess.php
@@ -52,6 +52,8 @@ class WorkerProcess
     {
         $this->output = $callback;
 
+        $this->cooldown();
+
         $this->process->start($callback);
 
         return $this;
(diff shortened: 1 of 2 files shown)


taylorotwell:
I don't understand this PR.

halaei:
There is .25 second sleep in the code that makes supervisor slow on starting workers. I need to have more than 100 workers, so it is not good for my case:

```php
class WorkerProcess
{...
    protected function restart()
    {
        ...
        usleep(250 * 1000);
        ...
    }...
}
```
Instead of sleeping, my solution was to consider all the worker processes to go to a 1-second cool-down period, to prevent attempts to restart or check if they are started.

taylorotwell:
The whole logic of what is happening is just very hard to follow. I've started at it for about 5 minutes and still have no real idea what this is doing.

halaei:
Sorry for the trouble. Regarding the difficulties in communication, I wish there was a better way to interact.

PR #179, opened 2017-10-04
configurable job expiry

halaei:
I installed horizon on a project with 100k jobs an hour. No need to waste memory for so many jobs.

### DIFF
--- a/src/Listeners/TrimFailedJobs.php
+++ b/src/Listeners/TrimFailedJobs.php
@@ -16,11 +16,11 @@ class TrimFailedJobs
     public $lastTrimmed;
 
     /**
-     * How many hours to wait in between each trim.
+     * How many minutes to wait in between each trim.
      *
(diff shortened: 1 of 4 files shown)


src/Listeners/TrimFailedJobs.php
```diff
@@ -16,11 +16,11 @@ class TrimFailedJobs
     public $lastTrimmed;
 
     /**
-     * How many hours to wait in between each trim.
+     * How many minutes to wait in between each trim.
      *
      * @var int
      */
-    public $frequency = 23;
+    public $frequency = 5040;
```
ntzm: Why not keep this at hours? Otherwise this is going to be a breaking change I believe

halaei: I am not sure how this can break things for real. In fact, this initial value is no longer used in the code.

PR #191, opened 2017-10-10
Reports: limit excessive payloads (commands) via configuration option

Artistan:
When you have a series of job that have a large payload (say 2mb each) then the recent jobs listing fails in horizon due to memory exhausted when dealing with the json.

### DIFF
--- a/config/horizon.php
+++ b/config/horizon.php
@@ -45,6 +45,18 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Payload Threshold
+    |--------------------------------------------------------------------------
+    |
+    | This option allows you to configure when the Job Payload will be truncated
+    | for reports. Size of payload is limited in bytes.
+    |
+    */
+
+    'payload' => 1000,
+
+    /*
+    |--------------------------------------------------------------------------
     | Queue Worker Configuration
     |--------------------------------------------------------------------------
     |
--- a/config/horizon.php
+++ b/config/horizon.php
@@ -49,11 +49,11 @@ return [
     |--------------------------------------------------------------------------
     |
     | This option allows you to configure when the Job Payload will be truncated
-    | for reports. Size of payload is limited in bytes.
+    | for reports. Size of payload is limited in bytes, default ~0.5 Mb.
     |
     */
 
-    'payload' => 1000,
+    'payload' => 500000,
 
     /*
     |--------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


Artistan:
hold on, needs some updates.

Artistan:
Ok, so this works well now and shows the difference in the payload command size
**Payload Size Exceeded (3063100 > 500000)**

```json
    {
      "id": "315",
      "connection": "redis",
      "queue": "data-company-full",
      "name": "App\\Jobs\\TestJobWIthJson",
      "status": "reserved",
      "payload": {
        "type": "job",
        "timeout": null,
        "tags": [
          "App\\Index:17"
        ],
        "id": "315",
        "data": {
          "command": "a:2:{s:7:\"command\";s:23:\"App\\Jobs\\TestJobWIthJson\";s:7:\"payload\";s:39:\"Payload Size Exceeded (3063100 > 500000)\";}",
          "commandName": "App\\Jobs\\TestJobWIthJson"
        },
        "displayName": "App\\Jobs\\TestJobWIthJson",
        "timeoutAt": null,
        "pushedAt": 1507655149,
        "maxTries": null,
        "job": "Illuminate\\Queue\\CallQueuedHandler@call",
        "attempts": 1
      },
      "exception": null,
      "failed_at": null,
      "completed_at": null,
      "retried_by": null,
      "reserved_at": "1507655150",
      "index": 0
    },
```

taylorotwell:
Why do you have a 2mb payload?

Artistan:
We have a large data source that gets queried for import into elasticsearch. This data is then chunked into jobs with the data as part of the payload (bulkJson) which is currently setup to be a 2mb json string. The queued jobs are then worked on by multiple workers to insert/upsert into elasticsearch.

taylorotwell:
Why don't you put the large string in a relational database table first and then query it back in from your queued job using the ID. Or store the string directly in Redis and then pull it back out using an ID within your queued job. I don't suggest putting such large payloads directly into Horizon.

Artistan:
It is not "in" horizon though.  It is in the Job. The reason it is in the primary job payload is to expedite  the process.  I do not see how this is wrong by any means.

Horizon reports are just reading the job and parsing the _entire_ payload for that job rather than just reporting the Job data, the additional data is never used in the reports. Why would we require people to refactor jobs to use horizon?  This update allows for any sized job to be properly reported.

Artistan:
**Thanks for considering this.**

Here is a quick work around for anyone that wants to queue up large payloads and still have the horizon console report properly without overloading your server memory limit or your browser memory.

https://gist.github.com/Artistan/078e570f0d059820a7b4be02378820ff
**Hope it helps others.**

Issue #193, opened 2017-10-11
Use of (p)redis transactions prevents use of sentinel for high-availability redis deployment


pdbreen:
Attempting to deploy Horizon using redis-sentinel (via https://github.com/monospice/laravel-redis-sentinel-drivers) for high-availability is mostly working.  However, it hits the following exception when attempting to gather stats (and possibly other operations).

```
Predis\NotSupportedException · Cannot initialize a MULTI/EXEC transaction over aggregate connections.
Raw
```

The following - https://github.com/nrk/predis/issues/404 - indicates this can be avoided 

> That said, it's still possible to avoid $client->transaction(), which uses the above-mentioned abstraction, and just use the good old MULTI and EXEC commands (that is, $client->multi() and $client->exec()).

Any chance of Horizon being updated to make use of this work around?

PR #194, opened 2017-10-11
Eliminate use of transaction so it can be used w/sentinel (via predis)

pdbreen:
For #193.  Eliminates use of transaction for straight read/delete.  High availability is far more important to me than ensuring accuracy of stats, which I *think* would be the only possible side effect of the removal.


pdbreen:
While no longer failing w/exception, metric collection is still not working properly w/sentinel.   Within metrics dashboard, job names are missing initial characters and queue names show "false".  Seems  smembers within measuredJobs/measuredQueues is not including the prefix that substr is stripping....  Digging.

taylorotwell:
There is other things here besides the transaction. Please only do one thing per PR.

pdbreen:
Sorry - wasn't aware this PR got modified when I promoted fix for #196 on my fork (I need both changes locally).  Will resubmit this change in isolation.

PR #198, opened 2017-10-11
For #193 - Eliminate use of transaction so it can be used w/sentinel

pdbreen:
For issue #193 - Eliminate use of transaction so it can be used w/sentinel.


taylorotwell:
This fundamentally changes the atomicity of the operation.

pdbreen:
Agreed.  But high availability of my queue is far more important to me than accurate stats.

PR #199, opened 2017-10-11
Use Lua script for baseSnapshotData instead of transaction (for #193)

pdbreen:
Another try at removing transaction dependency blocking use of sentinel (#193).  I'm not a redis expert, but my quick study indicates Lua scripts are atomic (https://redis.io/commands/eval), so this should be an equivalent substitution, while allowing high availability configurations w/sentinel.

### DIFF
--- a/src/LuaScripts.php
+++ b/src/LuaScripts.php
@@ -34,4 +34,20 @@ class LuaScripts
             redis.call('hmset', KEYS[1], 'throughput', throughput, 'runtime', runtime)
 LUA;
     }
+
+    /**
+     * Get the base snapshot data for a given key and remove it.
+     *
+     * KEYS[1] - The name of the key to snapshot
+     *
+     * @return string
+     */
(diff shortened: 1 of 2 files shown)


taylorotwell:
We are never going to commit ourselves to removing all transactions from Horizon.

pdbreen:
@taylorotwell HA is pretty much a requirement for any e-commerce site.  Horizon forces use of redis instead of DB, but neither Laravel/Horizon currently support directly support use of sentinel, the redis HA solution.  So, a bit of a puzzle.  

Predis does support multi / exec, so the approach used in the PhpRedisConnector::transaction() would work for Predis as well, just not the transaction() method on the Predis client.  Should I submit a framework PR on the PredisConnector to eliminate the issue?

Or, are you saying that HA just won't be supported any time soon and I should forget Horizon/redis and flip back to DB queues?

Issue #98, opened 2017-08-18
[Request] Option to change the route prefix


fletch3555:
Is there a way to change the route prefix? I'm looking to make this a standalone app, so it would be nice to have it served from the root instead of `/horizon`.

taylorotwell:
Route::redirect('/', '/horizon');

fletch3555:
That's not the same... It's what I'm currently doing as a workaround.  I think it's perfectly reasonable to desire the base path of a package to be changed.

activeRobert:
Not sure this is what you're looking for either, but I added this in my routes file. It works, but it looks like it still uses a "horizon" prefix. I am running this as a standalone app as well, so your request would be nice. 

`
Route::get('/', function () {
    return view('horizon::app');
});
`

browner12:
all done guys, will be customizable in the next release. 😃

justinwilson-apollidon:
@browner12 When will that be released? I'm also looking for a customizable solution, though I'm looking to customize the route and the views as well.

browner12:
there's no specific timeline on releases. whenever they get around to it.

emotality:
Was this feature added? 😊  @browner12

browner12:
https://github.com/laravel/horizon/pull/200

PR #200, opened 2017-10-12
allow horizon route to be configurable

browner12:
in reference to this issue: https://github.com/laravel/horizon/issues/98

I was going to run a `trim()` on the config value, but it looks like that is taken care of in the router, which makes sense to do it higher up the chain.

I'm not really sure how to add a test for this?  anyone able to help?

### DIFF
--- a/config/horizon.php
+++ b/config/horizon.php
@@ -92,4 +92,17 @@ return [
         ],
     ],
 
+    /*
+    |--------------------------------------------------------------------------
+    | Route Prefix
+    |--------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


browner12:
i was seeing inconsistent test results when running locally. I don't think this failure is related to my PR. can someone retrigger the test?

themsaid:
Your PR is not complete @browner12 and it's not working, you changed the route prefix without updating any of the JS files which results incomplete functionality. Please make sure you test your PRs fully before submitting.

PR #206, opened 2017-10-20
Feature/allow custom middleware

qschmick:
Added ability to override default `web` middleware for horizon routes. Similar feature to [PR 200](https://github.com/laravel/horizon/pull/200).

### DIFF
--- a/config/horizon.php
+++ b/config/horizon.php
@@ -92,4 +92,14 @@ return [
         ],
     ],
 
+    /*
+    |--------------------------------------------------------------------------
+    | Middleware Configuration
+    |--------------------------------------------------------------------------
+    |
+    | Here you may define which middleware to apply to horizon routes. By
+    | default the standard web middleware is applied.
+    |
+    */
+    'middleware' => env('HORIZON_MIDDLEWARE', 'web'),
 ];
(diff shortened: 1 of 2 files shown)


taylorotwell:
Why?

qschmick:
Recently working on adding Horizon into a project but the default web middleware has several customizations that are not needed for accessing the `/horizon` endpoints. Allowing custom middleware setting would allow for implementing specific checks that are needed for this endpoint set instead.

Additionally, if wanting to add additional middleware solely for this route it is cumbersome to setup instead of just allowing a user to set the middleware they want. This would allow default functionality for the majority of users while allowing others to override without editing vendor files or running off of a fork.

As a side question, is there a reason to not allow something like this or a better way to apply a different middleware set to a vendors' routes when needed without the vendor allowing for that customization? Thanks for the feedback and let me know if any other details are needed. Cheers.

taylorotwell:
@themsaid is there a better way to do this?

themsaid:
I think it is fine doing it this way.

julienbourdeau:
Typically, this is very useful to use with Laravel Spark's `dev` middleware. 💯

Issue #186, opened 2017-10-08
stats and metrics don't work at all


vesper8:
I'm using Horizon on multiple projects, I had to make them play nice so they all work side-by-side

Perhaps as a result of this, I've never been able to get any of them to display proper stats. I have a cron job running that triggers the snapshot every 5 minutes. 

The metrics are empty:
<img width="793" alt="screenshot 2017-10-08 04 51 02" src="https://user-images.githubusercontent.com/816028/31315327-765276c4-abe4-11e7-8887-1c6dd146e244.png">

The dashboard shows no data for most of the fields, on top of that I know the number of queued jobs to be way off

<img width="795" alt="screenshot 2017-10-08 04 50 40" src="https://user-images.githubusercontent.com/816028/31315337-bf27f798-abe4-11e7-9163-1cf39ea0e266.png">

Even odder! There's nothing showing in the Recent Jobs.. even though it's processing ~20 jobs per minute for hours now

What could be causing all this?

themsaid:
To me it looks like the snapshots command is not running.

vesper8:
I thought that could be the case which is why I added this on my Kernel.php

```
$schedule->call(function () {
            \Artisan::call('horizon:snapshot');
            \Log::debug('horizon:snapshot');
        })->everyFiveMinutes();
```

And indeed my log file is riddled with 'horizon:snapshot', confirming that my cron is working properly and a snapshot is being run every 5 minutes

@themsaid any other thought?

vesper8:
ok I think I might be on to something. I am running my large conversion script again which queues up nearly 8000 jobs. I'm doing this on a fresh Forge server with no other app running on it

I just ran my console command which queued up ~7800 jobs and this is my dashboard shortly after

<img width="796" alt="screenshot 2017-10-10 01 12 14" src="https://user-images.githubusercontent.com/816028/31370302-4b1512da-ad58-11e7-8a37-0e02cbcc365f.png">

Again it doesn't show jobs per minute and shows some very large numbers elsewhere

I think this is because my app currently is not one that creates new jobs periodically, I'm using queues at the moment to run a huge conversion task and because it queues it all up at once.. it is preventing Horizon from many proper sense of this?

I guess "jobs per minute" is not a metric that should be thought of "how many jobs are being processed" per minute but rather "how many jobs are being queued per minute" ?

Is that the case @themsaid ? If so.. is this an oversight? Shouldn't the metrics put emphasis on jobs processed per minute?

[...]

PR #212, opened 2017-10-27
add error message

browner12:
kind of driven by [this issue](https://github.com/laravel/horizon/issues/186).

it's not immediately obvious that the snapshot command has failed, since there is currently no output.  not until you dig into the code do you realize you need a response for a success.

### DIFF
--- a/src/Console/SnapshotCommand.php
+++ b/src/Console/SnapshotCommand.php
@@ -34,5 +34,9 @@ class SnapshotCommand extends Command
 
             $this->info('Metrics snapshot stored successfully.');
         }
+        
+        else {
+            $this->error('Unable to take snapshot.');
+        }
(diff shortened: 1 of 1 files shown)


themsaid:
But that doesn't really indicate a failure, the snapshot can only happen every 5 minutes, scheduling the command to run every 6 minutes will guarantee a snapshot is obtainable.

themsaid:
Having `Unable to take a snapshot` error message indicates something went wrong, but actually nothing went wrong.

browner12:
hmm... i disagree.  The point of the command is to take a snapshot. Whether it failed because it failed to get a lock, or it failed because it was faster than 5 minutes, or failed for some other reason, it was "unable to take a snapshot".

PR #223, opened 2017-11-06
Base path configuration

mikebautista:
This is a cleaned up version of https://github.com/laravel/horizon/pull/216

We are needing a way to use Horizon when it isn’t at the root of the domain. This is the solution we came up with and thought it would be useful for other situations similar to ours.

All you have to do is modify `$base_path` in the `config/horizon.php` file. It is set to empty in case you don’t need to use Horizon in a sub-directory.

It has been tested both at the root and subdirectory use cases. All views and API calls/responses work as expected.


taylorotwell:
I think I'm going to hold off on this for now.

b00f:
When do you think you can merge this PR?

eduarguzher:
It should be re-considered IMO, is there a temporal workaround?

jhoff:
We're currently having to maintain our own fork so that we can use horizon on projects that aren't in the root. Maybe someday ¯\\\_(ツ)\_/¯

emparker:
A bit dirty but i published the assets and change the routes in app.js file in public folder and it works :-D

eduarguz:
Hi @driesvints In case you have some time... what are your thoughts on this?

driesvints:
@eduarguz I'll try to send something in for this. We can definitely make this happen now.

PR #248, opened 2017-11-29
more customizable route attributes

challet:
It can be used for adding a 'domain' route group config, for instance


themsaid:
This is a breaking change, config `horizon.middleware` won't work anymore now.

challet:
Or I could just add a `horizon.domain` config ?

PR #253, opened 2017-12-05
Delete monitored tags

msaladna:
As I understand the method, it's to force-expire a collection of job ids. addDays(7) prolongs the jobs another week instead of expiring as the method name implies.


taylorotwell:
This is intentional. Otherwise you would remove "recent jobs" that were being monitored as well, even though they are still recent.

PR #269, opened 2017-12-29
separate queue notification details from where queueing on

hxnk:
In it's current state, when queueing the LongWaitDetected notification, it is pushed on the same connection/queue as the one being notified about.

This PR is an attempt to fix that :)

### DIFF
--- a/src/Notifications/LongWaitDetected.php
+++ b/src/Notifications/LongWaitDetected.php
@@ -18,14 +18,14 @@ class LongWaitDetected extends Notification
      *
      * @var string
      */
-    public $connection;
+    public $longWaitConnection;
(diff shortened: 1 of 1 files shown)


taylorotwell:
I don't understand. You just renamed the variable?

hxnk:
Yeah, because both $connection and $queue also exist in the Queueable trait, but mean something else (what queue to push on vs. what queue to notify about)... And thus in the current state when pushing the notification onto the queue, it's pushed on the queue that is running slow which you might not want to do.

Sorry, could have been a bit more specific about the issue.

taylorotwell:
Gotcha.

PR #278, opened 2018-01-17
[5.6] Support block_for config option

francislavoie:
See https://github.com/laravel/framework/pull/22284

I actually find Horizon without `blpop` to have really bad latency in processing new jobs if idle. Using blocking pop reduces latency significantly. Personally, I think it should default to using `blpop`. Without it, you need to wait until the next worker ends its cooldown cycle for the job to get picked up. If it's doing a blocking pop, redis will instead serve a job to a worker immediately as it enters the queue if there's an idle worker waiting for them. This _really_ speeds things up.

This option is in the 5.6 branch, so I guess this'll need to wait until 5.6 is released, but I'd love it if this was also backported to 5.5 (I don't think it should break BC?)

### DIFF
--- a/src/Connectors/RedisConnector.php
+++ b/src/Connectors/RedisConnector.php
@@ -19,7 +19,8 @@ class RedisConnector extends BaseConnector
         return new RedisQueue(
             $this->redis, $config['queue'],
             Arr::get($config, 'connection', $this->connection),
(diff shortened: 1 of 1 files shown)


taylorotwell:
I think it can technically go in now since the 5.5 version of Redis driver will just ignore the extra constructor argument.

PR #289, opened 2018-02-02
PHPUnit 7

carusogabriel:
The fails are related to the assertions themselves, not PHPUnit. 
@taylorotwell Can I ask your help to fix them?


taylorotwell:
Submit when it is fixed.

PR #296, opened 2018-02-12
Another attempt on horizon orphans

themsaid:
In this PR we're doing 3 things:

1. Instead of manually terminating the worker processes on termination we store a cache value similar to what happens in `queue:restart`
2. On Supervisor termination we wait until all worker processes aren't running anymore before we exit
3. We calculate the longest timeout before terminating the supervisors to prevent a case where it returns zero causing the master supervisor to terminate rightaway killing the child supervisor process before it kills all the workers.

### DIFF
--- a/src/Console/TerminateCommand.php
+++ b/src/Console/TerminateCommand.php
@@ -5,10 +5,13 @@ namespace Laravel\Horizon\Console;
 use Illuminate\Support\Str;
 use Illuminate\Console\Command;
 use Laravel\Horizon\MasterSupervisor;
+use Illuminate\Support\InteractsWithTime;
 use Laravel\Horizon\Contracts\MasterSupervisorRepository;
 
 class TerminateCommand extends Command
 {
+    use InteractsWithTime;
+
     /**
      * The name and signature of the console command.
      *
@@ -41,5 +44,7 @@ class TerminateCommand extends Command
 
             posix_kill($processId, SIGTERM);
         }
+
+        $this->laravel['cache']->forever('illuminate:queue:restart', $this->currentTime());
     }
 }
--- a/src/MasterSupervisor.php
+++ b/src/MasterSupervisor.php
@@ -159,10 +159,10 @@ class MasterSupervisor implements Pausable, Restartable, Terminable
         // First we will terminate all child supervisors so they will gracefully scale
         // down to zero. We'll also grab the longest expiration times of any of the
         // active supervisors so we know the maximum amount of time to wait here.
-        $this->supervisors->each->terminate();
-
         $longest = app(SupervisorRepository::class)
-                    ->longestActiveTimeout();
+            ->longestActiveTimeout();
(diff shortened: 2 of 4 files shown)


themsaid:
@mfn yes that terminates the master, I meant instead of terminating the worker processes.

mfn:
FTR, I'm using only `horizon:terminate` when deploying and did some tests with this change before I put it into production:

- `horizon:terminate` behaviour was not "measurably" changed. As in: I call terminate during a job which runs longer => no matter whether pre- or post- this PR => horizon waits until the job is done
- however, stopping horizon via supervisor did change

Btw, I'm using something like this to artificially delay the execution because `sleep` is affected by signals; it catches it and continues the flow of execution 🤷‍♀️:
```php
        // right on top of Job::handle()
        var_dump('Delaying indexing');
        for ($i = 0; $i < 1000000000; $i++) {}
        var_dump('Performing indexing');
```

#### Before this PR
I stopped horizon when "Delaying indexing" was written to the log the second time:
```
[2018-02-19 13:30:01][1028] Processing: App\Jobs\ElasticSearch\IndexObject
string(17) "Delaying indexing"
string(19) "Performing indexing"
[2018-02-19 13:30:09][1028] Processed:  App\Jobs\ElasticSearch\IndexObject
[2018-02-19 13:30:09][1029] Processing: App\Jobs\ElasticSearch\IndexObject
string(17) "Delaying indexing"
…
Horizon started successfully.
```
=> job never wrote `Processed` => did not finished, i.e. was in fact killed/lost

#### After this PR
I stopped horizon when "Delaying indexing" was written to the log the second time:
```
[2018-02-19 13:30:56][1030] Processing: App\Jobs\ElasticSearch\IndexObject
string(17) "Delaying indexing"
string(19) "Performing indexing"
[2018-02-19 13:31:05][1030] Processed:  App\Jobs\ElasticSearch\IndexObject
[2018-02-19 13:31:05][1031] Processing: App\Jobs\ElasticSearch\IndexObject
string(17) "Delaying indexing"
string(19) "Performing indexing"
[2018-02-19 13:31:13][1031] Processed:  App\Jobs\ElasticSearch\IndexObject
…
Horizon started successfully.
```
=> it finished properly because stopped horizon via supervisor _waited_ for it to finish

This is good news because I knew that I loose jobs during supervisor stop of it, but it was acceptable because it does not happen very often (once a few months, _maybe_).

themsaid:
@mfn that's great news, so far no orphan processes reported too, hope this issue is solved as well.

halaei:
Hello @themsaid this PR caused the restart process to take longer than it should, which can results in a huge delay. I think a better fix would be to use `ignore_user_abort(true)` function, and probably ignoring SIGHUP.
I submitted a question issue, which now actually I think it is a bug. It will be great if you help resolving it:
https://github.com/laravel/horizon/issues/400

PR #302, opened 2018-03-02
Fixed issue where metrics only showed CallQueuedListener

TobiasFroberg:
When using the commandName of the jobs we would only get the class
Illuminate\Events\CallQueuedListener in our metrics tab.

Switching to using the displayName instead solves this issue.

### DIFF
--- a/src/JobPayload.php
+++ b/src/JobPayload.php
@@ -149,13 +149,13 @@ class JobPayload implements ArrayAccess
     }
 
     /**
-     * Get the "command name" for the job.
+     * Get the "display name" for the job.
      *
      * @return string
      */
-    public function commandName()
+    public function displayName()
     {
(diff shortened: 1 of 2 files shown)


taylorotwell:
I'm a little concerned changing this method may have other consequences. @themsaid may be able to review.

themsaid:
@TobiasFroberg I'd leave the `commandName()` method as it is and create a new method and use it in the metrics, that way we guarantee we don't break the code of anyone customizing horizon.

taylorotwell:
I still feel like this must be a breaking change? @themsaid are you able to recreate the problem?

themsaid:
screen shot 2018-03-08 at 9 45 15 am
screen shot 2018-03-08 at 9 45 39 am

Yes, here's a before and after this change, I don't think the change is breaking.

PR #306, opened 2018-03-07
update timeoutAt of the payload on retry from gui

xandor:
If you retry a job in the Horzion GUI and it uses the _retryUntil way_ for limiting the retries, the job will fail again instantly.

With this PR the timeoutAt attribute of the payload will get updated, using the retryUntil method (if available).


taylorotwell:
Please re-submit with tests.

Issue #178, opened 2017-10-03
Horizon Orphan Process


dbpolito:
I'm running horizon on a latest forge machine as documented... the daemon with `php artisan horizon` and on deployments `php artisan horizon:terminate` but time to time i need to manually run `php artisan horizon:purge`.

This is the output i just got after after hours of last release:

```
$ php artisan horizon:purge
Observed Orphan: 17043
Observed Orphan: 30634
Observed Orphan: 31084
Observed Orphan: 31085
```

I can confirm it's orphans by running `htop` on tree mode (press f5) and i see these process as root process, not inside the master `php artisan horizon` process.

-------

And also, every time i run purge, it ALWAYS wrongly see 1 process as orphan:

```
$ php artisan horizon:purge
Observed Orphan: 17059
```

I haven't found a pattern yet why this is happening.

killgt:
I'm having the same issue. I woke up to my main server using all the ram and 4 gb of swap disk. Running purge killed 78 processes. I'm running horizon on a Ubuntu 16.04 with php 7.1.9-1

themsaid:
Hello everyone,

Ideally you shouldn't have any rogue processes running horizon, but for exceptions like you shared we built the purge command for you to run on regular basis if your sever is having many rogue processes, this is a safe approach until we're able to investigate the issue more.

So I recommend that you schedule the command to run every day.

On the other hand, can you please share some information about your setup? I'm trying to understand what might cause rogue processes.

dillingham:
*Daemon* command: `php artisan horizon` directory `/home/forge/sitename.com/current`
In envoyer when I just added `php artisan horizon:purge` to my deployment hook, it purged 20+
Im guessing each of my deployments since running horizon
I began investigating when I noticed old code being used even though I `horizon:terminate`

killgt:
Our envoy deploy script uses horizon terminate to kill the processes, the high number of orphan procesess maybe are related to this. We added a horizon purge too at the deployement script. @themsaid if you find difficult to reproduce this bug, we could arrange ssh access to my vps.

jkudish:
Sorry to hijack the thread, but I'm seeing weird behaviour with Horizon running old code after I've deployed fresh code. Wondering if it's related to the orphans.

I've got my stuff setup through envoyer/forge. 

I've setup the following hook in Envoyer, I currently have it running it after `Purge Old Releases` -- is that where others set it up too or does it make more sense to have it at a different spot?

[...]

PR #311, opened 2018-03-13
Fix Purge always reporting 1 orphan process issue

elghobaty:
Related to https://github.com/laravel/horizon/issues/178: Every single time "purge" runs, it always reports finding 1 process as an orphan, as it turns out that this case is kind of a false positive.

`ProcessInspector`'s `current` method runs the command `pgrep -f horizon` (within a process whose ID is `X`), this process ID `X` is reported back as a result of the same command (since it contains "horizon"). 

Running `pgrep -lf horizon` from the purge command confirms this, it returns something that looks like this:
```
14302 php
14306 php7.2
15912 php7.2
15945 php7.2
16765 php
16771 sh        ---> Used to be reported as an orphan process
```

I have used a simple regular expression that would match `horizon` (An actual process) but not `[h]orizon` (The process that runs `pgrep`).

### DIFF
--- a/src/ProcessInspector.php
+++ b/src/ProcessInspector.php
@@ -33,7 +33,7 @@ class ProcessInspector
     public function current()
     {
         return array_diff(
-            $this->exec->run('pgrep -f horizon'),
+            $this->exec->run('pgrep -f [h]orizon'),
             $this->exec->run('pgrep -f horizon:purge')
         );
     }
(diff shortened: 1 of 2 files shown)


taylorotwell:
Great. Thanks!

themsaid:
@elghobaty greetings from Hurghada to Alexandria, thanks man!

elghobaty:
You are more than welcome ya m3alem 😄 @themsaid, and thank you both for all your efforts, and especially for Horizon ^^

PR #319, opened 2018-04-11
Expose middleware and timeouts

toopay:
I believe user will have a hard time figuring out that these values can be used. Use case for `middleware` will be someone applying their own custom auth. Specifically for `timeout`, most likely user will need to adjust this value if they had adjusted `queue.connection.redis.retry_after`.


toopay:
Hey @taylorotwell would be nice to know why this PR not compliant/desirable?

PR #325, opened 2018-04-22
Availabilty to display current running config

Donny5300:
Hi Themsaid,

I've added a button on the dashboard page to give the option for showing the current running config. The reason why i've added it is because I couldnt find a good spot to compare my current running config with the config from the horizon.php file.

It needs a bit of styling in the modal because i dont use Bootstrap 4 yet, my apologise for it!


taylorotwell:
No plans to add right now.

PR #326, opened 2018-04-23
Add redirect on Horizon logo

Donny5300:
I've added a config option to set a url. Once filled in, you can click the logo to get redirected to any path you would like. Added this so you can return easily to your own application.


taylorotwell:
Hmm, no plans to add this right now.

PR #347, opened 2018-05-27
Composer Update

fstegmann:
Composer update and a change to service bindings because another package complained about it (IdeHelper)


fstegmann:
Also the Horizon landing page was showing scrollbars

taylorotwell:
1.0 branch is the current release of Horizon.

PR #372, opened 2018-07-20
add tooltip on tags column

BinotaLIU:
Tags column will be truncated, I had always open the developer tools and see the raw ajax response to know what tags the job has...

This PR added the tooltip to show the full tags.


taylorotwell:
@themsaid can you review and verify this in the UI?

Issue #251, opened 2017-12-05
Add default uri & middleware to config file


markwalet:
When browsing the code in the `HorizonServiceProvider` class, I found out that you can customize the `uri` and `middleware` options for Horizon.

Maybe it is a good idea to add this options to the configuration file. That will clarify which options are customizable.

SirLamer:
Having experimented a little, at this time I don't think it actually responds well to configuring those options yet.

markwalet:
I know. but it does give you the option to do so.

SirLamer:
Yeah but it breaks everything, so mostly don't do it.

I suspect it's setup for future support.

SirLamer:
Just to add to this, it responds well to changes to the 'middleware' but seemingly not the 'uri'.

Example:
```php
// ...

'middleware' => ['web', 'horizon'],

// ...
```

markwalet:
I agree. But what does the `horizon` middleware do in your example? In the base `Controller` class the `Authenticate` middleware gets registered regardless of the middleware config option. And `horizon` isn't an existing middleware.

This should be reworked. I think that the `Authenticate` middleware should be registered with a name. I suggest `horizon`. Then we can make the middleware optional via configuration.

SirLamer:
Well I introduced a hypothetical custom middleware group, otherwise I don't see the point of changing it.

In my case I am handling the Horizon auth check via a middleware implementation, similar to [here](https://github.com/laravel/docs/issues/3590#issuecomment-346991788) except modified as described below the linked post.

markwalet:
Ah okay. I understand

I still think the `Authenticate` middleware from this package should be optional. Not mandatory. You could even remove the `Horizon::auth()` method completely when you can provide your own middleware. But that's something for v2.

SirLamer:
It's required because otherwise people will install it and then expose their Horizon admin UI to the public, it's a huge security risk.

Note you only need to do auth handling when deployed to production, the default is to allow access from the local machine.

umairali:
I think simple solution is just to add this line to config/horizon.php file

`'middleware' => ['web', 'auth'],`

driesvints:
I'm not sure why the middleware option in the config file was removed again but it might return at a later time. At the moment you're free to add it to your own `horizon.php` file.

Issue #65, opened 2017-07-31
Can't middleware be used ?


tintnaingwinn:
Is there any way to use middleware for route?

themsaid:
For what purpose?

tintnaingwinn:
For Authentication
```
Route::group(['middleware' => ['auth', 'auth.admin']], function() {
     Horizon::auth(function ($request) {
            return true;
     });
}
```

themsaid:
Just use the `Horizon::auth` method in your service provider and decide using the given request object wether this user should have access to horizon or not:

```php
Horizon::auth(function ($request) {
            return $request->user()->is_admin;
        });
```

tintnaingwinn:
$request->user() always return null
```
Horizon::auth(function ($request) {
            return $request->user(); // Null
        });
```

jpmurray:
I was coming to report this also. I'm successfully logged in, but can't access user from the request object.

From the best of my knowledge, to get auth we need to have sessions enabled (they are in middlewares) and service providers are fired _before_ middlewares, no? Maybe @themsaid can confirm this, we might just be doing something wrong also...

themsaid:
It's fixed now in the latest commit :)

jpmurray:
Thank you @themsaid. Will definitely be checking how you can send logged user to a service provider in the source code now! ;-)

themsaid:
You register a callback, and the Horizon calls it in a special middleware :)

tintnaingwinn:
Thank you @themsaid. 👍

gadhiamitul41:
Can you tell me which service provided to add the code in? 

$request->user() always return null

Horizon::auth(function ($request) {
            return $request->user(); // Null
        });

tintnaingwinn:
@gadhiamitul41 waits stable version. It's fixed in the [useWebMiddleware](https://github.com/laravel/horizon/pull/68/commits/1317f543ba0025b01e3627e8138b0ceb708c6055) commit. You can play with 1.0.x-dev version.

ninjaparade:
This commit https://github.com/laravel/horizon/pull/68/commits/1317f543ba0025b01e3627e8138b0ceb708c6055 does indeed solve the issue of the user being null, just curious what approach people are taking when you're authing is API based?

jpmurray:
Surfing on what @ninjaparade says, it got me thinking: there is no way to redirect somewhere if a user isn't auth'ed, and it just throws an error page... Anyway to show something else?

[...]

PR #378, opened 2018-08-10
Add middleware to the default config

francislavoie:
See here: https://github.com/laravel/horizon/issues/65#issuecomment-412128134

Partially fixes https://github.com/laravel/horizon/issues/251

This option has been available for a long time via `HorizonServiceProvider` but never documented as such.


francislavoie:
Apparently this has already been attempted before in #319, #248 and #206 which implemented this feature, but apparently the config got removed since then? Why was this made undocumented?

francislavoie:
@taylorotwell why are we keeping it hidden? It's fully functional, just undocumented. There's some very clear and obvious use-cases for it, i.e. hiding the horizon dashboard behind an admin auth gate for example.

PR #379, opened 2018-08-10
Allow overriding the horizon default uri

VinceG:
Hey,

This PR allows one to change the default Horizon URI. It was actually believed to be a config option but the routes were hard coded in the vue components making the change impossible unless we dynamically load the route.

This might be a big PR but it basically has the following changes:

- Add 'uri' to the horizon config file with 'horizon' being default.
- Update the horizon app.blade.php file to include a global object so we can access the customized route in the vue application.
- Vue router base url was set to the horizon.uri variable 
- Axios baseUrl was set to the current location + the horizon.uri variable.
- Vue Components were updated to remove the hard coded /horizon/.. prefix
- Minor changes to few Vue Components to make sure loops have a key attribute.
- Ran npm run prod to update the assets.

Since there are no front end test i didn't add any. I tested this as much as i could (there aren't that many routes to test) and everything seems to be working as it should.

Let me know if you have any other questions.


config/horizon.php
```diff
@@ -28,6 +28,19 @@
 
     'prefix' => env('HORIZON_PREFIX', 'horizon:'),
 
+    /*
+    |--------------------------------------------------------------------------
+    | Horizon Redis Prefix
```
francislavoie: This isn't the redis prefix

VinceG: @francislavoie thanks, fixed.

taylorotwell:
No plans to allow this.

VinceG:
@taylorotwell OK, Thank You.

VinceG:
In case anyone is interested in this it's available under the 1.4.0 tag in the following repository
https://github.com/VinceG/horizon

https://github.com/VinceG/horizon/releases/tag/v1.4.0

andrewmclagan:
Forcing this behaviour basically makes Horizon unusable for any scenario in a multi-host routing environment.

Imagine you have multiple Laravel apps running on `laravel-app.com` that have a forward proxy doing path based routing: 

* `laravel-app.com/api/endpoint-1` Laravel app one
* `laravel-app.com/api/endpoint-2` Laravel app two
* `laravel-app.com/api/endpoint-3` Laravel app three

Its not possible to use easily access the horizon UI with the current implementation as the Vue app expects a hard coded `horizon` path.

VinceG:
@andrewmclagan Yeah, that was one of the issues we had with the implementation. I've forked the repository and made the changes, we use this fork in production with a custom route. Everything seems to be working so far without any issues.

I'll be updating the fork periodically even though i don't think there going to be a lot of activity in the official Horizon repository.

andrewmclagan:
I had a look through your repo - looks great. Interested in a couple of small PRs? To be honest I'm happy to migrate to a project that is not open-source by dictatorship.

VinceG:
Sure PRs are welcome 

-----
Vincent Gabriel
Software Engineer

Please excuse my brevity, sent from my iPhone

Issue #377, opened 2018-08-08
Metrics JS Error won't load the page


VinceG:
```
Uncaught (in promise) TypeError: Cannot create property 'time' on string '<'
```

![screen shot 2018-08-08 at 2 14 24 pm](https://user-images.githubusercontent.com/195199/43864736-5f941976-9b15-11e8-98c3-96663df3e156.png)
![screen shot 2018-08-08 at 2 11 05 pm](https://user-images.githubusercontent.com/195199/43864737-5fa983a6-9b15-11e8-8ef3-de662402e0dc.png)

PR #380, opened 2018-08-10
Fix issue with metrics queue route not loading due to parameter name

VinceG:
This fixes the issue #377. The reason why this was happening is due to the fact that a queue name most likely will be alpha numeric. Since the route binding expects {id} the {id} in most applications (at least for us) has a global pattern set to [0-9]+ or \d+ this prevented the route from working properly since our queue was named 'default' the {id} didn't match to that pattern.

The change is simple, instead of {id} i changed it to {slug} (it also matches the parameter name in the controller action).

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -20,7 +20,7 @@ Route::prefix('api')->group(function () {
 
     // Job Metric Routes...
     Route::get('/metrics/jobs', 'JobMetricsController@index')->name('horizon.jobs-metrics.index');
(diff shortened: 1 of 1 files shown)


taylorotwell:
What if someone has a rule on slug? IMO just don't use global routing rules like that. It's asking for problems with packages like this.

VinceG:
@taylorotwell That's unfortunate.

PR #381, opened 2018-08-13
Suggest predis instead of requiring it

ThomHurks:
Just like laravel/framework which suggests predis (and people can choose to use phpredis instead), Horizon should suggest it so people using phpredis do not needlessly install a dependency.

Horizon relies on the redis config in config/database.php anyway, and trusts that the user correctly fills in that information. Requiring the user to install phpredis or predis herself is no different from what other parts of Laravel using Redis already do.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,6 @@
         "illuminate/contracts": "~5.5",
         "illuminate/queue": "~5.5",
         "illuminate/support": "~5.5",
-        "predis/predis": "^1.1",
         "ramsey/uuid": "^3.5",
         "symfony/debug": "~3.3|~4.0"
     },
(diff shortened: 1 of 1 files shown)


taylorotwell:
Maybe on the next major version. I don't want to uninstall Predis from people's working applications on a point release.

ThomHurks:
@driesvints @taylorotwell Can we add this to version 3.0 maybe, since it was only just released?

browner12:
since it's released already, we can't make a change like this, even if it was just yesterday.

I would suggest PR'ing it to `master`.

driesvints:
Yes please pr to master but it'll be a while maybe before we make another major release.

ThomHurks:
@driesvints Created PR #531 to Master.

Issue #182, opened 2017-10-06
Add Retry All button


MECU:
It would be nice to have  Retry All button

Feature request

michaelnguyen547:
This may relate to #310 

Atm, a new job is created after failed retry. Ideally, after a failed retry, the original job should be deleted and a new job is created to prevent having duplicated jobs. 

e.g. Job#1 failed, Job#2 (duplicate of job#1) is created, then retry Job#2 which fails again, Job#3 (duplicate of job#2) is created. Now I have Job#1, Job#2 and Job3# do the same thing. If "Retry All" would work, I would execute 3 jobs.

dpde:
> It would be nice to have Retry All button
> 
> Feature request

Yes, would be a nice addition.
In the past, I just opened my Chrome dev tools console and executed `$('table i.icon').click();`. Good enough for me.

driesvints:
Hey all, we're gonna close this as Taylor has rejected this in the past. I don't think we'll re-consider this as we don't feel this is useful in a production environment. Sorry.

xwiz:
@driesvints What would you advice alternatively in a production environment?

The queue retry all command doesn't seem to use a cursor either so ends up getting killed if you have many jobs. It will also be great to be able to retry all the jobs on a particular connection.

xwiz:
For anyone interested, I created a model scaffold for FailedJobs table and ended up using something like this:

```

<?php

namespace App\Console\Commands;

use Illuminate\Console\Command;
use App\Models\FailedJob;
use Carbon\Carbon;

class FailedJobsCommand extends Command
{
    /**
     * The name and signature of the console command.
     *
     * @var string
     */
    protected $signature = 'jobs:failed {connection?}';

    /**
     * The console command description.
     *
     * @var string
     */
    protected $description = 'Retry failed jobs';

    /**
     * Create a new command instance.
     *
     * @return void
     */
    public function __construct()
    {
        parent::__construct();
    }

    /**
     * Execute the console command.
     *
     * @return mixed
     */
    public function handle()
    {
        \ini_set('memory_limit', -1);
        \set_time_limit(0);
        $this->retryJobs();
    }
    
    public function retryJobs()
    {
        $total = 0;
        $connection = $this->argument('connection') ?: null;

        pcntl_async_signals(true);
    
        pcntl_signal(SIGINT, function () use ($total) {
            $this->info("$total jobs queued");
            return;
        });
        pcntl_signal(SIGTERM, function () use ($total) {
[...]

PR #389, opened 2018-09-12
Retry all failed jobs

elghobaty:
I've implemented the logic to retry all failed jobs at once as requested here https://github.com/laravel/horizon/issues/182

### Approach:
1. Copy/Create a snapshot of the failed jobs.
2. Retrieve a page by page list of failed jobs.
3. For each page, re-push the failed jobs on queue excluding those with pending/completed retries and jobs which are already retries of other jobs (i.e., only the original job is re-queued).
4. Delete the snapshot created at step 1.

#### Why create a snapshot of the failed jobs and not process them directly?
The failed jobs Redis entry is a stack: recently failed jobs are always at the top. So, if for example while re-queueing the page no 1 of failed jobs (e.g. 50 jobs), 30 fresh jobs fail (and then added to the top of the stack), when processing page no 2, we would actually get 30 jobs that have already been processed. While I actually avoid re-queuing them, I guess this could lead to extended execution time in an unexpected way.

#### Creating a snapshot (Redis copying) approach:
I am not sure but I guess copying all the failed jobs could exhaust memory limits if done on Horizon level through the job repository. I decided to go for a Lua script. My knowledge of Redis is not profound but I came across [this article](https://redislabs.com/blog/the-7th-principle-of-redis-we-optimize-for-joy/ ) which provides a code snippet which is described as 

> The fastest, type-agnostic way to copy a Redis key

.

### Breaking changes
- Three methods have been added to the `JobRepository` interface.

--- 

If you agree with this approach, then there is another change I have in mind which I'd like to run by you: there could be only 1 "retry all failed jobs" process running at a given time.

### DIFF
--- a/src/Contracts/JobRepository.php
+++ b/src/Contracts/JobRepository.php
@@ -190,4 +190,28 @@ interface JobRepository
      * @return int
      */
     public function deleteFailed($id);
+
+    /**
+     * Get a chunk of jobs from the given type set.
+     *
+     * @param  string  $type
+     * @param  string  $afterIndex
+     * @return \Illuminate\Support\Collection
+     */
+    public function getJobsByType($type, $afterIndex);
+
+    /**
+     * Inserts a snapshot of the failed jobs into storage.
+     *
+     * @return string|null the snapshot ID or null if the operation fails.
+     */
+    public function snapshotFailedJobs();
+
+    /**
+     * Delete a snapshot of the failed jobs from storage by ID.
+     *
+     * @param  string  $id
+     * @return bool
+     */
+    public function deleteFailedJobsSnapshot($id);
 }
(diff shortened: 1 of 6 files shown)


taylorotwell:
No plans to add this at the moment. I believe you can do `php artisan queue:retry all`.

xwiz:
I believe an approach as mentioned by another user works for this. Something like this:

> I made a little endpoint which seems to do the trick:
> 
> ```
> Route::get('/retry-all', function () {
>     /** @var JobRepository $jobs */
>     $jobs = app(JobRepository::class);
>     return $jobs->getFailed()->map(function ($job) {
>         $id = $job->id;
>         dispatch(new RetryFailedJob($id));
>         return $id;
>     });
> });
> ```

This should consume much less memory than using `php artisan queue:retry all` which can fail deeply if the failed jobs is really a big list.

PR #393, opened 2018-09-27
Show correct timescale on the Dashboard.

JayBizzle:
I find it mildly annoying that we have our `horizon.trim.recent` config set to 24 hours, but the dashboard still says `Failed Jobs past hour` etc.

This PR uses the config value to display the number of hours.

Before...
screenshot 2018-09-27 at 19 32 47

After (assuming `horizon.trim.recent` set to `1440`)...
screenshot 2018-09-27 at 19 32 28

Not sure if this is the best implementation, but it was the best i could come up with.

Thanks

### DIFF
--- a/resources/assets/js/pages/Dashboard.vue
+++ b/resources/assets/js/pages/Dashboard.vue
@@ -143,14 +143,14 @@
                                 </span>
                             </div>
                             <div class="stat col-3 p-4">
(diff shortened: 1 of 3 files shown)


themsaid:
I think it's better to just say "Recent Jobs", "Recent Failed jobs", so people can use recency that's less than 60 minutes.

PR #399, opened 2018-10-15
change recent jobs status icon td and failed jobs retry td from last …

Mohanrau:
When Recent Jobs & Failed Job tables has long text jobs name or long text tags labels or long runtime numbers user have to scroll the table to check the status or to click the retry button, so my changes will enhance user experience on all screen size.

Before
before

After
after


themsaid:
I don't think it's a good idea to change the order of the columns like this. I do believe the UX can be enhanced, I think we'll do a frontend overhaul in 2.0

PR #401, opened 2018-10-20
Removed overflow:scroll, as they show scroll bars when not needed

SDekkers:
Currently the property causes scrollbars to be always visible in Chrome, Firefox and Edge, as shown in below screenshot:

before

After this commit:
after

### DIFF
--- a/resources/assets/sass/app.scss
+++ b/resources/assets/sass/app.scss
@@ -66,8 +66,6 @@
 }
 
 .stats {
-    overflow: scroll;
-
     .stat {
         border-right-style: solid;
         border-right-width: 1px;


themsaid:
I haven't seen this anywhere, not on Chrome or Firefox or Safari.

SDekkers:
@themsaid I’m not the only one seeing those scrollbars, have you tries any Windows based OS?

Another thread about Horizon on SO also shows the scrollbars
https://stackoverflow.com/questions/50148072/how-to-edit-laravel-horizon-view

themsaid:
If we add your line the layout will break for queue names that are long, the property was added to cause the view to scroll when needed.

Anyway we're planning a frontend overhaul soon, will check if the issue exists on windows and take care of it.

SDekkers:
Alright, thanks for looking into it

PR #402, opened 2018-10-20
dont wait for workers to terminate

### LINKED ISSUE #400: Restaring horizon via horizon:terminate & suprevisord can cause long delay on jobs
Currently, I issue a `artisan horizon:terminate` and wait for supervisord to start horizon after the current horizon process gets terminated.
This approach has an issue. If there is a job that is supposed to take 1 hour more to gets processed, horizon:terminate will wait for that job one hour without processing other jobs. So it forces a 1 hour delay.

PS: Another option is to call `artisan queue:restart`. However, this command has some limitation. One is that restarted jobs don't receive changes in the .env file. Another issue is it doesn't let the horizon know of any change in it's own configuration.

halaei:
This PR attempts to fix #400 without causing orphans as it was a concern of another PRs, one being #296  
I write some comments on the codes to make things clear.

### DIFF
--- a/src/Console/WorkCommand.php
+++ b/src/Console/WorkCommand.php
@@ -31,4 +31,16 @@ class WorkCommand extends BaseWorkCommand
      * @var bool
      */
     protected $hidden = true;
+
+    /**
+     * Execute the console command.
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        ignore_user_abort(true);
+
+        return parent::handle();
+    }
 }
--- a/src/ProcessPool.php
+++ b/src/ProcessPool.php
@@ -5,7 +5,6 @@ namespace Laravel\Horizon;
 use Closure;
 use Countable;
 use Cake\Chronos\Chronos;
-use Symfony\Component\Process\Process;
 
 class ProcessPool implements Countable
 {
@@ -175,7 +174,7 @@ class ProcessPool implements Countable
      */
     protected function createProcess()
     {
-        return new WorkerProcess((new Process(
+        return new WorkerProcess((new BackgroundProcess(
             $this->options->toWorkerCommand(), $this->options->directory)
         )->setTimeout(null)->disableOutput());
     }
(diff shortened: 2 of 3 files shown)


src/Console/WorkCommand.php
```diff
@@ -31,4 +31,16 @@ class WorkCommand extends BaseWorkCommand
...
     protected $hidden = true;
+
+    /**
+     * Execute the console command.
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        ignore_user_abort(true);
```
halaei: This line prevents the worker to being killed if it print to stdout after supervisor dies.

src/ProcessPool.php
```diff
@@ -175,7 +174,7 @@ protected function start()
      */
     protected function createProcess()
     {
-        return new WorkerProcess((new Process(
+        return new WorkerProcess((new BackgroundProcess(
```
halaei: This prevents the supervisor to send SIGKILL signal to worker on object distruction. It may also be a good idea to send SIGTERM in the BackgroudProcess::__destruct().

src/Supervisor.php
```diff
@@ -225,10 +225,6 @@ public function terminate($status = 0)
             });
         });
 
-        while ($this->processPools->map->runningProcesses()->collapse()->count()) {
```
halaei: No need to wait for children any more. They have already received the SIGTERM signal, and we made sure no other mechanism causes them to terminate while handling the current job.

taylorotwell:
Did you actually test this on a heavy-duty production application?

halaei:
Not yet. Only on development. Will use in production tomorrow.

halaei:
Hi @taylorotwell 
By the way I was thinking in some situations horizon actually must wait for all the workers to terminate:
1. When the server memory is limited running a new instance of horizon with the presence of workers of a previous horizon might lead to memory outage.
2. When an application must be sure there is no more than one worker on a queue.

I will send a PR tomorrow to add an option --wait to horizon:terminate.

taylorotwell:
Reverting until it's tested on a good production environment and proven to not cause orphans.

Issue #264, opened 2017-12-15
Horizon makes the assumption that getKey() will return a UTF-8 string


francislavoie:
This issue is a bit of a cross-report with https://github.com/spatie/laravel-binary-uuid/issues/28

If the primary key for a model is binary, then the auto-tagging in Horizon fails, then the web dashboard stops working as well - everything ends up in a bad state until I restart my redis instance (clearing out its data).

I get errors like this:

    [2017-12-15 21:05:25] local.ERROR: ERR Error running script (call to f_9bd406a64e43e31f0dde2849660e8cd56f340b85): @user_script:7: user_script:7: Expected value but found T_END at character 1  {"exception":"[object] (Predis\\Response\\ServerException(code: 0): ERR Error running script (call to f_9bd406a64e43e31f0dde2849660e8cd56f340b85): @user_script:7: user_script:7: Expected value but found T_END at character 1  at /var/www/laravel/vendor/predis/predis/src/Client.php:370)

Also, `/horizon/recent-jobs` shows a spinner forever with an error in the console saying `Cannot read property 'tags' of null` on every poll to get data.

Anyways - the problem is that [`$model->getKey()`](https://github.com/laravel/horizon/blob/1.0/src/Tags.php#L29) is expecting the key to be a UTF-8 string value to pass directly to Redis. This isn't always the case. I don't think that overriding `getKey()` is a solution either, because that's used directly in the Laravel framework to do `WHERE` queries to SQL DBs. I'm not sure what the best solution would be.

driesvints:
As said in the closed PR, returning something else than a readable key will probably cause more problems than just this. For now Horizon expects primary keys to be UTF-8 and I don't think this will change soon.

PR #405, opened 2018-10-23
Call a potential getReadableKey method instead of getKey

tdondich:
If the key for your eloquent model is binary, such as when using spatie/laravel-binary-uuid , then horizon will fail when it attempts to generate the tags.  You could use the manual tags feature; however, this functionality won't work if you're using broadcast notifications (since we cannot set tags explicitly on the related BroadcastNotificationCreated job)

My suggestion is to have the Tags system of horizon look for an optional getReadableKey method. If it exists, it will utilize the return value of it instead of the raw value of getKey, which in this case would be a binary value and would fail serialization attempts.

This PR provides a solution for https://github.com/laravel/horizon/issues/264

### DIFF
--- a/src/Tags.php
+++ b/src/Tags.php
@@ -26,7 +26,7 @@ class Tags
         }
 
         return static::modelsFor(static::targetsFor($job))->map(function ($model) {
-            return get_class($model).':'.$model->getKey();
+            return get_class($model).':'.(method_exists($model, 'getReadableKey') ? $model->getReadableKey() : $model->getKey());
         })->all();
     }
 


src/Tags.php
```diff
@@ -26,7 +26,7 @@ public static function for($job)
         }
 
         return static::modelsFor(static::targetsFor($job))->map(function ($model) {
-            return get_class($model).':'.$model->getKey();
+            return get_class($model).':'.(method_exists($model, 'getReadableKey') ? $model->getReadableKey() : $model->getKey());
```
driesvints: Write this as:

```suggestion
            $key = method_exists($model, 'getReadableKey')
                            ? $model->getReadableKey()
                            : $model->getKey();
            
            return get_class($model).':'.$key;
```

taylorotwell:
IMO having getKey() return something other than a nice string is going to cause a lot more problems than just with Horizon. It's used throughout Laravel.

tdondich:
@taylorotwell What would be your approach then given that many people use the spatie binary uuid package to allow for binary uuids for primary key?  Would love your thoughts to see if we should/could go another way.

PR #406, opened 2018-10-23
[1.0] Make wait optional in horizon termination

halaei:
Regarding #400 this PR is #402 plus making wait optional. As I mentioned in #402 wait option can be used if:
1. The server has not enough memory to run a new instance of horizon with the presence of workers of a previous horizon.
2. An application must be sure that there is no more than one worker on a queue.

I have this in production as well as local. It seems to be fine and my experience is that the deployment is now much smoother. So @taylorotwell I think it is OK to either merge this or revert #404.
Thanks.

### DIFF
--- a/src/Console/TerminateCommand.php
+++ b/src/Console/TerminateCommand.php
@@ -17,7 +17,8 @@ class TerminateCommand extends Command
      *
      * @var string
      */
-    protected $signature = 'horizon:terminate';
+    protected $signature = 'horizon:terminate
+                            {--wait : Horizon supervisors should wait for their workers to terminate}';
 
     /**
      * The console command description.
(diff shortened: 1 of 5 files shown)


halaei:
@driesvints  I will take take a loot.

halaei:
@driesvints It seems there was failed test in other commits as well. All tests was passing in my travis account:
https://travis-ci.com/halaei/horizon/builds
Maybe the failure has nothing to do with this PR?

driesvints:
@halaei hmm, seems you're right. I'll try to have a look in a bit.

driesvints:
@halaei seems like it's totally sporadic. Weird. Don't really know what's going on as it's different tests failing every time.

halaei:
@diesvints maybe it is something related to time, e.g. it assumes some process should terminate after 2 seconds, but it takes longer to terminate because Travis is slower than expectations. Just guessing...

taylorotwell:
Is it not possible to make this more opt-in? Via some configuration value perhaps to enable ignore_user_abort? And then the opposite of --wait? Could you submit a PR for that. That way we definitely avoid any breaking changes and you will be able to opt-in to the new configuration value / flag.

halaei:
@taylorotwell I do another attempt tomorrow and let you know. My concern is that eventually not waiting should be the right choice for most of the time. Then it will be a bad idea to not have it by default.

PR #410, opened 2018-10-24
[1.0] update Bootstrap to stable version

browner12:
looking through the changelogs and the source code, it doesn't appear there were any breaking changes that affect this package.

http://getbootstrap.com/docs/4.1/migration/#beta-3-changes
http://getbootstrap.com/docs/4.1/migration/#stable-changes

both `npm run dev` and `npm run prod` still compile successfully.

good to get us on stable, and 4.2 will have some good container upgrades to help us take better advantage of screen real estate.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     },
     "dependencies": {
         "axios": "^0.15.3",
-        "bootstrap": "^4.0.0-beta.2",
+        "bootstrap": "^4.0.0",
         "chart.js": "^2.5.0",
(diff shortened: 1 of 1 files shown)


browner12:
failing test seems unrelated

browner12:
figured I'd let you guys do the asset recompiling. don't want just anyone PRing compiled files to the repo, as that code is not going to be reviewed closely and malicious users could put bad stuff in there.

PR #413, opened 2018-10-26
Add home-link svg-button to navigate back to main app

oheck:
This adds a zondicon-home svg button to the top right corner of the app with a link to navigate back to the main application. 

The url can be defined in the horizon.php configuration file (home-link). If "home-link" is empty, the home-button will not be visible. (This is the default)

If you want to have a link / button to get back to the main application, the only way i found to extend horizon is via forking it, i thought you might be interested in including that feature in the main repository. 

In nova you gave us the option to easily extend features and add custom links for example. That was the reason for making it configurable via the configuration.


oheck:
screenshot 2018-10-26 00 14 38

oheck:
@driesvints please check again... or do you want me to exclude it completely?

driesvints:
@oheck it's ok :)

PR #417, opened 2018-10-27
Ability to make authorization by gate

slavarazum:
Provide ability to **Horizon** dashboard access by gate. Like it possible in **Nova** and **Telescope**.
```php
Gate::define('viewHorizon', function ($user) {
    return true;
});
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,6 +17,7 @@
         "ext-pcntl": "*",
         "ext-posix": "*",
         "cakephp/chronos": "^1.0",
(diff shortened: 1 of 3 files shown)


taylorotwell:
Can you send this to master branch?

slavarazum:
Of course, I made PR into *1.0* because it set as main branch.
Just have sent a new PR into `master`, but looks like **1.0** branch has some unmerged into master commits. Should we add these commits into `master` branch? Or I can create branch from the `master` and resend it.

PR #425, opened 2018-10-30
[1.0] catch exception on sending signal

halaei:
There is a small chance that the process is still running when calling `$this->process->isRunning()` but it stops during `$this->process->signal()`. It happened to our projects only once, but since I am not sure of the possible consequences, I am sending a fix. Note that isRunning() is called by signal() itself, and do some logic that is better to not be done twice. Therefore, I only call isRunning() after catching a Symfony exception, to see if the exception is not expected.

By the way, I didn't know they have interface for their exceptions. Cool!

### DIFF
--- a/src/WorkerProcess.php
+++ b/src/WorkerProcess.php
@@ -4,9 +4,9 @@ namespace Laravel\Horizon;
 
 use Closure;
 use Cake\Chronos\Chronos;
-use Symfony\Component\Process\Process;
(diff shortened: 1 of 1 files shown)


halaei:
Funny. It failed again

halaei:
Here is an example of what can happen in travis:
https://travis-ci.com/halaei/horizon/jobs/155094246

PR #426, opened 2018-10-30
[1.0] fast_termination config and --wait option

### LINKED ISSUE #400: Restaring horizon via horizon:terminate & suprevisord can cause long delay on jobs

halaei:
Reopening #406 to fix #400.

Installations of horizon don't see the change unless they opt-in by setting `'fast_termination' => true` in their horizon config file. Maybe in 2.0 release the default can be set to true.

After enabling fast_termination config, the previous behaviour can still be selected by `--wait` option in horizon:terminate command. I think by these settings its better to have `--wait` instead of `--dont-wait`.

### DIFF
--- a/config/horizon.php
+++ b/config/horizon.php
@@ -61,6 +61,21 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Fast Termination
+    |--------------------------------------------------------------------------
+    |
+    | If this option is set horizon:terminate command don't wait for all the
+    | workers to terminate unless --wait option is provided. Fast termination
+    | minimizes the deployment delay on the queued jobs by letting a new
+    | instance of horizon gets started while the workers of a previously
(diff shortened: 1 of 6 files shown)


mfn:
And is it guaranteed that existing workers properly terminate after the worked their current job?

halaei:
@mfn as far as I know, yes.

taylorotwell:
Thanks

PR #428, opened 2018-10-30
Don't display status icon until stats.status is initialised

MaxGiting:
When you first load horizon the status box has a red x until the status has been retrieved. 

Too many times when loading horizon to look into an error email I have received, I've had a split second panic when I see the red before it quickly changes. :sweat_smile:

This PR is simply to hide the icon until we actually know the status.

**Before:**
old

**After:**
new


taylorotwell:
Thanks

PR #430, opened 2018-10-31
[1.0] Replace resolve function calls

driesvints:


etiennellipse:
@driesvints Thanks, you're amazingly fast.

PR #431, opened 2018-10-31
[1.0] Build assets for production.

mvdnbrk:
This [PR](https://github.com/laravel/horizon/pull/428) by @MaxGiting made a change to `Dashboard.vue`.

This PR built the assets by running `npm run production`.


browner12:
I think we should require that maintainers build and commit assets. It would be too easy for a malicious user to submit a PR with malicious code and get it merged, since no one is going to review all that compiled code.

So my 👎 is not because I think this is a bad idea, just because I would prefer a maintainer do it.

mvdnbrk:
Ok no prob. You've got a point 👍

taylorotwell:
Will do.

PR #443, opened 2018-11-21
[2.0] Add flush command

rubenvanassche:
At the moment it is not possible to easily flush all the jobs in Horizon, this PR provides this functionality by providing following command:

```
php artisan horizon:flush
```

This feature is non-breaking.

### DIFF
--- a/src/Console/FlushCommand.php
+++ b/src/Console/FlushCommand.php
@@ -3,10 +3,13 @@
 namespace Laravel\Horizon\Console;
 
 use Illuminate\Console\Command;
+use Illuminate\Console\ConfirmableTrait;
 use Illuminate\Contracts\Redis\Factory as RedisFactory;
 
 class FlushCommand extends Command
 {
+    use ConfirmableTrait;
+
     /**
      * The name and signature of the console command.
      *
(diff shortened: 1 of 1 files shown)


mfn:
I believe this is dangerous and should not be done (that way).

This flushes the entire redis database on that connection and it might be shared with other application doing other stuff and would remove _their_ keys too.

This multi-use of the redis database is explicitly supported via the `prefix` available in `config/horizon.php`.

Btw, this isn't a new problem: Laravels `cache:clear` command suffers the same problem but at least it's documented.

I still think this is too dangerous/unexpected with documentation and should be done in a way not breaking other applications.

freekmurze:
Maybe add the `ConfirmableTrait` to make it a bit more safe.

rubenvanassche:
@mfn I'm currently working on a version that will only remove keys in the Redis store with the prefix set in the configuration.

freekmurze:
@brendt Thought it's not that common, you can actually persist Redis data.

vesper8:
to this day this remains an impossibility for me.. doing a cache:clear never results in the job log being flushed. The only way to flush the horizon jobs seems to do a full redis flush using redis-cli flushall

PR #444, opened 2018-11-23
rename uri to path and add to config

michaelnguyen547:
rename uri to path so it is consistent with Telescope config option, added path option to horizon config


driesvints:
We can't update this in the current release.

PR #466, opened 2019-01-13
[2.0] Use router-link on recent jobs's failed jobs link

swalker2:
So, because of some limitations on my production server I had to use this horizon.uri config option and got it working after some tweaking.

This PR fixes one link that wasn't using router-link and was not working for me.


swalker2:
Are you not using router-link anywhere? I am not sure why this can't be merged

driesvints:
@swalker2 Ah sorry, seems I was mistaken. We do indeed use it.

Issue #371, opened 2018-07-20
Problems with timestamps (NaNs, Invalid date)


iShotFT:
On both staging (centos) and local (ubuntu) environment I'm having issues with the timestamps.
- On the page 'Recent Jobs' I'm always getting Runtime 'NaNs'. 
- On the page 'Failed' I'm getting 'Invalid date' for 'Failed At' and 'NaNs' for 'Runtime'.

This combined with experiencing issue #366  is giving me headaches.

cmizzi:
Same bug here... I don't have any timestamp (NaNs values) on all queued (recent and failed) jobs.

davidruigrok:
Same issue. I think the problems occurs because horizon api returns failed_at and reserved_at as string (not as float) which can not be parsed within the vue component (result: _NaNs_ and _Invalid date_) :
```
{{ job.failed_at ? String((job.failed_at - job.reserved_at).toFixed(3))+'s' : '-' }}
```
```
{{ readableTimestamp(job.failed_at) }}
``` 
Part of API response (/horizon/api/jobs/failed):
```json
{"jobs": [
{
"id": "808121",
"connection": "redis",
"queue": "title:resources",
"name": "App\\Jobs\\TrackTitle",
"status": "failed",
"payload": {},
"exception": "...",
"failed_at": "1535147607,7085",
"completed_at": null,
"retried_by": [],
"reserved_at": "1535147607,6773",
"index": 0
}
]}
```
I have not yet figured out why failed_at and reserved_at are not set / parsed correctly in the api response. Same problems occurs with /horizon/api/jobs/recent and horizon/monitoring (detail page) API requests.

webpatser:
I had the same some time ago. Republishing the assets did the trick then..?

davidruigrok:
Unfortunately, republishing the assets did not solve it. 
Somehow, failed_at, completed_at, and reserved_at are stored as strings instead of floats.
Within redis (for _random_ job with id 1213931):
```
127.0.0.1:6379> HMGET horizon:1213931 id connection status exception failed_at completed_at retried_by reserved_at
 1) "1213931"
 2) "redis"
 3) "completed"
 4) (nil)
 5) (nil)
 6) "1538225702,0793"
 7) (nil)
 8) "1538225699,2691"
```
[...]

PR #477, opened 2019-01-24
[2.0] Fix storing floats in Redis

driesvints:
Redis will only accept floats with a point decimal sign and not a comma decimal sign. The sign is determined by the system's locale setting (LC_NUMERIC) and if set to one that prefers a comma will cause Redis to store the float as a string instead of a comma.

This change will make sure that before a float is stored it's always made sure that any comma is changed to a point. This was already done in some other parts of the system.

Fixes https://github.com/laravel/horizon/issues/371


driesvints:
@taylorotwell fixed

Issue #340, opened 2018-05-11
RedisJobRepository::findFailed() includes successful jobs


garando:
What I'm trying to implement: I want to dispatch job only if another job with the same tag isn't running already. Unfortunately, I haven't found such method in this framework, so I'm trying to implement it myself using Horizon's services. I would be glad if you point me how I am supposed to do it the right way.

A bug I found: I took job IDs from TagRepository and now I'm trying to check if this job is failed. I supposed that RedisJobRepository::findFailed() will return job only if it's failed and return null otherwise. However, right now it seems to just return job by ID regardless of its status.

driesvints:
I believe you're right and that this indeed isn't correct. I'll try to have a look at a later time to see if this can be fixed.

PR #478, opened 2019-01-24
[2.0] Fix findFailed method

driesvints:
The findFailed method was returning all kind of results, not just failed jobs

Fixes https://github.com/laravel/horizon/issues/340


driesvints:
This might conflict if you merge the other PR because I've added the same test class here. I'll rebase once you do.

Issue #480, opened 2019-01-27
Wrong number of processes in Current Workload card


fgilio:
Just as the title says, I see a wrong number of processes for each queue on the Current Workload card. The data from the API is correct though. Some caps:
<img width="1680" alt="screen shot 2019-01-27 at 12 02 49" src="https://user-images.githubusercontent.com/6857732/51802801-28355500-222c-11e9-819a-3adee365f983.png">
<img width="1679" alt="screen shot 2019-01-27 at 12 02 57" src="https://user-images.githubusercontent.com/6857732/51802802-28355500-222c-11e9-8645-5b92b7a4a1d7.png">

driesvints:
It's correct I believe? 6x2 = 12?

fgilio:
I mean the data from the API is correct, it shows that in that moment each queue had 2 processes assigned. But in the UI all queues are shown as having 12 processes.
This was working ok the last time I checked with Horizon 1.4.3, I'm now using 2.0.3.

driesvints:
Aha, I see what you mean. I'll take a look at this soon.

fgilio:
Thanks! And, of course, tell me if I can help with debugging or if you need any more info.

PR #481, opened 2019-01-28
[2.0] Fix incorrect processes count

driesvints:
On the Current Workload page, there was an incorrect reference to the stats object which caused an incorrect count to be shown.

Fixes https://github.com/laravel/horizon/issues/480


fgilio:
Thanks!

halaei:
When will it be tagged? Thanks.

PR #483, opened 2019-01-29
Let user choose custom env

setkyar:
I have 2 Horizon servers which handle different `supervisor`. For example, currently I have to use like the following and I need to update my `.env` file `environment` as `workerOne` and `workerTwo` separately. 

```php
'workerOne' => [
    'supervisor-1' => [
        'connection' => 'redis',
        'queue' => ['default'],
        'balance' => 'simple',
        'processes' => 10,
        'tries' => 3,
    ],
]
'workerTwo' => [
    'supervisor-2' => [
        'connection' => 'redis',
        'queue' => ['default'],
        'balance' => 'simple',
        'processes' => 10,
        'tries' => 3,
    ],
]
```

What happened was some worker check `.env` environment as well. So, I got environment sharing problems. Can we use separately? If that's open for PR. I would submit.

### DIFF
--- a/src/Console/HorizonCommand.php
+++ b/src/Console/HorizonCommand.php
@@ -48,7 +48,7 @@ class HorizonCommand extends Command
         });
 
         ProvisioningPlan::get(MasterSupervisor::name())->deploy(
-            config('app.env')
+            config('horizon.env') ?? config('app.env')
         );
 
         $this->info('Horizon started successfully.');


rudiedirkx:
This is a secret config without a default. Usually all the possible config is in the config file. Maybe add a `'env' => null` default to [config/horizon.php](https://github.com/laravel/horizon/blob/3.0/config/horizon.php) so people will know this option exists?

Issue #333, opened 2018-05-03
Job trimming for monitored tags


antons-:
From what I can tell, when you monitor a tag in Horizon it does so indefinitely until you stop monitoring that tag or your Redis instance runs out of ram.

We would like to keep monitoring tags over an extended period of time, but have a way of configuring job trimming for monitored tags, just like with recent and failed jobs.

Ideally I would like to trim jobs older than a week automatically with monitored tags.

Is this already possible or is there a different approach here I can take?

driesvints:
Could be a nice feature to add an input box with the desired minutes on how long to monitor a tag. I'll mark this as a feature request.

mfn:
I believe this is (or: related to) https://github.com/laravel/horizon/issues/271

driesvints:
@mfn @fgilio do you believe this can solve #271?

fgilio:
Yes, I think it can partially solve that issue.
What I mean with partially is that it will depend on the amount of jobs. If a system has 1000 jobs per minute then the Redis db could be filled in just an hour. But this may be solved by just making the value configurable and stating this fact clearly in the docs, so every dev chooses an appropriate value.

mfn:
@driesvints I second what @fgilio said above: depending on the amount of activity there seems to be always a possibility to "eat too much ram".

I guess the only real solution would be an implementation of an RRD which usually have fixed sizes, but I digress.

driesvints:
I've sent in a PR for this: https://github.com/laravel/horizon/pull/484

About the problem of speedy fill ups: I think you'll probably have bigger problems in that case and it doesn't really makes sense to monitor anyway if your DB can't keep up.

driesvints:
@mfn @fgilio can you both have a look at https://github.com/laravel/horizon/pull/484 and let me know on the PR itself if it looks ok to you?

Issue #271, opened 2018-01-04
Redis database grew too big, causing general Redis troubles


mfn:
Today we had the following issue:
- our Redis clients couldn't sporadically connect to Redis itself
- upon inspection we found out the redis database holding horizon had > 1 million keys
- it's size accounted to ~960MB (ie. 99%) of our whole Redis database

Errors we received were:
- `Connection closed`
- `Redis::pconnect(): connect() failed: Connection timed out`
- `RedisException: read error on connection`

(note: this error were recorded from non-Laravel PHP based applications; i.e. as explained the below, the Horizon database size seemed to affect Redis as a whole).

What we did:
- stopped horizon via supervisor
- ran `flushdb` in the horizon database (**Achtung:** be sure you've selected the right database)
  Running that command took 40 seconds
- started horizon via supervisor

In our case the Horizon Redis database was number `2`, here's the output from redis internal `info` command:
```
db2:keys=1192438,expires=22,avg_ttl=89314947
```
The `avg_ttl` looks suspiciously high.

Usually when inspecting such problems, I use the `keys *` command but I didn't dare to run it as it would _block redis completely_ during running it and we couldn't do this in production.

As such, at this time we don't have any information what keys where in there.

We can definitely exclude other applications having written to the same database; it's exclusively used by Horizon.

Our configuration:
- 13 queues
- 38 workers
- ~60 jobs per Minute
- we've (had) enabled monitoring for each job (each job has a default queue)
- `horizon:snapshot` running every 5 minutes

Does anyone have a clue what could cause this?
We were approximately running this since 2 months in production now.

mfn:
After this "reset" yesterday, todays `INFO` output for that database looks much saner:
```
db2:keys=4312,expires=4265,avg_ttl=1648948
```

When I look at the output from yesterday again, something feels very _off_ here: `keys=1192438,expires=22` => Only 22 keys expected to expire, from over 1Million (If I read that output correct).

mfn:
Todays `INFO` output:
```
db2:keys=889,expires=848,avg_ttl=1530403
```
I'm going to watch this a few days.

One thing I remember we did before the problem: we've enabled lots of "Monitor Tags". Basically one for every job-type we have.

We didnd't do this yet after we purged the database.

Can this be connected?

fgilio:
[...]

PR #484, opened 2019-01-29
[2.0] Expire monitored jobs

driesvints:
This commit fixes the problem where completed jobs which were monitored by a tag are stored indefinitely and end up filling up the DB and cause the system to run out of memory. The main problem at hand was that these jobs were never cleaned up. By adding the max TTL and TrimMonitoredJobs listener the monitored jobs will be cleaned up after the given amount of time. By default this is set to 7 days. The listener itself will run one time each day.

This might be breaking because of the new method on the `JobRepository` interface. If you want me to send this to master instead please let me know. This would at the very least require a point release.

Fixes https://github.com/laravel/horizon/issues/333 https://github.com/laravel/horizon/issues/271

### DIFF
--- a/config/horizon.php
+++ b/config/horizon.php
@@ -70,6 +70,7 @@ return [
     'trim' => [
         'recent' => 60,
         'failed' => 10080,
+        'monitored' => 10080,
     ],
 
     /*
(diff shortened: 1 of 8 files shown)


fgilio:
This looks ok to me and I'd use it right away.
I do agree with you @driesvints on this https://github.com/laravel/horizon/issues/333#issuecomment-458590636

PR #485, opened 2019-01-29
[2.0] Display error when posix_kill fails

driesvints:
At the moment when the posix_kill call for various commands fails the developer is left in the dark as to why this might be. By displaying the error it could give them an indication where to look for (file permissions, etc).

### DIFF
--- a/src/Console/ContinueCommand.php
+++ b/src/Console/ContinueCommand.php
@@ -39,7 +39,9 @@ class ContinueCommand extends Command
         foreach (array_pluck($masters, 'pid') as $processId) {
             $this->info("Sending CONT Signal To Process: {$processId}");
 
-            posix_kill($processId, SIGCONT);
+            if (! posix_kill($processId, SIGCONT)) {
+                $this->error("Failed to kill process: {$processId} (".posix_strerror(posix_get_last_error()).')');
+            }
         }
     }
 }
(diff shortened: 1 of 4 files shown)


sebastiaanluca:
```shell
$ php artisan horizon:terminate
Sending TERM Signal To Process: 1314
Failed to kill process: 1314 (Operation not permitted)
```

Works like a charm. Does `$this->error()` only print an error message or does it also exit the command with an error code? In some way, the command and thus the deployment should fail when e.g. terminating Horizon fails.

Thanks for the quick PR!

driesvints:
@sebastiaanluca no it doesn't. Imo the messages should be informative and not stop the process altogether. Especially not since it's in a loop and it's meant to kill multiple processes at the same time.

No problem!

driesvints:
I'm a bit afraid that it'll cause random problems suddenly. E.g. deployments not happening anymore because the behavior was changed for this command.

sebastiaanluca:
Good point, but on the other hand they might then finally know if their deployment process is even valid. If any of those commands previously failed, they weren't notified and maybe thought all was ok. If they now pass, nothing should halt. If they fail, they know and can adapt.

Perhaps this extension on this PR can be released in a new version as a breaking change? Maybe even as an exception being thrown so that if a command is executed in code (e.g. as a scheduled cron job but wrapped in a try/catch), it can be (optionally) handled. For the record, I do appreciate the effort and change, but there's not much that can be done with just an error message (no exit code handling, no try/catch, …).

driesvints:
@sebastiaanluca could you perhaps open up a new issue for this where we can discuss this further?

PR #495, opened 2019-01-31
[2.0] Refactor handle() methods.

mvdnbrk:
This PR refactors the handle method of the following commands:
- ContinueCommand
- PauseCommand
- TerminateCommand

It removes the temporary `$masters` variable by chaining the methods on the `Collection` class.

This removes the use of  the `array_pluck` helper method as well, which will be deprecated in Laravel 5.8.

### DIFF
--- a/src/Console/ContinueCommand.php
+++ b/src/Console/ContinueCommand.php
@@ -31,16 +31,17 @@ class ContinueCommand extends Command
      */
     public function handle(MasterSupervisorRepository $masters)
     {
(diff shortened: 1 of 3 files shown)


driesvints:
I wouldn't do this. There's currently a discussion going on of maybe letting these commands fail in 3.0 based on the `posix_kill` result and this refactor makes that impossible because those handles are moved to a closure.

Array pluck isn't being deprecated btw, just the helper. You can simply reference the class itself `Arr:pluck`.

PR #500, opened 2019-02-01
[2.0] Code style fixes

mvdnbrk:
Did my best to make the code base a bit more readable and consistent 😃

### DIFF
--- a/resources/js/app.js
+++ b/resources/js/app.js
@@ -1,6 +1,6 @@
 import Vue from 'vue';
 import _ from 'lodash';
-import axios from 'axios'
+import axios from 'axios';
 import moment from 'moment';
 import router from './router';
 import App from './components/App.vue';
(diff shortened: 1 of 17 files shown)


mfn:
What tool did you use to format it?

mvdnbrk:
@mfn Just used my eyes 😄

mfn:
Oh, k. Does that make even sense? Shouldn't this use something like prettier with a config file included in the repo?

mvdnbrk:
@mfn probably it should.

I gave it a try with prettier and a config file in the first place.
That gave some unexpected results. 
That's the reason why I went over all the files and fixed most of the changes by hand.

I am aware that it might not be perfect (yet). 
It is at least a start to get the code a bit more readable and consistent though.

taylorotwell:
Would prefer we do this using an actual tool with reproducable results in the future.

PR #502, opened 2019-02-05
[2.0] Fix jobs per minute over estimation

### LINKED ISSUE #501: Wrong ratio of jobs per minute over recent jobs
I have the following `horizon.trim` config:
```
    'trim' => [
        'recent' => 5,
        'failed' => 1000,
    ],
```

So the number of recent jobs should be around 5 times the number of jobs in the last minute, but it is not, the ratio is closer to 1 to 3 and always less than 1 to 4:

screenshot from 2019-02-04 13-06-53
screenshot from 2019-02-04 13-03-36

halaei:
Partially fixes #501 It doesn't fix underestimate during the first minute after each snapshot which is mostly because of the lack of enough data in that period

### DIFF
--- a/src/Repositories/RedisMetricsRepository.php
+++ b/src/Repositories/RedisMetricsRepository.php
@@ -334,9 +334,9 @@ class RedisMetricsRepository implements MetricsRepository
         $lastSnapshotAt = $this->connection()->get('last_snapshot_at')
                     ?: $this->storeSnapshotTimestamp();
 
-        return round(max(
+        return max(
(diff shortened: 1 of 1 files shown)


driesvints:
Can you add some tests which should the difference in behavior? Maybe add some screenshots of the dashboard which show the difference?

halaei:
Sure, in a day.

taylorotwell:
So this now returns a float?

halaei:
Yes this returns a float.
I added the screen shots to the issue. Let me know if it needs more tests.

Issue #458, opened 2019-01-03
Failed jobs not showing


shawnhind:
Recently on one of our environments the failed jobs page has stopped listing failed jobs.  The jobs are still being put into the failed jobs table in the database but when I go to the failed jobs page on Horizon it just says there aren't any failed jobs.

Recent and monitored tags still show up but failed jobs do not.

I've tested this by triggering a failed job and then checking to see it in the failed jobs table in the database and then loading the horizon page to see if it's there.

I've tried republishing the assets, restarting horizon on all 3 servers that it's running on and still haven't been able to get them to display.  Nothing in our configuration has changed recently and the failed jobs used to display on that page.  

Laravel version is 5.6
PHP version 7.2

driesvints:
Hi there,

Looks like this is a question which can be asked on a support channel. Please only use this issue tracker for reporting bugs with the library. If you have a question on how to use functionality provided by this repo you can try one of the following channels:

- [Laracasts Forums](https://laracasts.com/discuss)
- [Laravel.io Forums](https://laravel.io/forum)
- [Discord](https://discordapp.com/invite/KxwQuKb)
- [Larachat](https://larachat.co)
- [IRC](https://webchat.freenode.net/?nick=laravelnewbie&channels=%23laravel&prompt=1)
- [StackOverflow](https://stackoverflow.com)

Issue #109, opened 2017-08-31
Failed jobs page doen't show the jobs


aglipanci:
When I navigate into the failed jobs page it only shows the loader and it doen't list any failed jobs even though there are 4 of them and they are retrieved by the API. 

<img width="1437" alt="horizon" src="https://user-images.githubusercontent.com/2675283/29928258-d2f5f4b2-8e68-11e7-963b-762475bdec51.png">

themsaid:
I can't replicate this issue, try to republish the assets.

aglipanci:
Just did that restarted the horizon and supervisord but it's still the same. Any other idea what I can try?

themsaid:
I'm not sure, looks like the axios callback isn't getting executed for some reason. Let's see if more people report this issue.

aglipanci:
Thanks, let's see. I have another question non-related to this one: While I have horizon running, should I keep my queue:work process running as it was before or horizon will take care of that too?

themsaid:
Horizon takes care of that for you :)

wells:
@themsaid I can confirm I'm getting the same result on an Ubuntu 16.04 server. Axios is at least making the request and getting a response with data, but the Vue component remains in a loading state. The oddest part, I have successfully seen a list of failed jobs on this particular horizon instance before.

After examining the redis cache (using medis), it appears that some jobs are stored in redis with an integer based id (i.e. 4678) while others are being stored with a random string id (i.e. fb3AGkybiXix7IvvQRJFmiWEAUD7Q1Ig). My theory is that the jobs with random string ids are the cause for this screen remaining in a loading state.

I've also noticed that not all jobs are added to the recent_jobs zset either. My theory is that this mixed identifier typing is the reason some jobs are not showing up there altogether.

I can confirm that I am only running horizon via supervisor. No queue daemons/workers are running for this application. 

Also in my use case, I'm actually running horizon for two separate apps, each connected to their own redis db (the first on 0 and the second on 1). After running this for a couple days, I'm pretty certain their is no overlap or collision between the horizon instances for these two separate applications.

themsaid:
Jobs with string ID are jobs pushed before you install horizon, horizon jobs are integer based. You should have made sure your queue is empty before switching to horizon.

wells:
Interesting. So, in my case I have a separate application (queue worker) that runs horizon on behalf of queues dispatched to by other application(s) (they do not currently include horizon as a dependency).

[...]

PR #511, opened 2019-02-11
[2.0] Fix Failed Jobs page showing no results when failed jobs do exist

derekmd:
Fixes: https://github.com/laravel/horizon/issues/476, https://github.com/laravel/horizon/issues/458, https://github.com/laravel/horizon/issues/109

I've seen the Failed Jobs section behave like this since the Horizon beta was released.

The Vue component expects its `jobs` prop to be an `Array` but the API endpoint can return an `object` in its JSON body. A filtered PHP associative array needs to be kept as 0-indexed.

https://github.com/laravel/horizon/blob/032959e87c862070fe053f3b01f30bd9e4227e8d/resources/js/pages/Failed/Index.vue#L71-L73

https://github.com/laravel/horizon/issues/476#issuecomment-462529737 has a longer explanation of what I found. I wasn't able to figure out how to recreate the edge case in a PHPUnit test. Do we have an explanation of where this empty payload is coming from when fetching a set of jobs? https://github.com/laravel/horizon/blob/6db1be6a4dde066328d6e2a14acb543435bb837c/src/Repositories/RedisJobRepository.php#L193-L197

Does Redis `ZRANGE` (called in `RedisJobRepository@getJobsByType()`) sometimes return keys that don't exist? [This test attempt](https://github.com/laravel/horizon/compare/2.0...derekmd:fix-failed-jobs-empty-results-test) didn't seem to indicate that.

### DIFF
--- a/src/Repositories/RedisJobRepository.php
+++ b/src/Repositories/RedisJobRepository.php
@@ -194,7 +194,7 @@ class RedisJobRepository implements JobRepository
             $job = is_array($job) ? array_values($job) : null;
 
             return is_array($job) && $job[0] !== null;
-        }), $indexFrom);
+        })->values(), $indexFrom);
     }
 
     /**


driesvints:
> Do we have an explanation of where this empty payload is coming from when fetching a set of jobs?

Not that I know off.

But this is indeed the correct solution. I just tested this and it indeed needs it to be an array at all times. Can you maybe whip up a test that always expects the collection to not be a key-pair value?

PR #514, opened 2019-02-18
[2.0] Added auto refresh interval for dashboard

tomswinkels:
There are difference issues about the interval. 

You can change the interval so that you can see faster what is going on.

The name "Auto-refresh interval" is the same as on Beanstalk.

image


driesvints:
@tomswinkels you might want to give more explanation because otherwise the pr will just be closed.

tomswinkels:
@driesvints there are difference issues about this. You can change the interval so that you can see faster what is going on.

The name is the same as on Beanstalk.

image

driesvints:
I meant that Taylor will close any issues without a clear description in the first comment.

tomswinkels:
Later we can add this to the other pages.

taylorotwell:
I would prefer us to just choose a good convention and have less configuration.

tomswinkels:
@taylorotwell But check the issues. Not everyone would the same. In my case i wan't to see it more live than now. In Beanstalk the default is 500ms.

PR #524, opened 2019-02-27
fixes deprecated way of passing string to Symfony Process

SmoDav:
This change introduces a factory to fix the deprecated way of creating a new Symfony `Process`. It gives preference to the `fromShellCommandline` method which was introduced in Symfony 4.2.


driesvints:
Heya, thanks but this has already been fixed in the upcoming 3.0 release.

PR #529, opened 2019-02-28
[3.0] Add matrix to travis to tests against all supported version.

crynobone:


driesvints:
So what I've been thinking to do is to remove the nightly builds for stable Laravel releases. We do a new minor release every week so if anything's off we should catch it fairly soon. This isn't 100% fail proof of course but at least it takes quite a chunk of build time away.

PR #531, opened 2019-02-28
[4.0] Suggest predis instead of requiring it

ThomHurks:
Follow-up to PR #381 

Just like laravel/framework which suggests predis (and people can choose to use phpredis instead), Horizon should suggest it so people using phpredis do not needlessly install a dependency. Horizon relies on the redis config in config/database.php anyway, and trusts that the user correctly fills in that information. Requiring the user to install phpredis or predis herself is no different from what other parts of Laravel using Redis already do.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
         "predis/predis": "^1.1"
     },
     "suggest": {
-        "ext-redis": "Required to use the redis cache and queue drivers (*).",
+        "ext-redis": "Required to use the redis cache and queue drivers.",
         "predis/predis": "Required to use the redis cache and queue drivers (^1.1)."
     },
     "autoload": {
--- a/composer.json
+++ b/composer.json
@@ -26,9 +26,11 @@
         "mockery/mockery": "^1.0",
         "orchestra/database": "^3.7",
         "orchestra/testbench": "^3.7",
-        "phpunit/phpunit": "^7.0"
+        "phpunit/phpunit": "^7.0",
+        "predis/predis": "^1.1"
     },
     "suggest": {
+        "ext-redis": "Required to use the redis cache and queue drivers (*).",
         "predis/predis": "Required to use the redis cache and queue drivers (^1.1)."
     },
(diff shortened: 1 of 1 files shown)


taylorotwell:
All tests are going to fail unless you add it to require-dev.

driesvints:
You'll need to add it to the dev dependencies as well because the tests are failing.

Maybe we can also add phpredis to the suggest list so it's clear you can either use one or the other? We'll also need to update the install instructions after the release so people know to require it predis.

composer.json
```diff
@@ -27,7 +26,12 @@
         "mockery/mockery": "^1.0",
         "orchestra/database": "^3.7",
         "orchestra/testbench": "^3.7",
-        "phpunit/phpunit": "^7.0"
+        "phpunit/phpunit": "^7.0",
+        "predis/predis": "^1.1"
+    },
+    "suggest": {
+        "ext-redis": "Required to use the redis cache and queue drivers (*).",
```
driesvints: Is the version correct here? Don't we need a specific version? I don't know if Horizon is compatible with every single version of phpredis.

ThomHurks: I used the composer.json for laravel/framework as an example. It requires extensions like this:
```
"ext-json": "*",
"ext-mbstring": "*",
"ext-openssl": "*",
```
and suggests two like this:
```
"ext-pcntl": "Required to use all features of the queue worker.",
"ext-posix": "Required to use all features of the queue worker.",
```
I'm not sure about possible compatibility issues.

driesvints: Let's perhaps leave the ` (*)` away from here in that case.

ThomHurks: Will do.

driesvints:
Restarted build.

ThomHurks:
Thanks for the feedback @driesvints and @taylorotwell

ThomHurks:
Thanks @taylorotwell!

driesvints:
Thanks for sending this in :)

PR #539, opened 2019-03-05
[3.0] Prevent A non-numeric value encountered

MasterRO94:
At the moment strict comparison with null causes an error because frontend sends an empty string and it passes as `$afterIndex` parameter. After that executes `$afterIndex + 1` and php >= 7.1 generates an `A non-numeric value encountered` E_WARNING


themsaid:
Hey,

Frontend is never sending an empty string, which screen is that?

Can you please open an issue with steps of replication or more details about the issue? From my testing the frontend always sends an index or `-1` in case if it's the very first screen, but never an empty string.

PR #542, opened 2019-03-05
Update webpack.mix.js

papac:
Undefined variable `path`. The `path` module is not correctly loading.


themsaid:
That's not needed, `path` is already loaded by Mix.

Issue #468, opened 2019-01-14
Shell live check if horizon is running


loic-lopez:
Hello,

It is possible to check if horizon is running by running a command in a shell?
I need this feature to implement a readinessProbe and livenessProbe on Kubernetes.

Thanks for reply :)

mfn:
### Using shell
You mean something like using `ps`?

Not sure how accurate this can be, but there's only one uber process controlling everything:
```
           |                   `-php(9575)-+-php7.2(9721)---php7.2(9908)
           |                               |-php7.2(9722)---php7.2(9912)
           |                               |-php7.2(9723)---php7.2(9917)
           |                               |-php7.2(9724)---php7.2(9918)
           |                               |-php7.2(9725)---php7.2(9907)
           |                               |-php7.2(9726)---php7.2(9915)
           |                               |-php7.2(9727)---php7.2(9909)
           |                               |-php7.2(9728)---php7.2(9910)
           |                               |-php7.2(9729)---php7.2(9911)
           |                               |-php7.2(9735)---php7.2(9906)
           |                               |-php7.2(9736)---php7.2(9914)
           |                               |-php7.2(9738)---php7.2(9916)
           |                               `-php7.2(9739)---php7.2(9913)
```
The one with `9575` looks like this:
```
$ ps auxw|grep 9575
vagrant   9575  0.0  0.8 360060 51864 ?        S    Jan13   1:20 php /vagrant/artisan horizon
```
So it would be possible to use something like:
```
$ ps auxw|grep 'artisan horizon$'
vagrant   9575  0.0  0.8 360060 51864 ?        S    Jan13   1:20 php /vagrant/artisan horizon
```
### By inspecting Redis
Alternative you could also directly _look_ into redis, e.g.
```
127.0.0.1:6379[2]> ZRANGE horizon:masters 0 -1
1) "swatio-dev-xYjt"
```
gives you a list of currently running masters

Though I can understand the appeal to check the process table directly in a Pod?

loic-lopez:
I think it would be a good feature if we can 'ping' horizon worker or something like that

mfn:
Makes sense to me!

The closest there's right now would be `./artisan horizon:supervisors` but having a definite way to check if the "master" process is still running would be great.

driesvints:
@mfn could this be implemented in the library itself? Probably needs a dedicated endpoint?

ohnotnow:
At the moment I'm doing something hacky along the lines of :
```
[...]

PR #545, opened 2019-03-07
[3.0] Adding horizon:status command

dbpolito:
Adding `horizon:status` command to be able to check via cli the status of Horizon.

### DIFF
--- a/src/Console/StatusCommand.php
+++ b/src/Console/StatusCommand.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace Laravel\Horizon\Console;
+
+use Illuminate\Console\Command;
+use Laravel\Horizon\Contracts\MasterSupervisorRepository;
+
+class StatusCommand extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'horizon:status';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Get the current status of Horizon.';
+
+    /**
+     * Execute the console command.
+     *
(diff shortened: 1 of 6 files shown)


driesvints:
@dbpolito I believe this is for https://github.com/laravel/horizon/issues/468 ?

It's always better to be very thorough in your PR description so it has a higher chance of being accepted.

dbpolito:
Yes, that's my use case, this will solve the problem too. Sorry, i didn't look for related tickets. I agree that would be better.

driesvints:
The addition of `public function currentStatus();` on the `MasterSupervisorRepository` interface is a breaking change in this release. I think it might be better if we add this in the next one. Didn't check this on the previous PR, sorry.

dbpolito:
Well, this is not required. I can duplicate the logic on the command for now so we can have it. It will be definitely useful to people, not sure this has to wait a new version.

I did that to reuse the logic.

driesvints:
Yeah, that might be better for now.

dbpolito:
Done

driesvints:
@dbpolito I've merged this into master. You can send the changes you did at first there if you want.

delmicio:
@dbpolito How can I use it programmatically without Artisan::call function?

dbpolito:
@delmicio that depends on your use case or what you want to do... in my case i'm using as readiness/livinessProbe on kubernetes with bash script

delmicio:
@dbpolito I see there is a protected function called currentStatus on this PR. Any chance to do what that function does on a laravel controller for example?

delmicio:
found it on another package if some is interested: 
https://github.com/antonioribeiro/health/blob/master/src/Checkers/Horizon.php

PR #547, opened 2019-03-10
[3.0] Fix URL for retrying jobs

aaronhuisinga:
Horizon 3.0 added a config option for defining the Horizon path. The route for retrying failed jobs is still hardcoded to the default "horizon" path.


driesvints:
Thanks!

Issue #549, opened 2019-03-17
Setting supervisor niceness


halaei:
I was thinking of sending a PR to set process niceness on each supervisor via config:
```php
            'supervisor-1' => [
                'connection' => 'redis',
                'queue' => ['default'],
                'balance' => 'simple',
                'processes' => 10,
                'tries' => 3,
                'nice' => 5,
            ],
```
Then in supervisor doing this:
```php
if ($config['nice']) proc_nice($config['nice']);
```
Do you think horizon's config is a good place to set process niceness? Or instead just calling proc_nice() in the jobs. Please note that niceness can only be increased in a process (unless maybe the process is run by the root user).

driesvints:
Can you maybe give some background info on what this does and how it will benefit users?

halaei:
In cases of high server load, when CPU usage is 100%, setting proc_nice can help OS to prioritize other processes that needs to be near-real time, e.g. PHP processes that handle web requests or maybe redis and mysql servers.

The benefit can be for example users won't experience lag in site just because some workers are processing some audio files, that are fine to be processes a few second or even hours later.

driesvints:
Pinging @taylorotwell to see if we might get this into Horizon.

driesvints:
@halaei you can send in a PR if you want :)

halaei:
OK thanks. I have issues with backward compatibility, because I can't get away from adding parameter to `SupervisorOptions::__construct()`

driesvints:
If you need to make breaking changes you can send to master

francislavoie:
@driesvints this can be closed now, can't it? The PR's been merged.

halaei:
Yes

Issue #170, opened 2017-09-30
Windows installation


nadinyamaui:
This is not really an issue the problem is the following

Many people like me develop their apps in windows platforms and the production servers runs in Forge with Ubuntu.

Since horizon requires PHP extensions that are not compatible with windows developers might get composer issues saying that ext-pcntl or ext-posix are missing so people are not able to install the package and they probably wont use horizon in their local development only on production environments

So I think we should mention this in the Laravel docs: That Horizon is not compatible with windows, however users can install it by running composer install with the option --ignore-platform-reqs

PS: I created a PR on Laravel docs for this (https://github.com/laravel/docs/pull/3686)

fancenl:
We run multiple workers on Windows machines (because of Adobe After Effect) and we need Horizon for monitoring our workers. Is there no other solution for running this on Windows?

And yes, I agree this should be mentioned in the Laravel docs.

arasmit2453:
In my tests there is no way to run horizon on windows there is a way to compile one of the dependencies but that didn't work for me.


________________________________

driesvints:
Hey everyone. We'd be open to deprecate usage of pcntl and posix and make Horizon work on Windows if anyone can ever send in a PR to the next major version which would solve this in a clean way.

tcope25:
@driesvints any update on getting horizon to work on windows?

driesvints:
@tcope25 nope but we're open to pull requests.

SavageDan:
Why not use exec("kill -9 $pid"), which is available natively and can be switched out with the relevant windows kill command, depending on context? This would come with the added benefit that process workers need to be owned by the www user. Or am I missing something?

driesvints:
At the moment we're not looking to implement Windows support for Horizon ourselves but we're open to PRs which implement this in a clean way. Although I have to say that the internals would need to be overhauled by a lot in order to do so.

woganmay:
I've just run into this problem myself now, and I didn't want to open a new issue (so that the above context doesn't get lost).

I mainly use Windows as my dev environment too, and I've just spent some time looking at the Horizon code. I get why it would be a big overhaul to get it to run in Windows, but it doesn't seem impossible:

PHP has a built-in process control library that should work on Windows, and make it possible to spawn/kill worker threads from a supervisor: https://www.php.net/manual/en/book.exec.php

[...]

PR #551, opened 2019-03-19
[3.0] Supervisor nice option

halaei:
Closing #549 
This PR increments process niceness on each supervisor via config:
```php
            'supervisor-1' => [
                'connection' => 'redis',
                'queue' => ['default'],
                'balance' => 'simple',
                'processes' => 10,
                'tries' => 3,
                'nice' => 5,
            ],
```
The supervisor calls `proc_nice(5)` to set niceness:

I don't know if you can regard it as backward compatible because of the change in supervisor option constructor. Also not sure if it has to be sent to 3.0 or master.

### DIFF
--- a/src/Console/SupervisorCommand.php
+++ b/src/Console/SupervisorCommand.php
@@ -23,6 +23,7 @@ class SupervisorCommand extends Command
                             {--max-processes=1 : The maximum number of total workers to start}
                             {--min-processes=1 : The minimum number of workers to assign per queue}
                             {--memory=128 : The memory limit in megabytes}
+                            {--nice=0 : The process niceness}
                             {--paused : Start the supervisor in a paused state}
                             {--queue= : The names of the queues to work}
                             {--sleep=3 : Number of seconds to sleep when no job is available}
@@ -74,6 +75,10 @@ class SupervisorCommand extends Command
      */
     protected function start($supervisor)
     {
+        if ($supervisor->options->nice > 0) {
+            proc_nice($supervisor->options->nice);
+        }
+
         $supervisor->handleOutputUsing(function ($type, $line) {
             $this->output->write($line);
         });
@@ -106,7 +111,8 @@ class SupervisorCommand extends Command
             $this->option('timeout'),
             $this->option('sleep'),
             $this->option('tries'),
-            $this->option('force')
+            $this->option('force'),
+            $this->option('nice')
         );
     }
 
--- a/src/Console/SupervisorCommand.php
+++ b/src/Console/SupervisorCommand.php
@@ -23,7 +23,7 @@ class SupervisorCommand extends Command
                             {--max-processes=1 : The maximum number of total workers to start}
                             {--min-processes=1 : The minimum number of workers to assign per queue}
                             {--memory=128 : The memory limit in megabytes}
-                            {--nice=0 : The process niceness}
+                            {--nice=0 : Increment to the process niceness}
                             {--paused : Start the supervisor in a paused state}
                             {--queue= : The names of the queues to work}
                             {--sleep=3 : Number of seconds to sleep when no job is available}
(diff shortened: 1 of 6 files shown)


driesvints:
It's backwards compatible because adding constructor arguments with defaults won't break anything. This is fine 👍

driesvints:
@halaei Also: always add the full description for this PR to your main comment and not just link to a issue/pull request.

halaei:
@driesvints Updated comment

halaei:
Reading code, I see a --paused option:

    {--paused : Start the supervisor in a paused state}

Reading that, I wish there was an option to pause a supervisor for a few seconds before starting to scale for the first time. maybe a `--pause=10` option.

taylorotwell:
The PHP documentation states that the `proc_nice` value can also be negative?

halaei:
@taylorotwell It will only work if the process is running as a superuser, otherwise an exception will be raised with this message:

    proc_nice(): Only a super user may attempt to increase the priority of a process

browner12:
Support for Windows does not exist until PHP 7.2. Do we need to do anything in code to account for that, or will that be a documentation solution?

halaei:
@browner12 This is an optional feature. If it is not supported on an environment you can simply not use it.

driesvints:
@browner12 Windows support is lacking in general: https://github.com/laravel/horizon/issues/170

src/SupervisorOptions.php
```diff
@@ -48,6 +48,13 @@ class SupervisorOptions extends WorkerOptions
      */
     public $minProcesses = 1;
 
+    /**
+     * The process niceness.
+     *
+     * @var int
+     */
+    public $nice = 0;
```
mfn: This is problematic:
- a niceness of `0` is a valid value
- the (linux) default for niceness is usually `10`

IMHO the correct way is to use `null` here.

src/Console/SupervisorCommand.php
```diff
@@ -23,6 +23,7 @@ class SupervisorCommand extends Command
                             {--max-processes=1 : The maximum number of total workers to start}
                             {--min-processes=1 : The minimum number of workers to assign per queue}
                             {--memory=128 : The memory limit in megabytes}
+                            {--nice=0 : The process niceness}
```
mfn: `0` is not a good value, see my other comment

src/Console/SupervisorCommand.php
```diff
@@ -74,6 +75,10 @@ public function handle(SupervisorFactory $factory)
      */
     protected function start($supervisor)
     {
+        if ($supervisor->options->nice > 0) {
```
mfn: - `0` is a valid niceness level
- so are negative values

This check, see my other comment, should be `!== null`

It doesn't matter that e.g. negative values are only allowed for root user rights. Don't outsmart the dev who might have run as root _for whatever_ reason, etc.

mfn (changes requested):
No happy about the `0` default and not allowing negative values!

halaei:
@mfn The argument of proc_nice is the amount of **increment** to the niceness:

>  proc_nice ( int $increment ) : bool
proc_nice() changes the priority of the current process by the amount specified in increment.

http://php.net/manual/en/function.proc-nice.php

halaei:
@mfn zero increment means no increment, so no need for null. About letting user to use negative values, that is fine for me, although I don't feel I should use it or run my workers as sudo or use Windows at all ;)

halaei:
Maybe I should change the comments in the code from 'The process niceness' to 'Increment to the process niceness'?

mfn:
@halaei thanks for correcting me, you're right!

Still, I vote for allowing negative values. It's not your decision how others want to run the code.

driesvints:
This has now been released in v3.1.0

einenlum:
This doesn't seem documented: https://laravel.com/docs/11.x/horizon
Could it be added to the documentation somewhere?

Issue #266, opened 2017-12-19
Support for websockets as an alternative to client-side polling


francislavoie:
The horizon dashboard would probably a lot lighter if there it could be configured to listen for broadcasted events rather than polling constantly for new data. This should probably be a configurable option, defaulting to polling (since not every application has websockets set up). AFAICT, this should be reasonably easy to implement using https://laravel.com/docs/5.5/broadcasting

jonnywilliamson:
And now that https://github.com/beyondcode/laravel-websockets has been release... ! :)

MithrandirDK:
The idea is good, but since broadcasting uses queues... won't that make for an endless loop?
1. Job is done
2. Send broadcast
3. Queue broadcast job
4. Run broadcast job
5. Broadcast job is done
6. Go to 2.

francislavoie:
Ahahaha that's an interesting thought!

I think we could avoid that by instead pushing scheduled bundled updates every x seconds instead of pushing on every job completion. That would be way too many websockets events on production systems, most likely. Or instead, just request an update over websockets every x seconds, i.e. still polling, but doing so over the same connection which is much lighter.

driesvints:
Although I like this idea I think this would require quite a bit of work in order to pull this off properly. That's why we won't be working on this ourselves. If anyone can send in a clean PR for this we could consider it.

PR #558, opened 2019-03-26
[3.0] Change refresh interval

tomswinkels:
Change the refresh interval of stats on dashboard to 500ms. 

The default is now the same as the default in Beanstalk.


driesvints:
I believe this would be even more problematic for people with bad internet connections. This would increase their load ten fold.

I also still believe that making for of #266 is a better way to go.

taylorotwell:
No plans to merge at this time.

PR #570, opened 2019-04-08
[3.0] Serve content from https when site uses https as protocol

darki73:
Right now, when site is accessed through **_https://_**, content is still served via **_http://_**, and it doesn't matter if APP_URL is `https://site_name` or `http://site_name`.

With this pull request, assets will be served correctly based on the **_APP_URL_** variable, if it has **_https://_** in it, the `secure_asset` helper will be used


darki73:
The main reason for this **PR** is the usage of **Horizon Gate** and production website

driesvints:
This should be done automatically for you actually. The `formatRoot` method in `UrlGenerator` actually does this very thing internally.

Issue #586, opened 2019-05-01
More insights in to running jobs/processes


AJenbo:
At my place of work we are using Horizon to monitor our queues one thing we found lacking is that there is very little insights in to what job is currently being run and how much system resources it is taking.

To solve this we are working on extending Horizon to give us this information and some grater control over what is running, once it's working we would be happy to contribute it back to Laravel Horizon. To that end we would like to know if you would be interested in this, and also what your opinions are regarding how this should be done?

Some of the planned features:

**Phase 1:**
- Add "Running" status to Recent Jobs
- Show current CPU/mem load for a job/process
- Show current CPU/mem load for a scheduler

**Phase 2:**
- Terminate a running job (UI similar to Retry for failed jobs)
- Detect job failure via job<->process association instead of waiting for the timeout

brunogaspar:
I would be interested in some of the information you plan to add :)

Have been wanting to add these and then submit a pull request, but the lack of time have been preventing me to do so, so i would appreciate it!

driesvints:
Feel free to send in PRs if you want!

driesvints:
We've reverted the original PR that implemented this and probably won't be reconsidering this feature again. Please see my answer here: https://github.com/laravel/horizon/pull/589#issuecomment-502085735

PR #588, opened 2019-05-07
Display worker CPU and memory utilization in supervisor list

AJenbo:
So here is my first PR in relation to #586

This adds information about how much system resources (CPU and memory) the workers are currently consuming:
image

I have tried to communicate this in the most user friendly way possible, but here are some additional details:
 - CPU Threads shows how many logical CPU cores the workers are currently occupying, if the number is close to the actual number of logical CPU cores then the system is probably being maxed out by the workers. Other process will also take resources, so this is not meant to show you system load but rather how much is being used by the workers. As such if the server is doing other tasks it might not be a replacement for other devops tools.
- Memory, this shows how much memory is currently being used by the workers. This is shown as a percentage of total system memory.


driesvints:
Heya, thank you for working on this! I'm going to have to ask you to resubmit however without the compiled assets. We don't allow this for reasons stated here: https://laravel.com/docs/5.8/contributions#compiled-assets

Make sure that you don't have any commits added with compiled assets and feel free to resubmit. Thanks! :)

driesvints:
Also: if I'm correct I don't immediately see any breaking changes so you may send this to the latest active branch (3.0).

AJenbo:
Ok, no problem I'm not used to dealing with compiled assets :)

I thought this might be considered a new feature so that's why I targeted master, but cool to hear this is small enough to go in to the latest branch.

I don't know if it's against your normal work flow, but if you prefer I can also just update an existing PR instead of creating a new PR.

driesvints:
The new PR is fine.

Most of Laravel's libraries (except the framework and Lumen) are SemVer so new features can still be added on a minor release.

PR #589, opened 2019-05-07
[3.0] Display worker CPU and memory utilization in supervisor list

AJenbo:
So here is my first PR in relation to #586

This adds information about how much system resources (CPU and memory) the workers are currently consuming:
image

I have tried to communicate this in the most user friendly way possible, but here are some additional details:
 - CPU Threads shows how many logical CPU cores the workers are currently occupying, if the number is close to the actual number of logical CPU cores then the system is probably being maxed out by the workers. Other process will also take resources, so this is not meant to show you system load but rather how much is being used by the workers. As such if the server is doing other tasks it might not be a replacement for other devops tools.
- Memory, this shows how much memory is currently being used by the workers. This is shown as a percentage of total system memory.

### DIFF
--- a/src/Repositories/RedisSupervisorRepository.php
+++ b/src/Repositories/RedisSupervisorRepository.php
@@ -71,7 +71,7 @@ class RedisSupervisorRepository implements SupervisorRepository
     {
         $records = $this->connection()->pipeline(function ($pipe) use ($names) {
             foreach ($names as $name) {
-                $pipe->hmget('supervisor:'.$name, ['name', 'master', 'pid', 'status', 'processes', 'options']);
+                $pipe->hmget('supervisor:'.$name, ['name', 'master', 'pid', 'status', 'processes', 'options', 'cpu', 'mem']);
             }
         });
 
@@ -85,6 +85,8 @@ class RedisSupervisorRepository implements SupervisorRepository
                 'status' => $record[3],
                 'processes' => json_decode($record[4], true),
                 'options' => json_decode($record[5], true),
+                'cpu' => $record[6],
+                'mem' => $record[7],
             ];
         })->filter()->all();
     }
@@ -114,6 +116,7 @@ class RedisSupervisorRepository implements SupervisorRepository
         })->toJson();
 
         $this->connection()->pipeline(function ($pipe) use ($supervisor, $processes) {
+            $workerStats = $supervisor->workerStats();
             $pipe->hmset(
                 'supervisor:'.$supervisor->name, [
                     'name' => $supervisor->name,
@@ -122,6 +125,8 @@ class RedisSupervisorRepository implements SupervisorRepository
                     'status' => $supervisor->working ? 'running' : 'paused',
                     'processes' => $processes,
                     'options' => $supervisor->options->toJson(),
+                    'cpu' => number_format($workerStats->sum('cpu'), 2),
+                    'mem' => round($workerStats->sum('mem')),
                 ]
             );
 
--- a/src/SystemProcessCounter.php
+++ b/src/SystemProcessCounter.php
@@ -14,6 +14,27 @@ class SystemProcessCounter
     public static $command = 'horizon:work';
 
     /**
+     * Get CPU and memory usage for each Horizon workers for a given supervisor.
+     *
+     * @param  string  $name
+     * @return \Illuminate\Support\Collection
+     */
+    public function getWorkerStats($name)
+    {
+        $process = Process::fromShellCommandline('exec ps axo %cpu,%mem,command | grep '.static::$command.' | grep "supervisor='.$name.'" | grep -v "exec ps axo"', null, ['COLUMNS' => '2000']);
+
+        $process->run();
+
+        $rows = explode("\n", $process->getOutput());
+
+        return collect($rows)->filter()->map(function ($rows) {
+            $row = collect(explode(' ', $rows))->filter()->take(2);
+
+            return ['cpu' => $row->first() / 100, 'mem' => $row->last()];
+        });
+    }
+
+    /**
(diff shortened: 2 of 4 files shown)


src/Repositories/RedisSupervisorRepository.php
```diff
@@ -71,7 +71,7 @@ public function get(array $names)
     {
         $records = $this->connection()->pipeline(function ($pipe) use ($names) {
             foreach ($names as $name) {
-                $pipe->hmget('supervisor:'.$name, ['name', 'master', 'pid', 'status', 'processes', 'options']);
+                $pipe->hmget('supervisor:'.$name, ['name', 'master', 'pid', 'cpu', 'mem', 'status', 'processes', 'options']);
```
driesvints: Hmm, I hadn't considered the sequence of the parameters here. It's probably best if you add them at the end if people rely on the sequence to determine the values (see the array keys below). If they're new values at the end of the array it's not breaking.

src/SystemProcessCounter.php
```diff
@@ -14,17 +14,34 @@ class SystemProcessCounter
...
      *
      * @param  string  $name
-     * @return int
+     * @return \Illuminate\Support\Collection
      */
-    public function get($name)
+    public function getWorkerStats($name)
     {
-        $process = Process::fromShellCommandline('exec ps aux | grep '.static::$command, null, ['COLUMNS' => '2000']);
+        $process = Process::fromShellCommandline('exec ps axo %cpu,%mem,cmd | grep '.static::$command.' | grep "supervisor='.$name.'" | grep -v "exec ps axo"', null, ['COLUMNS' => '2000']);
```
AJenbo: `u` is removed as it formats the put put for human readability.
`%cpu,%mem,cmd` are all in the default output but being explicit both makes it easier to parse and reduces the amount of unused data being processed.

src/Repositories/RedisSupervisorRepository.php
AJenbo: Fixed

src/SystemProcessCounter.php
bonzai: This will not work on macOS:

```shell
$ exec ps axo %cpu,%mem,cmd

ps: cmd: keyword not found
```

On macOS, you should use `command` keyword.

AJenbo: Thanks, I have changed it to `command` and tested that it still works on Linux.

taylorotwell:
This works on Windows, Mac, and Linux?

AJenbo:
Horizon 3.0 as is doesn't work on Windows, as it depends on `php_pcntl`, `ps` and a few other UNIX only features (Redis isn't officially supported for Windows either).

This change simply parses the output from `ps` and so doesn't add any new dependencies, the changes have been on both Linux and Mac.

The UI of cause displays correctly on Windows as before.

JayBizzle:
@driesvints I see you released `v3.2.0` but I think the JS needs rebuilding for this PR to work as the `resources/js/screens/dashboard.vue` file was changed 👍

driesvints:
@JayBizzle that's odd. I see @themsaid compiled it here: https://github.com/laravel/horizon/commit/8ee690a763bd4473398a4ff7a303eb5a5a712fdd

So it should be working afaik...

JayBizzle:
Weird, can't get it to work after updating....let me try again...

driesvints:
Nvm. Seems the pr was merged after it. I'll release a patch version with the compiled assets 👍

driesvints:
@JayBizzle released v3.2.1

JayBizzle:
Yep, just tested locally, had to rebuild the assets 👍

JayBizzle:
Updated to v3.2.1, all working as expected 👍

Cannonb4ll:
Great work @AJenbo, love this addition to Horizon 🔥

AJenbo:
Thanks ☺️, I'll work on adding some more goodies over the next few weeks

tomswinkels:
Very nice @AJenbo !

andreladocruz:
If the front end is running in a different machine from the queue processor one, the stats are doesn't show. 

Maybe saving the stats in redis should solve the issue. ;)

AJenbo:
I have only tested it on systems where the frontend is running in a different machine. And the results are saved in Redish. @andreladocruz Could you describe more accurately exactly what you are experiencing?

andreladocruz:
@AJenbo, we have our frontend running in google appengine, our queue processor running in a Computer Engine machine, and our redis is a Memorystore stack.

Just updated horizon to 3.2.1 version and can´t any information about cpu and memory usage:

image

That´s what I´m talking about. 

If the queue processor is in another server rather then the frontend one, no information is shown.

AJenbo:
Since the columns labels are entirely missing your browser is still displaying Horizon 3.2.0 or older, maybe you have some caching that hasn't been cleared after the deployment or your AppEngine wasn't updated properly. This has nothing to do with running on multiple servers.

driesvints:
Please remember to follow the upgrade guide and re-publish the assets when updating horizon: https://laravel.com/docs/5.8/horizon#upgrading

andreladocruz:
@AJenbo and @driesvints , thanks for reminding to publish the assets. =P 

Everything is working okay.

Sorry about that.

CyrilMazur:
I experience a performance hit (+5-6% CPU usage) when upgrading from 3.1.2 to 3.2.1. Anybody else? I opened an issue: https://github.com/laravel/horizon/issues/601

driesvints:
Since this PR was causing quite some bugs for others and also introduced a performance hit we've decided to revert it entirely. We also won't be re-considering a new PR anymore. We'd like to encourage you to use other tools to track and measure CPU/Memory consumption and just keep Horizon to its core features.

Issue #562, opened 2019-04-02
Queued At Time zone issue


xyrintech:
- Horizon Version: 3.1
- Laravel Version: 5.8
- PHP Version: 7.3
- Database Driver & Version: Redis, Postgresql 11

### Description:
Queued At is not considering the local browser time instead showing the date in some other time zone. The previous versions seems to be local browser time.

driesvints:
Have you used `php artisan horizon:assets` as said in https://laravel.com/docs/5.8/horizon#upgrading to update the front-end assets?

xyrintech:
I did use this - php artisan horizon:assets but the issue is still there. I believe its taking the timezone from app.php rather than local (browser) time. I tried changing the timezone in app.php from UTC to Los Angeles time and horizon starts showing up the time in Los Angeles time. Before this version the horizon always have been timezone in local time. 

There is an another issue with this, when we add a monitoring then on page it shows maximum 3 records. The recent jobs show 50 records and this is fine but monitoring tags shows only 3 records at max.

driesvints:
Server time has always been used for Horizon and I believe we won't change that any time soon.

> There is an another issue with this, when we add a monitoring then on page it shows maximum 3 records. The recent jobs show 50 records and this is fine but monitoring tags shows only 3 records at max.

Please create a separate issue for this.

Issue #151, opened 2017-09-21
Horizon does not respect timezone settings


KristofMorva:
I have timezone set in Laravel config, on the VPS and also in PHP config to UTC. Yet, RecentJobs' QueuedAt column still uses some kind of local timezone setting. While it's not necessarily important, it'd be great if it was somehow reliable, preferably dependent on Laravel's config.

themsaid:
Horizon displays time based on the browser timezone.

KristofMorva:
@themsaid thank you, that makes it clear. However, it may be a little misleading when you compare these times with logs, or other team members in a different timezone.

PR #594, opened 2019-05-15
[3.0] Allow UI timezone to be customized

kbond:
Ideally, I think the timezone should come from the browser like Nova/Telescope (or at least have the option) but [there seems to be a reason not to](https://github.com/laravel/horizon/issues/562#issuecomment-480788256). In fact, [it seems at one point it did come from the browser](https://github.com/laravel/horizon/issues/151#issuecomment-331892388) but this ability was removed.

### DIFF
--- a/src/Horizon.php
+++ b/src/Horizon.php
@@ -50,6 +50,13 @@ class Horizon
     public static $useDarkTheme = false;
 
     /**
+     * The timezone to use in the UI (defaults to app.timezone).
+     *
+     * @var string
+     */
+    public static $timezone;
+
+    /**
      * The database configuration methods.
      *
      * @var array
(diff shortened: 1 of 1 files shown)


themsaid:
Where does it not show based on browser timezone so we can check?

kbond:
Anywhere there is a timestamp being shown: recent jobs screen, failed jobs screen. [This is the function in question](https://github.com/laravel/horizon/blob/2289331408d592242a41fbfd58820529c04dc70e/resources/js/base.js#L83)

taylorotwell:
I think we planned to show it in the browser's timezone. Not sure if @themsaid has fixed that yet.

themsaid:
Horizon will be using browser time starting next release next Tuesday.

PR #606, opened 2019-06-03
[3.0] Properly format numeric values

bonzai:
Recently, Horizon introduced [worker CPU and memory utilization in supervisor list](https://github.com/laravel/horizon/pull/589), but there are locales that use a comma as a decimal separator. So, right now, you'll see a lot of exceptions in logs like this:

```
[0000-00-00 00:00:00] local.ERROR: A non well formed numeric value encountered {"exception":"[object] (ErrorException(code: 0): A non well formed numeric value encountered at /.../vendor/laravel/horizon/src/SystemProcessCounter.php:33)
```

because Horizon expects values like `0.25` instead of `0,25`.

### DIFF
--- a/src/SystemProcessCounter.php
+++ b/src/SystemProcessCounter.php
@@ -28,7 +28,9 @@ class SystemProcessCounter
         $rows = explode("\n", $process->getOutput());
 
         return collect($rows)->filter()->map(function ($rows) {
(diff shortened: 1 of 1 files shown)


driesvints:
I've reverted this PR because the original PR that caused this bug (https://github.com/laravel/horizon/pull/589) was reverted.

PR #608, opened 2019-06-05
[3.0] Better page titels

tomswinkels:
Since the last release we have added the app name in the title and header but not on every page.

Now i have change all the titles to the same structure as the header.


driesvints:
Unfortunately these aren't blade views so this won't work...

PR #609, opened 2019-06-05
[3.0] Setting the default queue in the config to use the base laravel queue config env variables

jdavidbakr:
When I install horizon I always have to (and frequently forget to) set the queue to match the redis queue setting in my Laravel project.  This pull request set the default config/horizon.php queues to prefer the REDIS_QUEUE environment variable instead of being hard-coded as 'default.'

### DIFF
--- a/config/horizon.php
+++ b/config/horizon.php
@@ -142,7 +142,7 @@ return [
         'production' => [
             'supervisor-1' => [
                 'connection' => 'redis',
-                'queue' => ['default'],
+                'queue' => [env('REDIS_QUEUE', 'default')],
                 'balance' => 'simple',
                 'processes' => 10,
                 'tries' => 3,
@@ -152,7 +152,7 @@ return [
         'local' => [
             'supervisor-1' => [
                 'connection' => 'redis',
-                'queue' => ['default'],
+                'queue' => [env('REDIS_QUEUE', 'default')],
                 'balance' => 'simple',
                 'processes' => 3,
                 'tries' => 3,


driesvints:
Why don't you just use different environments?

jdavidbakr:
@driesvints I'm not sure what you mean?  I think that's what I want to do, but the default config file does not allow you to change the queue based on your environment variables.  Thus this pull request, which would make the default value match the default value in the redis config.

taylorotwell:
This may not be desirable as a default in all cases.

driesvints:
@jdavidbakr to further clarify: imagine if people currently have `REDIS_QUEUE` set to `something-else` and we add this setting by default then their horizon config for production will switch to it if we release this change.

jdavidbakr:
@driesvints I disagree that this is a breaking change, because your config is published to your project so changing this will only impact new deployments.  Unless I'm missing something?  But on the other hand, out of the box the current published config will _not_ work if you have set your `REDIS_QUEUE` in your `env` config.  I don't understand why you wouldn't want the default Horizon queue to match the default Redis queue.

An example of where the default can cause a problem, let's say for local dev I leave `REDIS_QUEUE` null and everything works fine.  However, when I deploy I am required to set `REDIS_QUEUE` to something else because we are using a shared Redis server across multiple projects.  Horizon will appear to be running fine because it will be watching the `default` queue instead of the one specified in my project unless I remember to change the queue in the config, but it won't actually run any of the commands.  I feel like this is a worse case situation that will happen more frequently, whereas this will not be a breaking change because it's not going to change anyone's configs in their deployed sites.

PR #611, opened 2019-06-11
Horizon Clear Command

alexbowers:
In development if you have a lot of jobs in your queues, it can be inconvenient to clear / flush your redis cache, especially if you use multiple connections and / or queues.

This command will go through all your redis connections and queues and clear all jobs in the queue.

### DIFF
--- a/src/Console/ClearCommand.php
+++ b/src/Console/ClearCommand.php
@@ -1,0 +1,76 @@
+<?php
+
+namespace Laravel\Horizon\Console;
+
+use Illuminate\Console\Command;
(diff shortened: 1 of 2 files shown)


taylorotwell:
Hmm, don't think I'm going to put this in the core right now. Feel free to add to your own application.

tomswinkels:
@taylorotwell why not?

tomswinkels:
@alexbowers i think it is better to make a new pull request with an button in the dashboard where you can clear it :)

PR #616, opened 2019-06-14
Revert "[3.0] Display worker CPU and memory utilization in supervisor list"

driesvints:
Reverts laravel/horizon#589


browner12:
for the record

> Since this PR was causing quite some bugs for others and also introduced a performance hit we've decided to revert it entirely. We also won't be re-considering a new PR anymore. We'd like to encourage you to use other tools to track and measure CPU/Memory consumption and just keep Horizon to its core features.

Issue #624, opened 2019-06-25
Horizon should support both queue:listen and queue:work styles


travisaustin:
When running Horizon with jobs that connect to a SOAP server (using `SoapClient`) or connect to AWS S3 using the AWS SDK for PHP, many connections are opened, used, and then left in the `CLOSE_WAIT` TCP state until Horizon is terminated. This is not a limitation of Horizon, but it would be really helpful to run Horizon like the command `queue:listen` does so that `CLOSE_WAIT` connections don't start to pile up. If Horizon job processes could end after each time they ran, any remaining TCP connections would be automatically be released after each job.

As it stands, we have to periodically restart Horizon to keep the number of open connections from hitting the open file maximum for the operating system (which then causes our jobs to fail until we restart Horizon). Restarting Horizon on a timer is a less-than-ideal solution, and I'd really like to run Horizon so that job processes die when complete and are not reused.

If I run the queues using queue:listen instead of using Horizon, the `CLOSE_WAIT` connections don't pile up and I don't have any issues.

I fully understand this is an issue with `SoapClient` and AWS S3 SDK at the core, but is there any possibility Horizon will support this configuration in the future?

Thanks!

shawnhind:
We have the same problem and we restart horizon every 4 hours to get around this problem.

driesvints:
I think if you can implement in a clean and non-breaking way that we're definitely open to a PR here.

travisaustin:
Hey @shawnhind, we submitted a PR with this functionality. Feel free to give it a try!

driesvints:
I'm afraid I'll have to close this because the PR for this was rejected. Please see Taylor's remark here: https://github.com/laravel/horizon/pull/626#issuecomment-507675354

travisaustin:
I'm not sure if it's better to comment here on my issue, or on my PR, so I'm adding a note here to say I argued my position on the PR.

https://github.com/laravel/horizon/pull/626#issuecomment-507758257

jes490:
If someone else needs this --once feature, you can do it like this without changing any vendor code (seems to be working): 

```php
// Add this code in any service provider (HorizonServiceProvider seems most fitting)

use Laravel\Horizon\SystemProcessCounter;
use Laravel\Horizon\WorkerCommandString;
use Illuminate\Queue\Listener;
use App\Console\Commands\HorizonListen;

// Register HorizonListen command (need to pass configured Listener class to it)
[...]

PR #626, opened 2019-07-01
[3.0] New supervisor config option 'once' makes workers run one job then restart

travisaustin:
Re #624 

Some long-running processes leave open files in the file system. After some time, this eventually can cause the number of open files to exhaust the operating system limits, which then can cause Horizon workers to cease working until Horizon is restarted.

Horizon currently runs all commands like `queue:work` does, which can leave these files open piling up (see #624 for information).

This pull-request adds an optional parameter to the configuration for each Supervisor. By setting `once` to `true` in `config\horizon.php`, Horizon will run worker processes similar to `queue:listen` does, and will restart each process after each job.

This change should have no affect on deployments that don't include `once`, or that set `once` to `false`. In this way, this change should be non-breaking.

Thank you for your consideration!

### DIFF
--- a/src/WorkerProcess.php
+++ b/src/WorkerProcess.php
@@ -35,11 +35,13 @@ class WorkerProcess
      * Create a new worker process instance.
      *
      * @param  \Symfony\Component\Process\Process  $process
+     * @param  bool  $once
      * @return void
      */
-    public function __construct($process)
+    public function __construct($process, $once = false)
     {
         $this->process = $process;
+        $this->once = $once;
     }
 
     /**
@@ -52,9 +54,12 @@ class WorkerProcess
     {
         $this->output = $callback;
 
-        $this->cooldown();
-
-        $this->process->start($callback);
+        if ($this->once) {
+            $this->process->start($callback);
+        } else {
+            $this->cooldown();
+            $this->process->start($callback);
+        }
 
         return $this;
     }
--- a/src/Console/SupervisorCommand.php
+++ b/src/Console/SupervisorCommand.php
@@ -24,6 +24,7 @@ class SupervisorCommand extends Command
                             {--min-processes=1 : The minimum number of workers to assign per queue}
                             {--memory=128 : The memory limit in megabytes}
                             {--nice=0 : The process priority}
+                            {--once : Worker processes will only process the next job on the queue and then be re-started}
                             {--paused : Start the supervisor in a paused state}
                             {--queue= : The names of the queues to work}
                             {--sleep=3 : Number of seconds to sleep when no job is available}
@@ -112,7 +113,8 @@ class SupervisorCommand extends Command
             $this->option('sleep'),
             $this->option('tries'),
             $this->option('force'),
-            $this->option('nice')
+            $this->option('nice'),
+            $this->option('once')
         );
     }
 
--- a/src/ProcessPool.php
+++ b/src/ProcessPool.php
@@ -181,7 +181,7 @@ class ProcessPool implements Countable
 
         return new WorkerProcess($class::fromShellCommandline(
             $this->options->toWorkerCommand(), $this->options->directory
-        )->setTimeout(null)->disableOutput());
+        )->setTimeout(null)->disableOutput(), $this->options->once);
     }
 
     /**
--- a/src/QueueCommandString.php
+++ b/src/QueueCommandString.php
@@ -22,6 +22,10 @@ class QueueCommandString
             $string .= ' --force';
         }
 
+        if ($options->once) {
+            $string .= ' --once';
+        }
+
         if ($paused) {
             $string .= ' --paused';
         }
--- a/src/SupervisorOptions.php
+++ b/src/SupervisorOptions.php
@@ -56,6 +56,13 @@ class SupervisorOptions extends WorkerOptions
     public $nice = 0;
 
     /**
+     * Worker processes will only process the next job on the queue and then be re-started
+     *
+     * @var bool
+     */
+    public $once = false;
+
+    /**
      * The working directories that new workers should be started from.
      *
      * @var string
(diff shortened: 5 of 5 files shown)


driesvints:
@travisaustin can you fix the StyleCI issue?

src/WorkerProcess.php
```diff
@@ -52,9 +54,12 @@ public function start(Closure $callback)
     {
         $this->output = $callback;
 
-        $this->cooldown();
-
-        $this->process->start($callback);
+        if ($this->once) {
```
parkourben99:
```
if (! $this->once) {
    $this->cooldown();
}

$this->process->start($callback);
```

Looks cleaner

taylorotwell:
This feels like a very inefficient way of running Horizon. Personally restarting Horizon on a cron like yo note in your issue description sounds much better?

travisaustin:
@taylorotwell, thanks for the feedback.

My issue with restarting Horizon is that it's hard to pick the right timeframe for restarting it.

When we first started using Horizon in production, I didn't reboot Horizon at all, and a few weeks in, very unexpectedly, all my workers started failing with odd errors (SSL connection errors mostly, because they couldn't open the certificate files since too many files/connections were open). It took me quite a bit of work to realize I had open files open in CLOSE_WAIT status, but once I realized that, I set Horizon to restart nightly. Easy enough, right?

The nightly restart worked for a few weeks, but as my app traffic increased, I started to hit errors again about 18-19 hours after each restart. I changed to restarting Horizon every 4 hours.

That worked for a while, but now I have to restart Horizon every hour to get all the hanging connections to close. And, even now, I'm getting up to 500 open connections piling up in that hour, and the file system default is 1024 open files. I'm going to soon be restarting Horizon every 10 minutes to be "safe".

But, how long will that work? What if our application gets even more traffic and restarting every 10 minutes isn't enough? Should I restart every 5 minutes? Every minute?

And how about my long-running jobs? I have a few jobs that take 20-30 minutes to complete. I don't like the idea of restarting Horizon mid-job (even though I know it will be safe, and I know I can turn on fast_termination to allow the new Horizon to get started right away).

That's what led me to believe that it is best to restart each worker after every job. This makes for a very clean system, with no straggling open files or connections, and no need to guess on a solid restart time-frame.

I do agree that it would be MUCH better if this weren't an issue, but I don't intend to rewrite PHP-SOAP or AWS S3 SDKs anytime soon. Our app does rely on these third parties, and these third party SDKs leave open connections until their parent processes terminate.

My PR does not affect anyone who doesn't have this issue. But, those of who do have this issue, will find it very helpful to allow Horizon to operate in a method similar to how `queue:listen` works today. We need a way to restart our workers after each job is processed.

Also, the PR supports enabling this option on a per-supervisor/per-queue basis. In this way, it is possible to allow most jobs to operate "normally", and only force the queues that run the problematic jobs to restart after each job.

If you have a suggestion on a better way to do this or to think about this problem, I would like to hear it. For now, though, I don't know of a better way to ensure that our jobs are correctly cleaned up remove straggling open connections so they don't pile up.

Thank you for your feedback, and for your great products!

src/WorkerProcess.php
travisaustin: Thanks!

travisaustin:
> @travisaustin can you fix the StyleCI issue?

@driesvints I don't see the StyleCI information on the PR any longer. Is that because the PR is closed?

shawnhind:
I think @travisaustin described it well why restarting on a schedule won't work consistently.

It also doesn't make sense to me why Laravel has this option available but Horizon would not.  Why would it be an inefficient way to run Horizon but a fine way to run queue workers without Horizon?  It may just be that i'm not understanding something which makes this less efficient for Horizon but not for running queue workers directly.

driesvints:
@travisaustin when you click the red X next to the commit message you'll can go to the StyleCI warning.

travisaustin:
@driesvints Thanks! (I don't work with PRs very often - I appreciate the help)

mattkingshott:
@travisaustin since the PR was closed, did you find an alternative solution to this?

I ask because I’ve got a similar issue with HTTP requests which is only solved by running queue:listen. 

If not, I’m thinking of forking and using your PR. 

Thanks!

travisaustin:
@mattkingshott No other solution for me. I've just been playing whack-a-mole for the last year restarting Horizon to be sure my open files don't pile up.

mattkingshott:
Damn. That sucks, man. 

I’m looking into seeing if I can get guzzle to force close requests before doing anything extreme like forking. 

It’s related to this issue by the way... https://github.com/laravel/framework/discussions/34353

travisaustin:
@mattkingshott Not sure if this is helpful, but I execute this command every 60 seconds across all of my instances and I store the value so I can centrally track it:

`lsof | grep php | grep CLOSE | wc -l`

That shows me how many processes are "stuck". Then, in `Kernel.php` I have this to hard-restart my queues every hour:

```
 $schedule->command('queue:restart')
            ->hourlyAt(52);
```

Here's what my file count looks like. It's very apparent that not performing `queue:restart` every hour would bring my system down in the course of just a few hours:

image

mattkingshott:
Yeah. Very interesting. 

In my case I don’t think it would help though as my problems happen before the second job even completes. 

I’ve found some guzzle configuration options related to closing connections that I’m going to try.

jes490:
@mattkingshott , @travisaustin see https://github.com/laravel/horizon/issues/624#issuecomment-693112998 as a workaround for this

travisaustin:
Spectacular! I will be giving this a try, for sure. Thank you.

Patabugen:
My use case for wanting this is slightly different: Now that I'm using sail for development I can use horizon for running my queues during dev. Being able to do the equivalent of `queue:listen` but `horizon:listen` would be really handy!

Also relevant: https://github.com/laravel/horizon/issues/624

Issue #621, opened 2019-06-22
Incorrect metric at the dashboard overview


ghobaty:
- Horizon Version: 3.2.3
- Laravel Version: 5.7.28
- PHP Version: 7.2.19
- Redis Driver & Version: predis/predis 1.1.1
- Database Driver & Version: MySQL 5.7

### Description:
I believe the dashboard displays an incorrect label or number for the failed jobs metric.
<img width="1206" alt="Screen Shot 2019-06-22 at 14 35 15" src="https://user-images.githubusercontent.com/6379163/59964009-0d28f180-94fb-11e9-97ca-f39dc9e3b22f.png">
In the screenshot, 19 is the number of jobs failed within the last hour and not last 7 days. 
Not sure if it is related, but my horizon trim configuration is:
```    
'trim' => [
    'recent' => 60,
    'failed' => 10080,
],
```

shawnhind:
I've also seen this value go down between reloads as if the number of failed jobs in the last 7 days has decreased, meanwhile new failed jobs are actually still occurring.  Would this be related?

JayBizzle:
We are also seeing the same issue...

These screenshots were taken on the 13th June...

<img width="218" alt="Screenshot 2019-06-13 at 20 17 36" src="https://user-images.githubusercontent.com/340752/60133659-affc9c80-9795-11e9-92d6-93bed8de8fe8.png">

Says there is 1 failed job in the last 7 days...

<img width="289" alt="Screenshot 2019-06-13 at 20 17 52" src="https://user-images.githubusercontent.com/340752/60133676-b723aa80-9795-11e9-85aa-ca540b8758b9.png">

...but surely it should say 6?

derekmd:
The problem is (by default) only the past hour of failed jobs are included in the dashboard count. The labelled "FAILED JOBS PAST 7 DAYS" period is from `config('horizon.trim.failed')` but the metric is actually filtered by `config('horizon.trim.recent')`.

I put some branches together, one of them can be put up as a pull request.

## Option 1 - make count behave as already labelled

![option-1-week](https://user-images.githubusercontent.com/823566/60773931-aef33580-a0da-11e9-8501-f883be0bbdc4.png)

Use `config('horizon.trim.failed')` (default 7 days) when counting how many minutes ago to include. This will make the dashboard count exactly match the number of rows under the "Failed Jobs" tab for URI /horizon/failed.

[...]

PR #633, opened 2019-07-09
[3.0] Correct dashboard "Failed Jobs Past 7 Days" metric

derekmd:
Fixes https://github.com/laravel/horizon/issues/621

option-1-week

The dashboard "Failed Jobs Past 7 Days" metric almost always shows zero because out-of-the-box the counter is actually from the past _1 hour_ of failed jobs. The time period label is calculated from `config('horizon.trim.failed')` but the Redis count query is filtered by `config('horizon.trim.recent')` minutes ago, same as the "Jobs Past Hour" label in the above screenshot.

This change makes the metric match the time period label so the count matches the number of results under the "Failed Jobs" page at URI /horizon/failed.

## Other possible implementations

I described two other possible approaches in the issue thread: https://github.com/laravel/horizon/issues/621#issuecomment-509032941

Another config/horizon.php option could be added to give devs more control over the failed job count filtered by timestamp.

## Possible deprecations for Laravel Horizon 4.0+

With this merged, `RedisJobRepository::countRecentlyFailed()` and Redis set `'recent_failed_jobs'` would no longer be used in Horizon. They could be earmarked for removal in 4.0? In the 3.0 branch, it's possible userland calls these in custom queue dashboards.

### DIFF
--- a/resources/js/screens/dashboard.vue
+++ b/resources/js/screens/dashboard.vue
@@ -51,10 +51,10 @@
             /**
              * Determine the recently failed job period label.
              */
-            recentlyFailedPeriod() {
+            failedJobsPeriod() {
                 return !this.ready
                     ? 'Failed jobs past 7 days'
-                    : `Failed jobs past ${this.determinePeriod(this.stats.periods.recentlyFailed)}`;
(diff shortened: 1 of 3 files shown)


driesvints:
@derekmd feel free to add the deprecated annotations

Issue #631, opened 2019-07-07
Horizon tries to load resources over HTTP on HTTPS


darki73:
- Horizon Version: 3.2.5
- Laravel Version: 5.8.27
- PHP Version: 7.3
- Redis Driver & Version: predis/phpredis 1.1.1
- Database Driver & Version:

### Description:
Even though Horizon url is accessed through the HTTPS, it still tries to load the resources from the HTTP. The rest of application works just fine, Laravel Nova is also working fine. For some reason only Horizon fails to understand that if it is accessed through the HTTPS, then all resources must be loaded through HTTPS.
![Annotation 2019-07-07 161201](https://user-images.githubusercontent.com/1318627/60768942-a3037580-a0d2-11e9-8235-abb2d429c4f8.png)

Just a side note, Nova is loading resources from the relative path when horizon is trying to load resources from the absolute path (including domain name and protocol)

### Steps To Reproduce:

Install Traefik as reverse proxy, create a service with Nginx running on port 80, proxy container port 80 to real domain with valid SSL certificate, here comes that page
![Annotation 2019-07-07 161128](https://user-images.githubusercontent.com/1318627/60768888-fb864300-a0d1-11e9-8460-86a3627a6aa7.png)

driesvints:
Are you behind a load balancer? Have you configured your trusted proxies correctly? https://laravel.com/docs/5.8/requests#configuring-trusted-proxies

darki73:
This is the content of my `TrustProxies.php` middleware, no load balancing, no nothing.
![Annotation 2019-07-08 152541](https://user-images.githubusercontent.com/1318627/60809974-ac551680-a194-11e9-9940-10ffaf9222ff.png)


I (presumably) found a way to actually **Nova**_fy_ Horizon by changing:

**From (styles):**
```
<link href="{{ asset(mix($cssFile, 'vendor/horizon')) }}" rel="stylesheet" type="text/css">
```
**To**
```
<link href="{{ mix($cssFile, 'vendor/horizon') }}" rel="stylesheet" type="text/css">
```

**And (js) from:**
```
<script src="{{ asset(mix('app.js', 'vendor/horizon')) }}"></script>
```
**To:**
```
<script src="{{ mix('app.js', 'vendor/horizon') }}"></script>
```

This is pretty much how it is done in Nova, so from now on Horizon will try to fetch files from relative path rather than absolute

driesvints:
I think you're correct. @themsaid do you also follow this reasoning?

[...]

PR #638, opened 2019-07-30
[3.0] Absolute to Dynamic resource paths conversion

darki73:
Allow Horizon to load resources from dynamic path rather than absolute

**Framework Version:** 5.8.29
**Horizon Version:** 3.2.6
**Fixes Issue:**  https://github.com/laravel/horizon/issues/631

### Problem
Right now, Horizon tries to fetch resources (CSS and JS) files from absolute paths, which sometimes may break the panel due to the files being fetches from the **http://** instead of **https://**
Annotation 2019-07-07 161201
### Possible solution
We can fix that by following the same code for loading resources as **Laravel Nova** is using.
Annotation 2019-07-08 153130
This will force Horizon to fetch resources from relative paths rather than absolute.

### DIFF
--- a/resources/views/layout.blade.php
+++ b/resources/views/layout.blade.php
@@ -11,7 +11,7 @@
 
     <!-- Style sheets-->
     <link href="https://fonts.googleapis.com/css?family=Nunito" rel="stylesheet">
-    <link href="{{ asset(mix($cssFile, 'vendor/horizon')) }}" rel="stylesheet">
+    <link href="{{ mix($cssFile, 'vendor/horizon') }}" rel="stylesheet">
 </head>
 <body>
 <div id="horizon" v-cloak>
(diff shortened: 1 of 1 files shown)


bonzai:
Won't this break Horizon for people that use custom `asset_url`?

Can't you use `URL::forceScheme('https')` in your service provider?

Frozire:
> Won't this break Horizon for people that use custom `asset_url`?
> 
> Can't you use `URL::forceScheme('https')` in your service provider?

Setting `URL::forceScheme('https')` in a service provider does indeed fix this issue.

taylorotwell:
Setup your trusted proxies correctly.

LGZ1998:
This may not be the best practice, but setting .env`ASSET_URL=/.` allows me to access normally

PR #640, opened 2019-08-05
[3.0] Changed default Redis prefix

ivantorgov:
When you are running multiple installations of Horizon on the same server you have to change `HORIZON_PREFIX` environment variable to unique value for every installation.
People often forget to do this. 
So I provide more flexible default Redis prefix based on `APP_NAME` variable (inspired by default session prefix in Laravel).

### DIFF
--- a/config/horizon.php
+++ b/config/horizon.php
@@ -1,5 +1,7 @@
 <?php
 
+use Illuminate\Support\Str;
+
 return [
 
     /*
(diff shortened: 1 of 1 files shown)


driesvints:
This might be breaking for people who haven't published the config and are relying on the current default.

ivantorgov:
> This might be breaking for people who haven't published the config and are relying on the current default.

Hmm. I think you are right. Have you an idea how does make it compatible with existing projects?

ivantorgov:
I think this improvement is important because I often was forgetting to change HORIZON_PREFIX when I ran multiple installations of Horizon on the same server. And I found that many people have had this problem too.
Maybe can you accept this improvement in a next major release?

driesvints:
@ivantorgov I think we can. Feel free to submit this to master.

Issue #639, opened 2019-08-04
Horizon Page Open + redirect()->back()


knno:
- Horizon Version: **3.2.7**
- Laravel Version: **5.8.30**
- PHP Version: **7.3.7-2**
- Redis Driver & Version: predis/phpredis **4.0.9 (redis-cli in homestead)**
- Database Driver & Version: **mysql server version 5.7.27**

### Description:
`redirect()->back()` is giving me "302 Found" to random horizon routes:
Such as
... *Redirecting to horizon/api/workload* ...
... *Redirecting to horizon/api/stats* ...
When I have the horizon tab open in my browser, this happens to all tabs in short times. 

### Steps To Reproduce:
1. Add this route to web.php
```php
Route::get('/test/redirect-bug', function() {
	dd(redirect()->back());
});
```
2. Open horizon on a tab in your browser.
3. Visit "<sitename.com>/test/redirect-bug" and keep refreshing the page. You will see the title part of "Redirecting to" in the response object's content.

### Note
I am unsure if it's the default wanted behavior, if so, how can I evade this?
Thanks.

driesvints:
Can you post your entire `web.php` routes file?

knno:
It is actually not much..

```php
Route::get('/test/redirect-bug', function() {
	dd(redirect()->back());
});
Route::get('/test/empty', function() {
	return view('empty');
});
```

I try to visit /test/empty without having horizon open in a browser tab, and it works fine.
When horizon is open, this bug happens. It happens regardless of horizon's auto load entries.

SDekkers:
The problem might be in the `startSession` middleware that isn't able to detect the `/horizon/api/`-routes as ajax request.

driesvints:
I can't reproduce this. Something tells me that there's still something missing here. Can you first please try one of the support channels below? If you can actually identify this as a bug, feel free to report back and I'll gladly help you out and re-open this issue.

- [Laracasts Forums](https://laracasts.com/discuss)
- [Laravel.io Forums](https://laravel.io/forum)
- [StackOverflow](https://stackoverflow.com/questions/tagged/laravel)
- [Discord](https://discordapp.com/invite/KxwQuKb)
- [Larachat](https://larachat.co)
- [IRC](https://webchat.freenode.net/?nick=laravelnewbie&channels=%23laravel&prompt=1)

Thanks!

SDekkers:
@driesvints I was able to reproduce this;
1. Have a horizon instance running
[...]

PR #642, opened 2019-08-05
Forcing Vue to use the X-Requested-With header, set to XMLHttpRequest

SDekkers:
-- resubmitting without compiled assets.

This PR is a fix for issue [Issue 639](https://github.com/laravel/horizon/issues/639)

This adds an additional header to every request to the horizon API routes. 
By setting the header `X-Requested-With` to `XMLHttpRequest`, Laravel knows the request is an `ajax`-request, so it does not update previous URL in the user session, which was undesired behavior.

Without this PR, users having Horizon open in one tab, and other parts of their project in another, might have their previous URL in their session updated to for example: `/horizon/api/workload`, so any part of their project using `URL::previous()` might return the user to an API resource, in stead of the previous URL the user visited manually

### DIFF
--- a/resources/js/app.js
+++ b/resources/js/app.js
@@ -11,6 +11,8 @@ require('bootstrap');
 
 let token = document.head.querySelector('meta[name="csrf-token"]');
 
+axios.defaults.headers.common['X-Requested-With'] = 'XMLHttpRequest';
+
 if (token) {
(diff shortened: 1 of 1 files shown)


driesvints:
@SDekkers I think it's best that you explain in detail in your first comment what this actually fixes. Prs without thorough explanation tend to get closed (see the pr template).

SDekkers:
@driesvints done

PR #643, opened 2019-08-05
[4.0] Changed default Redis prefix

ivantorgov:
When you are running multiple installations of Horizon on the same server you have to change HORIZON_PREFIX environment variable to unique value for every installation.
Sometimes I was forgetting to change HORIZON_PREFIX when I ran multiple installations of Horizon on the same server. And I found that many people have had this problem too.
So I provide more flexible default Redis prefix based on APP_NAME variable (inspired by default session prefix in Laravel).
I talked with @driesvints and he suggested that this improvement can affect existing projects. I agree with him and I think that it will be better to add this improvement in a next major release (like 4.0.0).

### DIFF
--- a/config/horizon.php
+++ b/config/horizon.php
@@ -1,5 +1,7 @@
 <?php
 
+use Illuminate\Support\Str;
+
 return [
 
     /*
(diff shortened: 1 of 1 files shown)


fgilio:
A really welcomed change, thanks!

PR #646, opened 2019-08-07
[4.0] Change format

tomswinkels:
Better format, the default app_name already set in the app config.


driesvints:
We've reverted this because using the config helper inside config files is discouraged (caching, circular calls).

Issue #622, opened 2019-06-25
maxProcesses and minProcesses does not work


tomswinkels:
- Horizon Version: 3.2.3
- Laravel Version: 5.8.23
- PHP Version: 7.2
- Redis Driver & Version: predis/phpredis 1.1.1

We have the next config;

```
            'Default' => [
                'connection' => 'redis',
                'queue' => ['Default'],
                'balance' => 'auto',
                'minProcesses' => 1,
                'maxProcesses' => 8,
                'tries' => 3,
                'sleep' => 1,
                'timeout' => 60,
                'delay' => 0
            ],
```

There are always running 8 processes for default, even when there are no jobs. On the dashboard and on the server processes list.

tomswinkels:
When i check the process list on the server i see the min and max settings in the supervisor command. (see screen) But i see always running 8 horizon workers. (other screen)

![image](https://user-images.githubusercontent.com/9036151/60247859-c4d64e80-98c1-11e9-95a3-e9b377947c72.png)

![image](https://user-images.githubusercontent.com/9036151/60247908-dcadd280-98c1-11e9-9484-64de45d95c31.png)

ronnievisser:
I got this exact same problem! Is there a solution for already?

francescomalatesta:
Same problem here :( any ideas?

SDekkers:
If you look at [AutoScaler.php on Line 99](https://github.com/laravel/horizon/blob/3.0/src/AutoScaler.php#L99), you can see that if the time to clear is 0, it'll always return the max processes

We could add another line so the function would be something like this:

	protected function numberOfWorkersPerQueue(Supervisor $supervisor, Collection $timeToClear)
	{
		$timeToClearAll = $timeToClear->sum();

		return $timeToClear->mapWithKeys(function ($timeToClear, $queue) use ($supervisor, $timeToClearAll) {
			return $timeToClearAll > 0 && $supervisor->options->autoScaling()
				? [$queue => (($timeToClear / $timeToClearAll) * $supervisor->options->maxProcesses)]
				: ($timeToClearAll == 0) ? [$queue => $supervisor->options->minProcesses]
					: [$queue => $supervisor->options->maxProcesses / count($supervisor->processPools)];
		})->sort();
	}

But I don't know if that would have any side effects.

tomswinkels:
@driesvints check this comment :)

SDekkers:
I see the CI failed on 2 tests, but the 2 tests that fail:

[...]

PR #647, opened 2019-08-07
[3.0] Fix AutoScale by adding a return statement when queue is empty, to return minProcesses

SDekkers:
This PR is a fix for #622. 

It changes the way the amount of desired workers is calculated. 
When the total time needed to clear all jobs is zero, it returns the minimum amount of processes, instead of the max amount of processes. When the amount of time needed to clear the jobs increases, processes will automatically scale back up.


taylorotwell:
Going to hold off on this for now. I don't particularly like nested ternary statements.

SDekkers:
Would you prefer normal if/elseif/else statement or another solution ?

PR #652, opened 2019-08-22
[4.0] Remove Predis support

driesvints:
This will remove support for Predis as it's not maintained anymore. We've also begun to deprecate Predis support in Laravel itself.


taylorotwell:
Let's wait on this for a a few more months.

driesvints:
Sure thing. I'll attempt again in November.

Issue #554, opened 2019-03-21
Publish assets with Composer hook


driesvints:
We're currently getting a lot of issues where people forgot to update their Horizon assets. Technically it should be possible for us to implement a Composer hook in the `composer.json` command that runs `php artisan horizon:assets` so updated assets get published.

drbyte:
It may be similarly prudent to specifically mention appropriate upgrade steps in the docs at
https://laravel.com/docs/5.8/horizon#installation

driesvints:
@drbyte I was thinking to do that first but if we can automate this then this is unnecessary.

drbyte:
True, but docs also help with troubleshooting when automation goes awry. :)

driesvints:
Yeah, true. You can send in a PR if you want :)

drbyte:
Done

barryvdh:
What do you think about something like this? https://github.com/laravel/telescope/pull/729

Eg. show a warning when the assets are out-of-date. Still requires a manual step but at least people are warned.

I don't think composer hooks work for non-root projects.

driesvints:
Hmm I'm not sure if something like that should make its way into the actual UI.

> I don't think composer hooks work for non-root projects.

Yeah I'm not sure either. I need to look into this but bit swamped atm.

barryvdh:
Yeah the alternative is not publishing it, but loading the assets through a controller, example: https://github.com/fruitcake/laravel-telescope-toolbar/blob/master/src/Http/Controllers/ToolbarController.php#L67-L86

barryvdh:
> Hmm I'm not sure if something like that should make its way into the actual UI.

Why shouldn't it make the way into the UI? It's only visible when it's outdated and a relatively simple check to prevent common errors.

driesvints:
Because you're displaying an implementation detail into an app. Imo the app shouldn't have any knowledge about how the ui is build. This is a slippery slope imo.

fgilio:
I personally don't find it bad for the UI to help with this. But of course, if this can be avoided altogether that would be the best solution.
Maybe adding a command similar to `php artisan package:discover` but that takes care of re publishing all packages assets. It would only work out of the box for new Laravel installs.

driesvints:
I still think we can solve this with Composer hooks.

barryvdh:
https://getcomposer.org/doc/articles/scripts.md
[...]

PR #656, opened 2019-09-02
[3.0] Publish assets with Composer

tomswinkels:
Resolve https://github.com/laravel/horizon/issues/554

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -55,5 +55,10 @@
         "sort-packages": true
     },
     "minimum-stability": "dev",
-    "prefer-stable": true
+    "prefer-stable": true,
+    "scripts": {
+        "post-update-cmd": [
+            "@php artisan horizon:assets"
+        ]
+    }
 }


driesvints:
I think we need to indeed have something like this but this specific pr seems to break the build. Let's see in the issue what the best way to tackle this would be before we attempt a new pr.

composer.json
```diff
@@ -55,5 +55,10 @@
         "sort-packages": true
     },
     "minimum-stability": "dev",
-    "prefer-stable": true
+    "prefer-stable": true,
+    "scripts": {
+        "post-update-cmd": [
+            "@php artisan horizon:assets"
```
stancl: How about creating a class that would check whether `artisan` exists and if so, it would run `Artisan::call('horizon:assets')`?

stancl: [Nevermind](https://github.com/laravel/horizon/issues/554#issuecomment-529456544)

Issue #192, opened 2017-10-10
Jobs lists fail when there is a large payload for the jobs


Artistan:
When you have a series of job that have a large payload (say 2mb each) then the recent jobs listing fails in horizon due to memory exhausted when dealing with the json.

Possible Solution:
https://github.com/laravel/horizon/pull/191

**Some project details on how this is affecting us...** 
- passing JSON through redis to be used with elasticsearch
- bulkJson for elastic is configurable, currently running at 2mb.
- loading the last 50 jobs results in fatal error due to memory exhausted.

```log
[2017-10-10 15:39:34] local.ERROR: Allowed memory size of 268435456 bytes exhausted (tried to allocate 100405248 bytes) {"exception":"[object] (Symfony\\Component\\Debug\\Exception\\FatalErrorException(code: 1): Allowed memory size of 268435456 bytes exhausted (tried to allocate 100405248 bytes) at /var/www/vhosts/laravel-app/vendor/laravel/framework/src/Illuminate/Http/JsonResponse.php:67)
[stacktrace]
#0 /var/www/vhosts/laravel-app/vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php(137): Symfony\\Component\\Debug\\Exception\\FatalErrorException->__construct()
#1 /var/www/vhosts/laravel-app/vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php(123): Illuminate\\Foundation\\Bootstrap\\HandleExceptions->fatalExceptionFromError()
#2 /var/www/vhosts/laravel-app/vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php(0): Illuminate\\Foundation\\Bootstrap\\HandleExceptions->handleShutdown()
#3 /var/www/vhosts/laravel-app/vendor/laravel/framework/src/Illuminate/Http/JsonResponse.php(67): json_encode()
#4 /var/www/vhosts/laravel-app/vendor/symfony/http-foundation/JsonResponse.php(50): Illuminate\\Http\\JsonResponse->setData()
#5 /var/www/vhosts/laravel-app/vendor/laravel/framework/src/Illuminate/Http/JsonResponse.php(27): Illuminate\\Http\\JsonResponse->__construct()
#6 /var/www/vhosts/laravel-app/vendor/laravel/framework/src/Illuminate/Routing/Router.php(709): Illuminate\\Http\\JsonResponse->__construct()
[...]

PR #657, opened 2019-09-02
[3.0] Fix problem with too large payloads

Zlob:
Hi!
I would like to propose a solution for issue with unnecessary loading of jobs payload and exception description on the reports pages [issue 1](https://github.com/laravel/horizon/issues/192) , [issue 2](https://github.com/laravel/horizon/pull/191). In case with a timeout of 3 seconds between requests and a payload of 10MB, jobs request do not have enough time to be executed, which leads to accumulation of requests and freeze the application. The option to reduce jobs payload is not always suitable. My suggestion is to remove data from "heavy" fields such as `payload->data` and `exception` on the report pages, where they are don`t even uses. This change is safe because it does not affect the rest of the Horizon functionality.

### DIFF
--- a/src/Http/Controllers/Controller.php
+++ b/src/Http/Controllers/Controller.php
@@ -23,6 +23,7 @@ class Controller extends BaseController
         return $collection->transform(function ($item, $key) {
             $item->payload->data = [];
             $item->exception = [];
+
             return $item;
         });
     }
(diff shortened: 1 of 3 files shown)


taylorotwell:
I am closing this pull request because it lacks sufficient explanation, **tests**, or both. It is difficult for us to merge pull requests without these things because the change may introduce breaking changes to the framework.

Thanks!

PR #665, opened 2019-09-16
Filter recent jobs by tag

### LINKED ISSUE #454: Filtering options
Will you guys be open to accept a PR that'll add filtering options to the Recent Jobs & Failed pages. Filter by Connection, Queued At and Status.

sebdesign:
This PR adds a search box to the recent jobs page, that allows to filter recent jobs by tag.
The implementation is similar to the search tags functionality on the failed jobs page.

Screen Shot 2019-09-17 at 01 27 01

## Description

When a job is pushed, we are temporarily storing the job id for each of its tags as `recent:{tag}`, for as long as the `horizon.trim.recent_failed` or `horizon.trim.recent` configuration values are persisting the recent jobs. This is handled by the `StoreTagsForRecentJob` listener, which is similar to the `StoreTagsForFailedJob`.

The filtering is performed like the `FailedJobsController`. If the request query has a tag, we are fetching the jobs ids for this tag with the `TagRepository`, and then we are fetching the matching jobs from the `JobRepository`.

## Caveats

This partially fixes #454, and it doesn't seem to break backwards compatibility. Of course once the users update Horizon, they will not be able to get search results the **existing** recent jobs, since the recent tags will be missing. But for any upcoming job, the recent tags will start to appear.

## Testing

I have copied two tests from the failed job tags functionality, which ensure that the `recent:{tag}` is being stored when a job is pushed, and that the tags have an expiration.


Cannonb4ll:
Since this PR was merged, I am for some reason not able to use the 'Next' or 'Previous' buttons anymore to paginate. Have you tested this if this still worked after your change @sebdesign ?

GC-Mark:
We updated this morning and have just tested the `Next` and `Previous` buttons and they don't work for us either

sebdesign:
@Cannonb4ll I can confirm this is a bug. The Next button is not working for me either, unless you are searching for a tag. I will fix this today.

driesvints:
We've decided to revert this for now in light of https://github.com/laravel/horizon/issues/740

See: https://github.com/laravel/horizon/pull/741

PR #669, opened 2019-09-20
Process limit PER QUEUE in supervisor (simple FIFO)

LasseRafn:
This will allow you to set a `maxQueueProcesses` config on the supervisor, which limits the amount of processes for each queue in the supervisor.

Think of this as a simple way to eg. implement FIFO; by setting the variable to 1, but keeping the max supervisor processes high. It would then never scale beyond one process for each queue, but the total processes running could be as high as `queues * maxQueueProcesses` (3 queues * maxQueueProcesses(1) = 3 processes in the supervisor).

It is especially useful for a multi-tenant setup where you'd have maybe a queue per tenant, and want a total of X processes, but still only want N processes for each tenant/queue. 

I can't think of any use cases besides (naive) FIFO where maxQueueProcesses is set to 1.

It is implemented as opt-in so it shouldn't break for anyone who doesn't add the config key.

This does not work with `balance => false`.

A test has been added as well.

(Naming and implementation might be in need of a change!)

### DIFF
--- a/src/AutoScaler.php
+++ b/src/AutoScaler.php
@@ -2,9 +2,9 @@
 
 namespace Laravel\Horizon;
 
+use Illuminate\Contracts\Queue\Factory as QueueFactory;
 use Illuminate\Support\Collection;
 use Laravel\Horizon\Contracts\MetricsRepository;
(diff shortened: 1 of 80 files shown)


LasseRafn:
Still some changes to be made.

PR #679, opened 2019-10-02
[3.x] Default monitored tags

mustorze:
Tag tracking functionality is very useful and used by users. Based on this I realized that we always need to test horizon in other environments and unfortunately we need to recreate all tags for monitoring.

With this proposed functionality we implemented tags to be monitored by default, thus avoiding all rework.


taylorotwell:

Arkanius:
Sad about that =/. I have almost 70 tags that I use to monitor. Sometimes I have problems in my elastic cache and I have to flush it; then, I lost all my tags and have to re-created all again. 

Save the tags in a config file makes a lot of sense. @mustorze if you create a package of it you would help a lot! Please think about it since it seems that we won't have this nativelly.

mustorze:
@Arkanius i'm finally finished the package.

https://github.com/mustorze/horizon-config-tags

enjoy =))

PR #680, opened 2019-10-03
[3.x] Add Horizon FlushCommand

driesvints:
This adds a new Horizon FlushCommand which will clear the meta data of Horizon from its Redis connection. Kudos to @freekmurze for the inspiration.

### DIFF
--- a/src/Console/FlushCommand.php
+++ b/src/Console/FlushCommand.php
@@ -22,7 +22,7 @@ class FlushCommand extends Command
      *
      * @var string
      */
-    protected $description = 'Flush out all data for the Horizon Redis connection';
+    protected $description = 'Flush out all the Horizon meta information for the dashboard.';
 
     /**
      * Execute the console command.
--- a/src/Console/FlushCommand.php
+++ b/src/Console/FlushCommand.php
@@ -1,0 +1,37 @@
+<?php
+
+namespace Laravel\Horizon\Console;
+
+use Illuminate\Console\Command;
+use Illuminate\Console\ConfirmableTrait;
+use Illuminate\Contracts\Redis\Factory as RedisFactory;
+
+class FlushCommand extends Command
+{
+    use ConfirmableTrait;
+
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'horizon:flush';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Flush out all data for the Horizon Redis connection';
+
+    /**
+     * Execute the console command.
+     *
+     * @param  \Illuminate\Contracts\Redis\Factory  $redis
+     * @return void
+     */
+    public function handle(RedisFactory $redis)
+    {
+        $redis->connection('horizon')->flushdb();
+    }
+}
--- a/src/Console/FlushCommand.php
+++ b/src/Console/FlushCommand.php
@@ -32,6 +32,10 @@ class FlushCommand extends Command
      */
     public function handle(RedisFactory $redis)
     {
+        if (! $this->confirmToProceed()) {
+            return;
+        }
+
         $redis->connection('horizon')->flushdb();
     }
 }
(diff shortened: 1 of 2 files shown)


taylorotwell:
What exactly does this flush. Would be helpful to put that in the command description.

driesvints:
It basically resets all the data that's being kept in Redis for the dashboard of Horizon. Basically flushes the connection set by the `horizon.use` config setting. I've pushed a commit which hopefully clarifies this.

mfn:
It's a pretty dangerous thing to do.

Let's not forget how many people have been bitten that when using the Redis cache driver, `artisan cache:clear` also performs a flushdb; after which you hoped you haven't "shared" your redis database with something else => like the workers.

Not sure this is **needed** for horizon. IMHO the PR description doesn't really make a case for "why", also cites someone regarding inspiration?!

driesvints:
@mfn @freekmurze told me he uses this on every Horizon project he builds. It's also imo a command only to be used locally (hence the confirmable trait). It gives you a nice way to clear the dashboard to start fresh when testing locally.

Please also note that this only flushes the Redis "horizon" connection and not your other Redis connections. The "horizon" connection is instantiated by default by Horizon and only keeps the dashboard meta data.

fitztrev:
Just tested this.

> It's also imo a command only to be used locally (hence the confirmable trait).

It's not prompting. Missing the `$this->confirmToProceed()` check.

> Please also note that this only flushes the Redis "horizon" connection and not your other Redis connections.

By default, the `horizon.use` config is the `default` connection. This would also flush anything else that uses that default connection (cache, queue). Could have unintentional effects.

driesvints:
> It's not prompting. Missing the $this->confirmToProceed() check.

Right! Thanks for mentioning this. Added this.

> By default, the horizon.use config is the default connection. This would also flush anything else that uses that default connection (cache, queue). Could have unintentional effects.

I was indeed wrong about this. This will indeed flush anything from the default connection. As I said, this should only be used locally and not in production. Consider it as "migrate:fresh" but for Horizon :)

fitztrev:
> I was indeed wrong about this. This will indeed flush anything from the default connection. As I said, this should only be used locally and not in production. Consider it as "migrate:fresh" but for Horizon :)

If the flush affects more than just Horizon, then I think having it under the `horizon:` artisan namespace is confusing.

driesvints:
@fitztrev I don't agree. It makes use of the connection defined for Horizon. Plus like I said this only needs to be used locally.

fitztrev:
The default connection for Horizon is `default`, which is the same as queue workers and the cache. By default, there is no separate `horizon` connection. And it's not mentioned anywhere to set one up. So `horizon:flush`, under all the default settings, will do more than what its command signature says. That's why I think it would be confusing.

Could the horizon metric keys be tagged instead and then *only* those keys are deleted as opposed to flushing the entire db?

mfn:
> Could the horizon metric keys be tagged instead and then _only_ those keys are deleted as opposed to flushing the entire db?

afaik they are, they've a common prefix: but this requires _iterating_ over keys which is a performance issue in Redis with a *lots of keys*.

I still think this is too dangerous and generally not needed :|

driesvints:
> there is no separate horizon connection. 

Actually, I'm not sure anymore. There is a separate connection set here, which is being used (with a prefix):

https://github.com/laravel/horizon/blob/2c328cbd3ef0fe1e69971cadff6e0cf522cb3bcd/src/Horizon.php#L103

So as I understand it, even though it's the same Redis database, it uses a separate prefix for storing the horizon data. So if you did a flush on that, wouldn't it just flush the data that's prefixed and leave the rest alone?

> I still think this is too dangerous and generally not needed :|

I disagree. Like I said a couple of times above, you would only be using this in development :)

fitztrev:
No, it flushes the entire database--not just things prefixed with `horizon:`

taylorotwell:
I'll just hold off on this for now. If people want to flush something Redis related they can just run FLUSHDB in the Redis console already.

PR #681, opened 2019-10-03
Prevent 'memory exhausted' when deleting monitored tag

### LINKED ISSUE #625: 'Memory Exhausted' when removing monitored tag
- Horizon Version: 3.2.3
- Laravel Version: 5.8.24
- PHP Version: 7.3.4
- Redis Driver & Version: predis 1.1.1
- Database Driver & Version: mysql 5.7

### Description:

I was monitoring a tag on my production server. That tag was linked to a couple of 100 000 jobs.
I suddenly noticed that my redis-server was using a lot of memory (almost 10GB) and I discovered that it was caused by the monitoring.
When I removed the monitoring, the issue persisted and the RAM usage did not chage.
After further investigation, I saw that the `StopMonitoringTag` job class failed because the memory was exhausted. That's because all the jobs (couple of 100 000 jobs) are all fetched/loaded in php and then deleted.

Should this not be done paginated?
I suspect this may be an issue in other places as well, if Horizon fetches a huge amount of jobs (e.g. trimming recent/failed/monitored jobs)

(The only way I could fix this issue, was `redis-cli flushdb`, which is normally not ideal on a production server 😅)

### Steps To Reproduce:

- Add monitoring for a certain tag
- Generate huge amounts of jobs that use the monitored tag
- Check the RAM usage increase by your redis-server + check the amount of redis keys on the redis-server
- Remove the monitored tag
- See that your RAM usage did not change + see that the amount of redis keys did not change
- Try to run dispatch the `StopMonitoringTag` job class manually via tinker. This will throw a 'memory exhausted' error.

gdebrauwer:
Fixes #625.

**What?**

When you monitor a tag and a lot of jobs are attached to tag, you will get a 'memory exhausted' php error when you remove the tag from monitoring list. This is caused by fetching all jobs of that tag in one query (in order to delete them all).

**How?**

Instead of fetching all jobs of the tag in one query, I used the `paginate()` method that is already present in `TagRepository`. By fetching and deleting small batches of jobs we will keep the php memory usage low at all times.

_By default the pagination has a limit of 25 per page. Maybe we should use a limit that's a bit higher in this case (50, 100, ...)?_

### DIFF
--- a/src/Jobs/StopMonitoringTag.php
+++ b/src/Jobs/StopMonitoringTag.php
@@ -36,7 +36,13 @@ class StopMonitoringTag
     {
         $tags->stopMonitoring($this->tag);
 
-        $jobs->deleteMonitored($tags->jobs($this->tag));
+        $tagJobs = $tags->paginate($this->tag);
+
+        while (count($tagJobs) !== 0) {
+            $jobs->deleteMonitored($tagJobs);
+
+            $tagJobs = $tags->paginate($this->tag);
+        }
 
         $tags->forget($this->tag);
     }


gdebrauwer:
I didn't know `deleteMonitored` deletes the jobs async 🤦‍♂️ I will take another look at it. I will close my PR for the moment.

PR #691, opened 2019-10-20
New "Back to Site" link in the Horizon sidebar, controlled via config

Kovah:
Adds a config option to display a link on the Horizon dashboard. This link may contain any site url, e.g. to the admin panel or just the front page, to allow the user to quickly return to their site from the Horizon dashboard without manually editing the URL in the browsers' address bar, or clicking back a few times. The link is only displayed if the corresponding config option is present and valid.
This is just a small change, but would have saved me some minutes already.

I used the arrow from the dashboard icon for the new link as I'm not sure which icon should be used instead.
If the link should be hidden by default I can update the config to be null and add a proper example to the description, if this is preferred. I may also move it to the bottom.

Preview:
Screenshot 2019-10-20 at 16 24 53

Hope this get merged. 😊

### DIFF
--- a/config/horizon.php
+++ b/config/horizon.php
@@ -56,6 +56,19 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Horizon Site Link
+    |--------------------------------------------------------------------------
+    |
+    | The URL set here will enable a new point in the sidebar of the Horizon
+    | dashboard which allows you to go back to your site, without having to
+    | modify the URL in the browsers' address bar or clicking back a few times.
+    |
+    */
+
+    'site_link' => url('/'),
+
+    /*
+    |--------------------------------------------------------------------------
     | Horizon Route Middleware
     |--------------------------------------------------------------------------
     |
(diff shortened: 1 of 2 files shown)


Christophvh:
Laravel installs have an APP_URL in the main config/app file, it would be more logical to use this instead of adding another config value.

Kovah:
Yes, the app url is already available. I would not like to use it to give users the possibility to further customize the used URL. The user might not want to return to the front page, but instead to the admin panel at /nova, /admin or whatever. Also, if the config is remove or disabled, the link disappears from the sidebar.

config/horizon.php
```diff
@@ -54,6 +54,19 @@
...
+    | Horizon Site Link
+    |--------------------------------------------------------------------------
+    |
+    | The URL set here will enable a new point in the sidebar of the Horizon
+    | dashboard which allows you to go back to your site, without having to
+    | modify the URL in the browsers' address bar or clicking back a few times.
+    |
+    */
+
+    'site_link' => url('/'),
```
setkyar: Should we use `env()`? horizon domain and the app domain can't be the same all the time?

taylorotwell:

PR #692, opened 2019-10-21
[3.x] Set same ttl for tags as same as recent jobs

vigneshgurusamy:
After introducing **Filter recent jobs by tag** feature in PR #665, my Redis storage size started increasing gradually. When I started analysing the keyspace usage, I found out 70-75% of my Redis memory is occupied by recent tags.

My horizon trim configuration is

```php
'trim' => [
    'recent' => 15,
    'recent_failed' => 10080,
    'failed' => 10080,
    'monitored' => 10080,
],
```

Based on the above configuration my recent jobs are cleared after **15 minutes**, but recent tags will be cleared only after **7 days**.

My PR will set TTL for recent tags as same as recent jobs.

### DIFF
--- a/src/Listeners/StoreTagsForRecentJob.php
+++ b/src/Listeners/StoreTagsForRecentJob.php
@@ -38,7 +38,7 @@ class StoreTagsForRecentJob
         })->all();
 
         $this->tags->addTemporary(
-            config('horizon.trim.recent_failed', config('horizon.trim.recent', 60)),
+            config('horizon.trim.recent', 60),
             $event->payload->id(),
             $tags
         );


sebdesign:
@vigneshgurusamy @taylorotwell This setting was intended. The reason i'm using the `recent_failed` and the `recent` as a fallback is because in the recent jobs we are also displaying jobs that **failed** recently. And the users should be able to search tags for the recent jobs that failed.

In your case, that means that you can only search for tags for jobs that were handled in the last **15 minutes**. You might have some recent failed jobs from the last **week** but you won't be able to search for them.

vigneshgurusamy:
@sebdesign Approximately ratio between recent_jobs & recent_failed_jobs is **1000:1**. So storing the tags more than the trim period for recent_jobs does not make sense.

Alternately what I can think of is to just increase the TTL based on config `horizon.trim.recent_failed` for tags only when the job is failed.

driesvints:
@sebdesign but you can still search for those jobs in the failed tabs right? So I don't think it's necessary to keep references twice?

PR #694, opened 2019-10-24
[3.0] Rename asset command to publish

dwightwatson:
I noticed that Horizon is the odd one out when it comes to the naming the command to re-publish assets.

```
roomies git:(master) ✗ php artisan | grep publish
  horizon:assets       Re-publish the Horizon assets
  nova:publish         Publish all of the Nova resources
  telescope:publish    Publish all of the Telescope resources
  vendor:publish       Publish any publishable assets from vendor packages
```

This simply updates the command to match - `horizon:publish`.

In addition it adjusts the order of commands in the service provider - the bottom half appeared to be alphabetical but the top half was not, so I made it all alphabetical.

### DIFF
--- a/src/Console/PublishCommand.php
+++ b/src/Console/PublishCommand.php
@@ -4,21 +4,21 @@ namespace Laravel\Horizon\Console;
 
 use Illuminate\Console\Command;
 
-class AssetsCommand extends Command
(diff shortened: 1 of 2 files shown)


taylorotwell:
I think this would technically need to go to master branch.

PR #701, opened 2019-11-06
[4.0] Added "Attempts" to recent job list, added "reserved" icon to indicated which jobs are being worked on.

piotrgrabowskielastique:
Added "Attempts" to recent job list, added "reserved" icon to indicated which jobs are being worked on. Horizon does not display if a job is being tried again and or processed. This helps debugging big job queues.


piotrgrabowskielastique:
github is an ass.

PR #704, opened 2019-11-13
[3.0] set default 'tries' to 1

browner12:
trying a Job more than 1 time can have unintended side effects, and for the less experienced developer, defaulting to 3 tries could cause problems.

my opinion would be that we default to 1 try, and allow the developer who understands the considerations for running a Job multiple times to increase this if they like.

### Scenario

```php
class PlaceOrder implement ShouldQueue
{
    public $tries = 3;

    public function __construct(array $products)
    {
        $this->products = $products;
    }

    public function handle()
    {
        //create the Order
        $order = new Order;
        $order->product = $this->products;
        $order->save();

        //do something that could generate an Exception
    }
}
```

If an Exception is generated, the Job will fail and retry. However, the Order has already been created and saved, and it will do so again on the following try, possibly resulting in 3 Orders when there should have only been one.

### DIFF
--- a/config/horizon.php
+++ b/config/horizon.php
@@ -146,7 +146,7 @@ return [
                 'queue' => ['default'],
                 'balance' => 'simple',
                 'processes' => 10,
-                'tries' => 3,
+                'tries' => 1,
             ],
         ],
 
@@ -156,7 +156,7 @@ return [
                 'queue' => ['default'],
                 'balance' => 'simple',
                 'processes' => 3,
-                'tries' => 3,
+                'tries' => 1,
             ],
         ],
     ],


piotrgrabowskielastique:
This is already the default behavior of laravel queue worker. Job retry should be opt-in instead of opt-out as its easier for new developers to understand whats happening and to debug.

browner12:
great point @piotrgrabowskielastique. If you're running a normal queue worker, and not Horizon, the default is 1 try.  Keeping it consistent is another good reason to switch.

GC-Mark:
So, if we are currently relying on jobs retrying 3 times by default, we now need to specify `public $tries = 3;` once this change is released. Am i correct?

Perhaps should have gone on the `4.0` branch.

JayBizzle:
So, if we are currently relying on jobs retrying 3 times by default, we now need to specify `public $tries = 3;` once this change is released. Am i correct?

Perhaps should have gone on the `4.0` branch.

driesvints:
@JayBizzle afaik you always publish your horizon config so it should be fixed in your config file.

JayBizzle:
Ah, yes, I get it now. Thanks :+1:

browner12:
correct, this will only apply to new projects publishing their config asset for the first time.

antimech:
@browner12 
`php artisan vendor:publish --tag="horizon-config" --force`

Issue #715, opened 2019-12-03
High memory usage in redis


mauri870:
- Horizon Version:  v3.4.3
- Laravel Version: v6.6.0
- PHP Version: 7.3.11
- Redis Driver & Version: predis/phpredis 1.1.1 or php-redis extension 5.1.1, same result
- Database Driver & Version:

### Description:

After we upgraded from Laravel 5.8 to 6.6 / Horizon 3.2.2 to 3.4.3 the resource consumption of our redis 4 server started to grow exponentially, sitting at around 5Gb.

Horizon Dashboard:
![horizon-dashboard](https://user-images.githubusercontent.com/10168637/70083601-77d27300-15eb-11ea-9da5-3718097068bd.png)

Redis instance dedicated to horizon:
![redis-usage](https://user-images.githubusercontent.com/10168637/70083581-7143fb80-15eb-11ea-8511-0aa624706bb3.png)

Previous version vs new (the release was pushed Nov 25):

![Screenshot_1575402597](https://user-images.githubusercontent.com/10168637/70084480-14494500-15ed-11ea-8a09-692f7fbf36f5.png)

### Steps To Reproduce:

We have no clue of what is causing this behavior, but our number of jobs is almost the same as before, the only difference is the framework and horizon versions.

SDekkers:
I have the same issue, the jobs when finished are not (seemingly) cleared from the redis memory, resulting in high memory usage. Running about 500k jobs per day

mauri870:
```php
    'trim' => [
        'recent' => 30,
        'recent_failed' => 30,
        'failed' => 60,
        'monitored' => 0
    ],
```

mauri870:
The problem seems to be that `trim.recent` is used to either expire a job that has been pushed(but not processed) and jobs that are already completed. Maybe it should be considered to add a new `trim.completed` to expire the completed jobs without loosing jobs that are still on queue.

mauri870:
For now our solution was to simple expire/delete the completed job to free up some memory:

```
Queue::after(function (JobProcessed $event) {
	Redis::expireat(config('horizon.prefix') . $event->job->getJobId(), Carbon::now()->addMinute()->timestamp);
});
```

mauri870:
We also got it working in a horizon fork:

```diff
diff --git a/config/horizon.php b/config/horizon.php
index b9803a8..318a945 100644
--- a/config/horizon.php
+++ b/config/horizon.php
@@ -98,6 +98,7 @@ return [
[...]

PR #720, opened 2019-12-09
[3.x] Allow trimming completed jobs

themsaid:
High memory consumption was reported in https://github.com/laravel/horizon/issues/715. Between v3.2.2 and the most recent version, I couldn't locate a change that may cause this.

However, the proposed solution in the issue seems useful. Basically, this PR adds the ability to purge completed jobs separately, so for example you may keep your recent jobs fo 60 minutes, failed jobs for 7 days, but purge completed jobs every 5 minutes.

One side effect of this change is that the list of recent jobs may include a lot of completed jobs that were purged but still exist in the `recent_jobs` set and thus paginated results will look strange.

One thing we can do is to have a separate list for:

- Pending Jobs
- Completed Jobs
- Failed jobs

### DIFF
--- a/config/horizon.php
+++ b/config/horizon.php
@@ -95,6 +95,7 @@ return [
 
     'trim' => [
         'recent' => 60,
+        'completed' => 60,
         'recent_failed' => 10080,
         'failed' => 10080,
         'monitored' => 10080,
--- a/src/Repositories/RedisJobRepository.php
+++ b/src/Repositories/RedisJobRepository.php
@@ -43,6 +43,13 @@ class RedisJobRepository implements JobRepository
     public $recentJobExpires;
 
     /**
+     * The number of minutes until completed jobs should be purged.
+     *
+     * @var int
+     */
+    public $recentCompletedExpires;
+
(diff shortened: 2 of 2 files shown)


taylorotwell:
Personally I'm not comfortable with the way this basically breaks pagination. I'm open to other suggestions though.

taylorotwell:
Personally I'm not comfortable with the way this basically breaks pagination. I'm open to other suggestions though.

neorganic:
This undermines `recent`. If you have a config from earlier versions with e.g. ` 'recent' => 1440, // 24 hours` , then it is being ignored (recent jobs won't be visible for 24 hours). It's a breaking change and should have been added to a major release only.

neorganic:
`completed` is also not mentioned in the docs @taylorotwell

themsaid:
@neorganic please upgrade to Horizon 4.0 where the issue doesn't exist.

neorganic:
@themsaid  Well that's not possible because my project is on 5.8 Laravel, but will have this in mind.

PR #736, opened 2019-12-30
Reset the retryUntil value

themsaid:


src/Jobs/RetryFailedJob.php
```diff
@@ -55,10 +56,26 @@ public function handle(Queue $queue, JobRepository $jobs)
...
+    /**
+     * Prepare the timeout.
+     *
+     * @param  array  $payload
+     * @return integer|null
+     */
+    protected function prepareNewTimeout($payload)
+    {
+        return $payload['timeoutAt']
+                        ? Chronos::now()->addSeconds(ceil($payload['timeoutAt'] - $payload['pushedAt']))->getTimestamp()
```
satyago: @themsaid that is a nice solution, very good, thank you !

We had solved this internally by just overriding the `RetryFailedJob` class and setting `timeoutAt` to `null`, but this looks like a better approach indeed. But that's exactly why I haven't pushed a pull request; I didn't feel like I could make the best code decisions while contributing to Horizon just yet.

satyago:
I know I'm not supposed to review stuff, but I did haha.
Thank you !

PR #741, opened 2020-01-09
[3.x] Revert filtering by tag

driesvints:
This PR reverts https://github.com/laravel/horizon/pull/665, https://github.com/laravel/horizon/pull/678 and https://github.com/laravel/horizon/pull/692. https://github.com/laravel/horizon/pull/665 was causing high memory loads in particularly cases (see https://github.com/laravel/horizon/issues/740).

Until we can provide a fix for this implementation that fixes the memory load we've decided to revert the changes for now.

Fixes https://github.com/laravel/horizon/issues/740

Issue #740, opened 2020-01-08
High Memory Usage since v3.4.4


BramRoets:
- Horizon Version: 3.4.4
- Laravel Version: 5.8.35
- PHP Version: 7.2
- Redis Driver & Version: predis/phpredis v1.1.1
- Database Driver & Version: Redis v5.0.0

### Description:
I've updated Laravel Horizon from v3.2.2 to v3.4.4 at 21 nov 2019. I'm seeing an increase in memory usage since then.

Available Memory, free memory started dropping after 21 november

![Screenshot from 2020-01-08 10-33-36](https://user-images.githubusercontent.com/2612285/71967037-f467fb00-3202-11ea-94b7-0c26dd6a63f5.png)

Debugging memory usage with RedisInsight show the recent keys uses almost all memory available.

![Screenshot from 2020-01-08 09-57-53](https://user-images.githubusercontent.com/2612285/71967255-57599200-3203-11ea-8d03-704a49bb0c15.png)
![Screenshot from 2020-01-08 11-23-09](https://user-images.githubusercontent.com/2612285/71970589-4744b100-3209-11ea-9474-e973ab1c00e7.png)

### Steps To Reproduce:
1) Run 50.000 jobs per hour, jobs have different models (using SerializesModels) attached to them, no tags are configured in Horizon.
2) Config horizon.trim.recent = 60 minutes
3) Notice high memory usage

I'm not 100% sure why the memory usage keeps rising over a longer period of time. We do have some additional load over the last months, but not enough to cause this.

I've looked at https://github.com/laravel/horizon/issues/715 and i've used the following piece of code: 

```
Queue::after(function (JobProcessed $event) {
	Redis::expireat(config('horizon.prefix') . $event->job->getJobId(), Carbon::now()->addMinute()->timestamp);
});
```

That did not seem to make any difference. The recent keys still have a TTL of 60 minutes.

If I change horizon.trim.recent to 5 minutes. Will this solve my problem? I'm not sure what will happen with pending jobs when:
- They get release back on the queue e.g. `$this->release(300)`
- When the Laravel Horizon workers are down for more than 5 minutes (planned maintenance, ..)
- A burst of jobs, causing a long queue longer than 5 minutes.

Prashank25:
I am having this issue as well, it seems to be the `horizon:recent:TAGNAME` keys in my case, they seem to keep growing and never get cleared? Am I missing something?

BramRoets:
[...]

PR #743, opened 2020-01-13
[3.x] Trim job IDs from failed job tags

sebdesign:
In light of #715 and #740, and after reading the insights from @travisaustin, I think I found where the memory hog comes from.

Everytime a job fails, the ID is added to each tag. If there is no other failure in the next 48 hours, the tags expires.  But otherwise, the `StoreTagsForFailedJob` keeps adding job IDs and updates the expiration date, but the previous IDs were **never** removed, which probably causes memory issues.

In this commit we'll remove the job IDs for that are older than 48 hours for each tag, in respect of the TTL in `StoreTagsForFailedJob`.

For example:
- On `2020-01-01 00:00:00` a job with ID of `1` and tagged `App\User:1` fails. The tag `failed:App\User:1` is added with the job ID 1.
- So after 2880 minutes, i.e. on `2020-01-03 00:00:00` the `failed:User:1` tag should expire.
- Meanwhile, on `2020-01-02 00:00:00` another job with ID of `2` and tagged `App\User:1` fails. The ID `2` is added to the `failed:App\User:1` tag, and the expiration of the tag is delayed for 2880 minutes, i.e. `2020-01-04 00:00:00`.
- If there are more failures in the next 2880 minutes, the new IDs will keep getting appended, and the previous IDs are never removed.

Now with this PR:
- If there is a failure on `2020-01-03 00:00:00` for a job with ID of `3`, it will be added as usual to the `failed:App\User:1` tag, along with the existing IDs 1 and 2, but now the score is the precise timestamp that it was added, like in `RedisJobRepository@storeJobReference`.
- Then, it will trim all the job IDs that were added 2880 minutes ago or earlier from this tag, like in `RedisJobRepository@trimFailedJobs`. So the id `1` will be removed.

If this is correct, then we could revert #741 and trim the job IDs from the `recent:` tags each time a job is added, like with the `failed:` tags.

### DIFF
--- a/src/Contracts/TagRepository.php
+++ b/src/Contracts/TagRepository.php
@@ -87,4 +87,12 @@ interface TagRepository
      * @return void
      */
     public function forget($tag);
+
+    /**
+     * Trim the job IDs from the given tags.
+     *
+     * @param  array  $tags
+     * @return void
+     */
+    public function trim(array $tags);
 }
--- a/src/Listeners/StoreTagsForFailedJob.php
+++ b/src/Listeners/StoreTagsForFailedJob.php
@@ -40,5 +40,7 @@ class StoreTagsForFailedJob
         $this->tags->addTemporary(
             2880, $event->payload->id(), $tags
         );
+
+        $this->tags->trim($tags);
     }
 }
--- a/src/Repositories/RedisTagRepository.php
+++ b/src/Repositories/RedisTagRepository.php
@@ -2,6 +2,7 @@
 
 namespace Laravel\Horizon\Repositories;
 
+use Cake\Chronos\Chronos;
 use Illuminate\Contracts\Redis\Factory as RedisFactory;
 use Laravel\Horizon\Contracts\TagRepository;
 
@@ -96,7 +97,7 @@ class RedisTagRepository implements TagRepository
     {
         $this->connection()->pipeline(function ($pipe) use ($minutes, $id, $tags) {
             foreach ($tags as $tag) {
-                $pipe->zadd($tag, $id, $id);
+                $pipe->zadd($tag, str_replace(',', '.', microtime(true) * -1), $id);
 
                 $pipe->expire($tag, $minutes * 60);
             }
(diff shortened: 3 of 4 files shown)


travisaustin:
I think you are mostly correct, but I think you are only solving have the issue.

For example, in my production environment, the Redis keys that are filling up are:

`horizon:failed:<TAG FOR MODEL CLASS NAME>:<ID OF MODEL>`
`horizon:recent:<TAG FOR MODEL CLASS NAME>:<ID OF MODEL>`

In my case, each of these keys has a TTL of 15 minutes. The problem is that I create a job of each ID of each model every 2 minutes (our workflow polls a remote system for each model every 2 minutes). So the TTL starts at 15 minutes, and when it hits ~13 minutes, a new job is dispatched for the model, which adds a new Job ID to that key, and the TTL is reset to 15 minutes. Since it never times out, my keys have 50,000 or more job IDs _each_ even though Horizon only shows about 1,000 recent jobs.

I have not tested your PR, but, based on your writeup, you are only removing keys from `horizon:failed`. I would like to see `horizon:recent` trimmed, as well.

Thanks for the PR. I am excited to have it!!

sebdesign:
@travisaustin the `horizon:recent` tags feature was removed in #741, that's why i'ts not implemented in this PR.

How did you set the TTL to 15 minutes? The TTL for the failed tags is hardcoded to 2880 seconds [here](https://github.com/laravel/horizon/blob/v3.4.6/src/Listeners/StoreTagsForFailedJob.php#L41).

Regarding your example, if this PR gets merged, you won't accumulate 50,000 IDs per key, because each time a job ID is added to the queue, the old IDs (now() - 2880 min) will be removed. Redis will know which IDs to remove because they are now added with a score that matches the timestamp they were added. I hope that makes sense to you!

src/Contracts/TagRepository.php
```diff
@@ -87,4 +87,12 @@ public function paginate($tag, $startingAt = 0, $limit = 25);
...
      */
     public function forget($tag);
+
+    /**
+     * Trim the job IDs from the given tags.
+     *
+     * @param  array  $tags
+     * @return void
+     */
+    public function trim(array $tags);
```
driesvints: We can't add this in a minor release because it'll break implementations who don't have this method defined.

sebdesign: That means when `trim` is called it will also break those implementations, since they don't define this method?

driesvints: It'll break as soon as they install this package because the implementation won't adhere to the contract;

taylorotwell:
I think I will just hold off on this for now.

src/Contracts/TagRepository.php
sebdesign: I mean if we didn't add it to the contract, the `trim` method would fail during runtime because it's probably not defined on the implementation, when it's called on the listener.

PR #749, opened 2020-01-23
Allow Horizon to be used on a subdomain without a subfolder

shabushabu:
I've just started on a project where I'd like to use Nova in the root of the `admin` subdomain, Telescope on the `telescope` subdomain and Horizon on the `horizon` subdomain. Nova and Telescope can handle this without issues, but Horizon ended up linking to URLs like `https://monitoring` when `domain` was set to `horizon.domain.com` and `path` was set to `/` in the config file.

So, I looked at the Horizon and Telescope codebases, figured out why Telescope could handle these config settings and then just added those Telescope bits to Horizon.

Would be great if you could merge this. Cheers!

### DIFF
--- a/resources/js/app.js
+++ b/resources/js/app.js
@@ -21,10 +21,19 @@ window.Popper = require('popper.js').default;
 
 Vue.prototype.$http = axios.create();
 
+window.Horizon.basePath = '/' + window.Horizon.path;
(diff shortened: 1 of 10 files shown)


shabushabu:
Thanks!

Issue #388, opened 2018-09-10
Configurable Tag Display on Recent Jobs 


ragingdave:
Some Jobs might have hundreds of model ids in cases where bulk queuing is in use. This makes the recent jobs ui basically useless as each recent job takes up a full window height.

Any chance that could be converted to something like a tooltip for all and then maybe a simplified tag display?

driesvints:
Hmm yeah I can see how this can be improved. Just for anyone who's wondering how this would look:

![screen shot 2018-10-23 at 4 31 42 pm](https://user-images.githubusercontent.com/594614/47368032-2e71dd80-d6e1-11e8-8a0d-98ac9da68999.png)

I've tagged this as a feature request so we can take a look at a later time. If you're up for it, feel free to send in a PR.

pactode:
Would it make sense to show a count of the tags instead? Then only show the tags in the tooltip. 

Something similar to:
![screenshot 2019-01-10 at 11 50 37](https://user-images.githubusercontent.com/5956778/50963814-09199200-14ce-11e9-922c-b39416b16085.png)

Let me know your thoughts.

themsaid:
No plans to change this at the moment since I don't think it's a common problem.

driesvints:
Talked to @themsaid about this and we're going to keep this open for now. I personally think clipping the content and clicking to show the rest is a good solution but feel free to say a different proposal.

ragingdave:
So v3 is SO much better but I feel like there could still be improvement for batch jobs. Here is an example of a normally occurring batch job and how the tag display ends up looking (again way better in v3 but still):

![screen shot 2019-03-05 at 9 16 50 am](https://user-images.githubusercontent.com/1168344/53811444-6f97bb00-3f27-11e9-8efd-59480d42c5cd.png)

And there are more tags than in that screenshot for that single job

driesvints:
https://github.com/laravel/horizon/pull/754 introduces a new way to display tags.

PR #751, opened 2020-01-31
[3.x] Ability to view job details in recent jobs overview

Cannonb4ll:
I sometimes find myself in a situation where I need to know what data is passed into the job (whatever its status was).

This is similar to reading failed job data in the Failed Jobs tab.

Img
Img


driesvints:
@Cannonb4ll this looks really good 👍 

I was thinking that maybe we can also tackle https://github.com/laravel/horizon/issues/388 together with this? Maybe in a new "Tags" panel below the "Data" pannel? How does the "Tags" data now look in the first panel with lots and lots of tags as displayed in the issue?

Cannonb4ll:
@driesvints Oh dear, I have not checked how it would look with a lot of tags. I will check this, please wait before merging as I could do some nice stuff with this 👏

Will send new updates in this PR.

Cannonb4ll:
@driesvints What do you think of this?

The overview shows a max of 3 tags, and then if there's more than 3 it will show (50 more). The job detail view (which I created earlier) will then also show the tags via vue-pretty-print at the bottom.

I have also added a collapse feature (default from bootstrap, does not come with any additional libs) to make it more easier to quickly read the tags card if you have a lot of data.

See screenshots below for reference, let me know what you think!

img
img2

themsaid:
There will always be people who want to see all tags in the table even if the layout isn't the best, specially that one tag name can already break the layout if it's really long `App\Forge\Servers\Networking\FirewallRules\Rule:12341`.

driesvints:
@Cannonb4ll I've talked to @themsaid about this and it's best that we do the tags box in a different PR. Can you revert your last commit and send in a separate PR for it once this one is merged? Thanks!

Cannonb4ll:
Reverted.

driesvints:
Thanks!

driesvints:
@Cannonb4ll feel free to send in a follow up pr!

Cannonb4ll:
@driesvints I would, but then probably only for the tags on the job detail, not the index I think? As @themsaid did not agree on limiting results on the index.

driesvints:
@Cannonb4ll feel free to send in pr with the changes to the index you made.

brunogaspar:
@Cannonb4ll Noticed a small issue, if the job hasn't run or is running, the completed at shows `1970-01-01 01:00:00` :)

I can try to send a pr to address the issue tomorrow or so, just letting you know in case you want to fix it, not a big deal for now hehe

Cannonb4ll:
No need, thats fixed in #755

If that wont be merged ill send it in loose.

PR #754, opened 2020-02-04
[3.x] Another way for tags displaying & collapsable panels

Cannonb4ll:
As I showed in PR #751 the index in this PR will now display a max of 3 tags and a small label if there are more with (53 more).

Then in the job detail view from PR #751 I display the tags at the bottom in vue-json-pretty.

I have also taken the opportunity to make the panels collapsable in the job detail so whenever there's a lot of data you can collapse that.

Screens to show what I mean:

img
img2

### DIFF
--- a/resources/js/screens/recentJobs/index.vue
+++ b/resources/js/screens/recentJobs/index.vue
@@ -132,16 +132,19 @@
                 <h5>Recent Jobs</h5>
             </div>
 
-            <div v-if="!ready" class="d-flex align-items-center justify-content-center card-bg-secondary p-5 bottom-radius">
+            <div v-if="!ready"
+                 class="d-flex align-items-center justify-content-center card-bg-secondary p-5 bottom-radius">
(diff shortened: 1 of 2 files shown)


driesvints:
This will solve https://github.com/laravel/horizon/issues/388

driesvints:
@Cannonb4ll thanks for all your work on these prs!

Cannonb4ll:
@driesvints No problem, I have a lot more idea's actually. Also, in favor of the earlier comment about too long tags. Wouldn't horizontal navigation solve this? (We could also make this configurable)

img

driesvints:
I'm not sure. It looks good but it would be different from Telescope that way.

Cannonb4ll:
Yes true, I will leave that as it is.

vigneshgurusamy:
@Cannonb4ll Instead of using **View detail** as a hyperlink, we could use the Job name as a hyperlink and UX will be similar to failed job listing page.

jandante:
Great work @Cannonb4ll -> Very handy additions in the 3.6 update!

PR #755, opened 2020-02-04
[3.x] Ability to see if a job is delayed

Cannonb4ll:
Sometimes I find myself in a situation where I want to see if a job was dispatched with a delay, this PR allows you to see if it is via a badge.

Examples:

img
img2

### DIFF
--- a/src/Http/Controllers/RecentJobsController.php
+++ b/src/Http/Controllers/RecentJobsController.php
@@ -38,7 +38,7 @@ class RecentJobsController extends Controller
         $jobs = $this->jobs->getRecent($request->query('starting_at', -1))->map(function ($job) {
             $job->payload = json_decode($job->payload);
 
-            $job->delayed = optional(unserialize($job->payload->data->command))->delay ? true : false;
+            $job->delayed = strtotime(optional(unserialize($job->payload->data->command))->delay);
 
             return $job;
         })->values();
--- a/src/Http/Controllers/RecentJobsController.php
+++ b/src/Http/Controllers/RecentJobsController.php
@@ -38,8 +38,6 @@ class RecentJobsController extends Controller
         $jobs = $this->jobs->getRecent($request->query('starting_at', -1))->map(function ($job) {
             $job->payload = json_decode($job->payload);
 
-            $job->delayed = strtotime(optional(unserialize($job->payload->data->command))->delay);
-
             return $job;
         })->values();
 
@@ -65,8 +63,6 @@ class RecentJobsController extends Controller
     protected function decode($job)
     {
         $job->payload = json_decode($job->payload);
-
-        $job->delayed = strtotime(optional(unserialize($job->payload->data->command))->delay);
 
         return $job;
     }
--- a/src/Http/Controllers/RecentJobsController.php
+++ b/src/Http/Controllers/RecentJobsController.php
@@ -38,6 +38,8 @@ class RecentJobsController extends Controller
         $jobs = $this->jobs->getRecent($request->query('starting_at', -1))->map(function ($job) {
             $job->payload = json_decode($job->payload);
 
+            $job->delayed = optional(unserialize($job->payload->data->command))->delay ? true : false;
+
             return $job;
         })->values();
 
@@ -63,6 +65,8 @@ class RecentJobsController extends Controller
     protected function decode($job)
     {
         $job->payload = json_decode($job->payload);
+
+        $job->delayed = strtotime(optional(unserialize($job->payload->data->command))->delay);
 
         return $job;
     }
--- a/public/app-dark.css
+++ b/public/app-dark.css
@@ -10941,7 +10941,3 @@ button:hover .fill-primary {
   background: #aa2e28;
 }
 
-.badge-sm {
-  font-size: 0.75rem;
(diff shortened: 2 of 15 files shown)


driesvints:
This looks okay but I would use a different color since blue is often associated with something that goes okay. Maybe the same yellow from the pause symbol? Also I'd just say "delayed" and not "delayed job". Maybe add that timestamp on hover?

Cannonb4ll:
Like this?

img

driesvints:
That looks better imo but let's see what others say :)

EvanHandgraaf:
> This looks okay but I would use a different color since blue is often associated with something that goes okay. Maybe the same yellow from the pause symbol? Also I'd just say "delayed" and not "delayed job". Maybe add that timestamp on hover?

I would go with the purple Horizon color because it's more calm to look at. Furthermore a "delayed job" is expected behaviour, its not an error.

EvanHandgraaf:
> This looks okay but I would use a different color since blue is often associated with something that goes okay. Maybe the same yellow from the pause symbol? Also I'd just say "delayed" and not "delayed job". Maybe add that timestamp on hover?

I would go with the purple Horizon color because it's more calm to look at. Furthermore a "delayed job" is expected behaviour, it's not an error.

Cannonb4ll:
To sum up the colors, my personal favorite is the purple (primary) one.

img
img
img

tomswinkels:
@Cannonb4ll first, very nice!! I like this PR!

- I think it is better to say Delayed (all in the list are jobs nothing else)
- My favorite color the orange one
- The suggestion of @driesvints with a hover on Delayed i think this is nice to have :)

Cannonb4ll:
Just committed the tooltip on hover:

img

For this I had to move popper.js import above the bootstrap import [because the docs say so](https://getbootstrap.com/docs/4.0/components/tooltips/) and jQuery import to the app.js in order for the tooltips to work.

Then I made a directive for the tooltip so each new record would also get the tooltip instance bound to its element. (Otherwise new records that show up dynamically will not have the tooltip anymore)

This also resulted that I removed the jQuery imports from the specific components. I also took the liberty to reduce JS size by importing the specific lodash package "take" in "Stacktrace.vue" as an example. We could do this with more components as I noticed the full lodash library is imported there which gives a lot of overhead of JS.

Let me know what you think!

taylorotwell:
It's weird to me to show this label for a job that is already finished processing. It only makes sense to me to see it while pending and even better if it shows a human readable thing like "Delayed for 10 minutes"

tomswinkels:
@taylorotwell i think you right. It is better to show the label when the delay is not passed. But where do you want to show the "Delayed for 10 minutes"? Only in the hover?

Cannonb4ll:
I agree with you @taylorotwell, which option do you prefer more?

Option 1:

img

Option 2:

img2

tomswinkels:
If you choose option 1 than remove the "Delayed for" in the label. The label are say that it is a delayed job.

taylorotwell:
I'm fine with the tooltip... I think purple is a bit "in your face" for this informational stuff... I would have gone with a gray background or something.

Cannonb4ll:
We can grey these out, with `badge-secondary`:

img

taylorotwell:
Is this ready for review on your end?

Cannonb4ll:
Yes, it is. Not sure why the travis builds fail randomly though.

Cannonb4ll:
Sorry, quickly added 1 more small detail in the job detail screen, the Pushed At:

img

src/Http/Controllers/RecentJobsController.php
```diff
@@ -38,6 +38,8 @@ public function index(Request $request)
         $jobs = $this->jobs->getRecent($request->query('starting_at', -1))->map(function ($job) {
             $job->payload = json_decode($job->payload);
 
+            $job->delayed = strtotime(optional(unserialize($job->payload->data->command))->delay);
```
taylorotwell: If you are unserializing the entire command to get this won't that re-issue the database queries needed to get the Eloquent models / collections required by the job?

Cannonb4ll: @taylorotwell Just tried, indeed it will introduce queries to be run whenever a model is type hinted into the job class, this will introduce reduced performance once there are jobs with a lot of models.

Is there a way to disable this magic once we call an unserialise on the payload?

taylorotwell: Not that I can think of off of the top of my head. It may be possible to extract the delay from the raw serialized string but I'm not sure.

Cannonb4ll: I am going to think and try some stuff out. Will get back on this, thanks, good find 👍

Cannonb4ll:
Ok, requesting a review again for this one. 

What I did is remove the PHP way with unserialising as this was causing DB queries. Instead I used the phpunserialize JS package which was already included.

In order to do this I exported each TR row to its own component to make proper use of computed records. Now there is no DB queries anymore.

tomswinkels:
I think it is a good way. @taylorotwell can you check again?

fgilio:
Thank you for this!

PR #760, opened 2020-02-11
[3.x] Pass the queued job to the notification tags function

ctyler-liquidfish:
I added an envelope (anonymous class) around SendQueuedNotification jobs in the Tags class, for the purpose of passing the job to the notification when getting tags, so the recipient and channel can be included in tags.

In my application, I'm using this change with this trait I add to my notifications:
```php
<?php

namespace App\Traits;

use App\Models\User;
use Illuminate\Notifications\SendQueuedNotifications;

trait hasNotificationTags
{
    /**
     * @param SendQueuedNotifications $job
     * @return array
     */
    public function tags($job = null){
        if(!empty($job) && $job instanceof SendQueuedNotifications){
            return [
                '#'.(head($job->channels)??'?'),
                ($job->notifiables instanceof User ? 'User:'.$job->notifiables->id : get_class($job->notifiables).':'.$job->notifiables->id??'?'),
            ];
        } else {
            return [];
        }
    }
}
```
Which adds a `User:{id}` and a `#{channel}` tag to the job. I don't actually need the conditionals, afaik. Anyway, there was just no other way to add the channel to the tags except to modify the Tags class.

### DIFF
--- a/src/Tags.php
+++ b/src/Tags.php
@@ -119,7 +119,7 @@ class Tags
         $models = [];
 
         foreach ($targets as $target) {
-            if($target instanceof static::$notificationTagEnvelopeClass) {
+            if(static::$notificationTagEnvelopeClass && $target instanceof static::$notificationTagEnvelopeClass) {
                 $target = $target->job->notification;
             }
             $models[] = collect(
--- a/src/Tags.php
+++ b/src/Tags.php
@@ -88,11 +88,25 @@ class Tags
             case $job instanceof SendQueuedMailable:
                 return [$job->mailable];
             case $job instanceof SendQueuedNotifications:
-                return [$job->notification];
+                return [
+                    static::$notificationTagEnvelopeClass ?
+                        new static::$notificationTagEnvelopeClass($job) :
+                        static::$notificationTagEnvelopeClass = new class ($job){
+                            public $job;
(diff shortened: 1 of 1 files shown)


taylorotwell:
I'm sorry it's just really hard for me to follow this PR. Maybe you can use the `Queue::before(fn)` hook to set some variables that you can access from your notification for the current job.

ctyler-liquidfish:
I can't use Queue::before, it's executed well _after_ the job is added to the queue, while the tags are set _as_ the job is added to the queue.
Look, the basic problem is this: When I send a notification along multiple channels, Horizon displays one job for each channel, but there's no way to tell which channel the job is on until/unless it _fails_. I want to add the channel to the tags, but the only place to hook into that is the Tags class itself, and it's not created via service container, so I can't override it.

The purpose of the PR is to pass the SendQueuedNotification job, which knows the recipient and channel, to the Notification's `tags` function so they can be used in tag generation. From there, the rest of the changes are for the purpose of allowing everything else to continue as intended.

I supposed it would be simpler if I didn't use an Anonymous class, but I was trying to restrict my changes to a minimum, and it seemed like adding an additional class file was more than I should do. 🤷

Maybe I'll resubmit after refactoring.

PR #763, opened 2020-02-13
[3.x] Pass the job-in-queue to the tags() function.

ctyler-liquidfish:
Primarily for the purpose of adding the notification channel as a tag, pass the actual queued job to the target when calling the tags function. This also allows a Notification to access the Notifiable while generating tags.

Additionally, my apologies for my previous (vastly over-complicated) pull request. Once I took a look at refactoring, this much simpler solution became apparent.

### DIFF
--- a/src/Tags.php
+++ b/src/Tags.php
@@ -41,7 +41,7 @@ class Tags
     {
         return $job instanceof CallQueuedListener
                     ? static::tagsForListener($job)
-                    : static::explicitTags(static::targetsFor($job),$job);
+                    : static::explicitTags(static::targetsFor($job), $job);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


taylorotwell:
Use the Queue events like I mentioned in the previous thread.

ctyler-liquidfish:
@taylorotwell The queue events are not raised when a job is added to the queue. They don't provide sufficient hook to add these tags when the job is added.

ctyler-liquidfish:
Oh, I see, _Horizon_ adds more event hooks. That aren't mentioned in documentation.
Well, in case anyone else comes across this and wants a solution:
```php
Event::listen(JobPushed::class,function(JobPushed $event){
    if($event->payload->commandName() === SendQueuedNotifications::class){
        $job = unserialize($event->payload['data']['command']);
        // Do stuff
        $event->payload->set(['tags'=>array_merge($event->payload['tags'], [
            // add tags
        ])]);
    }
});
```

PR #765, opened 2020-02-16
[3.x] Use 'default' connection as a default redis connection

binaryk:
This PR will fix the issue when the `horizon` is used as a dependency on another package. As an example, it was used by @freekmurze (https://github.com/freekmurze) in the https://mailcoach.app/ package, and the installation was stopped because of the value received in the `Laravel\Horizon@use` method was `null` (since I didn't have the `config/horizon.php` file locally).

### DIFF
--- a/src/HorizonServiceProvider.php
+++ b/src/HorizonServiceProvider.php
@@ -127,7 +127,7 @@ class HorizonServiceProvider extends ServiceProvider
             __DIR__.'/../config/horizon.php', 'horizon'
         );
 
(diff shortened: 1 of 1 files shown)


themsaid:
What do you mean by stopped? if it's an error can you please share what you get?

binaryk:
@themsaid the error bellow is shown after installation horizon via third party package:
image
That's because in the `Horizon.php` we receive the connection as null if `artisan horizon:install` was not run before:
image

PR #767, opened 2020-02-17
[4.x] Add separate screen for completed jobs

themsaid:
This PR adds separate screens for Pending and Completed Jobs.

Trimming of pending and completed jobs is separate from trimming of recent jobs.

```
'recent' => 1440,
'completed' => 15,
'pending' => 60,
'recent_failed' => 10080,
'failed' => 10080,
'monitored' => 10080,
```

Recent jobs sorted set is only used to show the count of recent jobs in the dashboard screen.

Screen Shot 2020-02-17 at 7 44 11 PM

---

Screen Shot 2020-02-17 at 7 44 01 PM

---

Screen Shot 2020-02-17 at 7 44 06 PM


brunogaspar:
Thanks @themsaid , looks awesome!

Cannonb4ll:
So a few days later having worked with Horizon V4 I am not too sure if this was a right move, which I can elaborate.

It feels like I have to do way more clicks now. If I am on the "Pending" overview, the job gets processed and the list will be emptied. But how do I know if it failed or succeeded? It feels more cumbersome this way. The old overview would let you know it failed and was clickable right away.

A beter solution would be to have it like it was, with filter options.

Wonder what others think about this.

driesvints:
@Cannonb4ll the main reason for this was because it was very hard to do this in combination with how data was stored in Redis, especially for projects with lots and lots of jobs. In a queueing monitoring system like Horizon you're usually not watching the pending screen very often. We're going to leave it like this for now.

If you can come up with a proper solution to the problems that were caused by the previous screen and group everything you're free to send in a PR.

Cannonb4ll:
@driesvints Thank you for answering. I will give this some more thought and might prepare a concept PR. I think we can optimise a lot UX wise 🙏

PR #776, opened 2020-02-21
[3.x] Switch to Github Actions

driesvints:


browner12:
Why do you have `fail-fast` set to `true`?  Wouldn't we want to continue running all the matrix builds so we can see which ones pass and which ones fail?

driesvints:
It's better to see a build result faster so you can act faster accordingly and so builds don't stack up.

PR #783, opened 2020-03-03
[4.x] Show warning when manifest is outdated

andresayej:
This adds a check to the base view to show a warning when the manifest file in the public folder is different from the one in the vendor dir. This prevents broken components or missing updates.

This PR is with the same implementation, title, and descriptions as done by @barryvdh for [Telescope](https://github.com/laravel/telescope/pull/729).

@driesvints I've reviewed the telescope PR and saw in your last reply that you encourage people to try a PR for Horizon to see if @taylorotwell will accept it, so this is it 🙌.

### DIFF
--- a/resources/views/layout.blade.php
+++ b/resources/views/layout.blade.php
@@ -92,6 +92,12 @@
             </div>
 
             <div class="col-10">
+                @if (! $assetsAreCurrent)
+                    <div class="alert alert-warning">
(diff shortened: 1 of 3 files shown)


driesvints:
I still feel we should try to force update the assets when horizon is updated through Horizon but I guess this'll do for now. Thanks for your pr.

PR #794, opened 2020-03-13
Add metrics options and dark mode config

bert-w:
I wanted to have some more control over how the Metrics graphs were shown. It turns out only the last 24 snapshots are saved, so when creating a snapshot every 5 minutes that would result in (just) 2 hours of data. With the new `metrics.trim_snapshots.job` option, you are able to keep an x amount of snapshots, thus increasing the amount of points on the graph and showing you a bigger timespan.

The following screenshots show the new metrics, with about 200 plot points (which is working on my machine without performance impact). Notice changed colors and the new date format.
dark
light

In addition, I added the `'theme' => 'light'` config, because I think this file should be the place to define such options (not in the service provider).

Feedback is very welcome. I've PRed this against version 3.0, I'm not sure if that is allowed? The changes are backwards compatible, and I'd like to have it on the 3.x branch because I can't yet migrate to Laravel 7.x due to dependencies.

### DIFF
--- a/config/horizon.php
+++ b/config/horizon.php
@@ -103,6 +103,38 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Metrics
+    |--------------------------------------------------------------------------
+    |
+    | Here you can configure how many snapshots should be kept to display in
+    | the metrics graph. You should use this in conjunction with the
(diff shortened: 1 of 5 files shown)


taylorotwell:
Please send to 4.x branch.

PR #795, opened 2020-03-13
[4.x]  Add metrics options and dark mode config

bert-w:
I wanted to have some more control over how the Metrics graphs were shown. It turns out only the last 24 snapshots are saved, so when creating a snapshot every 5 minutes that would result in (just) 2 hours of data. With the new `metrics.trim_snapshots.job` option, you are able to keep an x amount of snapshots, thus increasing the amount of points on the graph and showing you a bigger timespan.

The following screenshots show the new metrics, with about 200 plot points (which is working on my machine without performance impact). Notice changed colors and the new date format.
dark
light

In addition, I added the `'theme' => 'light'` config, because I think this file should be the place to define such options (not in the service provider).

Feedback is very welcome.

### DIFF
--- a/config/horizon.php
+++ b/config/horizon.php
@@ -109,6 +109,38 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Metrics
+    |--------------------------------------------------------------------------
+    |
+    | Here you can configure how many snapshots should be kept to display in
+    | the metrics graph. You should use this in conjunction with the
+    | `horizon:snapshot` schedule to define how long of a timespan
+    | you want to see in the metrics.
+    |
+    */
+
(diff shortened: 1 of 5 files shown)


driesvints:
This PR kind of broke how `Horizon::night()` works, which can't force night mode anymore.

driesvints:
See https://github.com/laravel/horizon/issues/818

bert-w:
Hmm so the `Horizon.php` class is probably constructed after the `Horizon::night()` call has taken place.

themsaid:
Hey @bert-w 

I've reverted that part in your PR. Please submit a new PR that works if you want.

PR #802, opened 2020-03-19
[4.x] Align design with Pending/Completed jobs screen

danijelk:
With this PR we align the monitoring screen to reflect changes of v4 for completed/pending screen pages.

Currently the view-details doesn't work on monitoring screen as it got changed in between v3->v4.


driesvints:
Can you share screenshots of the before/after?

PR #809, opened 2020-03-26
[4.x] Graceful shutdown on SIGTERM also

zorn-v:
This can be useful when run horizon in docker for example. 
Docker by default send SIGTERM on stop container

### DIFF
--- a/src/Console/HorizonCommand.php
+++ b/src/Console/HorizonCommand.php
@@ -54,11 +54,13 @@ class HorizonCommand extends Command
 
         pcntl_async_signals(true);
 
-        pcntl_signal(SIGINT, function () use ($master) {
+        $termFunc = function () use ($master) {
             $this->line('Shutting down...');
 
             return $master->terminate();
-        });
+        }
+        pcntl_signal(SIGINT, $termFunc);
+        pcntl_signal(SIGTERM, $termFunc);
 
         $master->monitor();
     }
--- a/src/Console/HorizonCommand.php
+++ b/src/Console/HorizonCommand.php
@@ -58,7 +58,7 @@ class HorizonCommand extends Command
             $this->line('Shutting down...');
 
             return $master->terminate();
-        }
+        };
         pcntl_signal(SIGINT, $termFunc);
         pcntl_signal(SIGTERM, $termFunc);
 
--- a/src/Console/HorizonCommand.php
+++ b/src/Console/HorizonCommand.php
@@ -54,13 +54,14 @@ class HorizonCommand extends Command
 
         pcntl_async_signals(true);
 
-        $termFunc = function () use ($master) {
+        $terminate = function () use ($master) {
             $this->line('Shutting down...');
 
             return $master->terminate();
         };
-        pcntl_signal(SIGINT, $termFunc);
-        pcntl_signal(SIGTERM, $termFunc);
+
+        pcntl_signal(SIGINT, $terminate);
+        pcntl_signal(SIGTERM, $terminate);
 
         $master->monitor();
     }


driesvints:
Please add a more thorough description about what this exactly does. Thanks!

zorn-v:
Currently only SIGINT signal (e.g. CTRL+C) handled for graceful shutdown (no pick up new jobs, but wait until current running jobs will done). Same what `horizon:terminate` do.
This PR make same behavior on SIGTERM signal. This signal is default for programs like `kill` and default signal which docker send to process in container while stop it.

driesvints:
We reverted this for now because we're considered this could trigger Supervisor to SIGINT and SIGTERM later on.

zorn-v:
What mean "could trigger" ? It is console command (`./artisan horizon`) class, and it can handle signals only from outer world.

djtarazona:
> We reverted this for now because we're considered this could trigger Supervisor to SIGINT and SIGTERM later on.

Hey @driesvints, can you elaborate on this consideration?

I need this as well as we run Horizon in Docker containers on Kubernetes. Best practices for Docker is to not run Supervisor. Docker / Kubernetes acts like the process manager itself. It the Horizon process stops running, Kubernetes will notice and will start another container running Horizon — no need for Supervisor.

djtarazona:
Hmm, after some more digging, we may be good here. `HorizonCommand` [calls `MasterSupervisor::monitor()`](https://github.com/laravel/horizon/blob/28c9d04d9c444833d6dbbe3c05a02f6f5941aea2/src/Console/HorizonCommand.php#L63), that eventually [calls `listenForSignals()`](https://github.com/laravel/horizon/blob/28c9d04d9c444833d6dbbe3c05a02f6f5941aea2/src/MasterSupervisor.php#L205) which [listens for `SIGTERM`](https://github.com/laravel/horizon/blob/28c9d04d9c444833d6dbbe3c05a02f6f5941aea2/src/ListensForSignals.php#L25). If a `SIGTERM` is received, the `MasterSupervisor` [records the signal](https://github.com/laravel/horizon/blob/28c9d04d9c444833d6dbbe3c05a02f6f5941aea2/src/ListensForSignals.php#L26) and processes it just a bit later in the [`loop()`](https://github.com/laravel/horizon/blob/4.x/src/MasterSupervisor.php#L237). When the `SIGTERM` is processed, [`terminate()` is called](https://github.com/laravel/horizon/blob/28c9d04d9c444833d6dbbe3c05a02f6f5941aea2/src/ListensForSignals.php#L52).

In the end, I think `php artisan horizon` / the `MasterSupervisor` process does gracefully exit when a `SIGTERM` is received. I'm going to do some testing to be sure...

zorn-v:
But `docker-compose` waits 10secs and then send `SIGKILL` for container with horizon. And exit code is 127 in `docker-compose ps` then.
It may be workarounded by `stop_signal: SIGINT` in `docker-compose.yaml`

zorn-v:
Check my implementation of supervisor for symfony messenger

https://github.com/zorn-v/messenger-supervisor-bundle/blob/master/Command/SupervisorCommand.php

djtarazona:
@zorn-v Hmm, my testing shows that Horizon gracefully exists with Docker Compose. I tested with `docker-compose restart` and `docker-compose stop` - both scenarios yielded a 0 exit code. Perhaps you need to allow more time for all the Horizon workers and supervisors to shutdown? Try `docker-compose stop -t 30` to give them 30 seconds. You should give at least as much time as the longest job timeout - that's how long Horizon gives each worker to shutdown.

If you edit Horizon's `MasterSupervisor` and add a log message at the start of the [`terminate()` method](https://github.com/laravel/horizon/blob/28c9d04d9c444833d6dbbe3c05a02f6f5941aea2/src/MasterSupervisor.php#L157), you should see the message in your logs after `docker-compose stop`, showing that the `SIGTERM` is being handled and the `terminate()` method is being called.

PR #819, opened 2020-04-02
[4.x] Fix dark mode config by checking for null value before assigning static property

bert-w:
This fix should now properly assign the dark mode setting.

Via `SomeServiceProvider.php`:
```
// Set to night.
Horizon::night();

// Set to day.
Horizon::night(false);
```

or via `config/horizon.php`:
```
// Set to night.
'theme' => 'dark',

// Set to day (default).
'theme' => 'light',
```` 

The `Horizon::night()` function will take precedence over defining the theme in `config/horizon.php`.

### DIFF
--- a/config/horizon.php
+++ b/config/horizon.php
@@ -6,6 +6,21 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Theme
+    |--------------------------------------------------------------------------
+    |
+    | Here you can define which theme Horizon should load. Horizon currently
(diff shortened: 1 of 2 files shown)


taylorotwell:
Let's just leave it as is for now.

bert-w:
I guess thats fair considering `laravel/telescope` also uses a `Telescope::night()` function (and probably other packages). However I would've liked both to be in the config file.

Issue #823, opened 2020-04-13
Problem with cakephp/chronos


sojeda:
- Horizon Version: ^4.0
- Laravel Version: 7.5.2
- PHP Version: 7.4
- Redis Driver & Version: predis/phpredis #.#.#
- Database Driver & Version:

### Description:
I use "cakephp/chronos": "^2.0" on my project. When i try to install horizon:

 - laravel/horizon 4.x-dev requires cakephp/chronos ^1.0 -> satisfiable by cakephp/chronos[1.3.0, 1.0.0, 1.0.1, 1.0.2, 1.0.3, 1.0.4, 1.1.0, 1.1.1, 1.1.2, 1.1.3, 1.1.4, 1.2.0, 1.2.1, 1.2.2, 1.2.3, 1.2.4, 1.2.5, 1.2.6, 1.2.7,
1.2.8, 1.x-dev].

### Steps To Reproduce:

composer require cakephp/chronos

composer require laravel/horizon

driesvints:
I'll mark this as an enhancement since nothing is really broken here.

mfn:
Would be awesome if Horizon wouldn't depend on Chronos when Laravel itself already depends on Carbon 😄

mfn:
Wooo! https://github.com/laravel/horizon/pull/826

Octet83:
I'm trying to install Horizon but I've an issue related to cakephp/chronos. I do not use it anywhere but when I try to install Horizon with composer I get the following error : 

> Problem 1
>     - Conclusion: don't install laravel/horizon v4.3.0
>     - Conclusion: remove cakephp/chronos 1.3.0
>     - Installation request for laravel/horizon ^4.3 -> satisfiable by laravel/horizon[4.x-dev, v4.3.0].
>     - Conclusion: don't install cakephp/chronos 1.3.0
>     - laravel/horizon 4.x-dev requires cakephp/chronos ^2.0 -> satisfiable by cakephp/chronos[2.0.0, 2.0.1, 2.0.2, 2.0.3].
>     - Can only install one of: cakephp/chronos[2.0.0, 1.3.0].
>     - Can only install one of: cakephp/chronos[2.0.1, 1.3.0].
>     - Can only install one of: cakephp/chronos[2.0.2, 1.3.0].
>     - Can only install one of: cakephp/chronos[2.0.3, 1.3.0].
[...]

PR #824, opened 2020-04-13
[4.x] Chronos 2.0

sojeda:
https://github.com/laravel/horizon/issues/823

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
         "ext-json": "*",
         "ext-pcntl": "*",
         "ext-posix": "*",
-        "cakephp/chronos": "^1.0",
+        "cakephp/chronos": "^2.0",
         "illuminate/contracts": "^7.0",
         "illuminate/queue": "^7.0",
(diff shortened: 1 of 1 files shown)


driesvints:
I guess it's fine to do this since the tests are passing.

composer.json
```diff
@@ -14,7 +14,7 @@
         "ext-json": "*",
         "ext-pcntl": "*",
         "ext-posix": "*",
-        "cakephp/chronos": "^1.0",
+        "cakephp/chronos": "^1.0|^2.0",
```
driesvints, line 17: Why did you change this later on? I think the bump to just `^2.0` is perfectly fine.

sojeda: In case there are projects that use version 1

driesvints: Those projects need to define the chronos dependency in their own composer.json is they depend on it directly. We're changing the dependency for Horizon here.

sojeda: Ok, I leave it like it was before

PR #826, opened 2020-04-13
[5.x] Switch from Chronos to Carbon

browner12:
The original motivation to use Chronos over Carbon was for its immutability.  Now that Carbon v2 offers immutability, we should switch back to Carbon to keep package dependencies consistent across the Laravel Ecosystem.

`CarbonImmutable` is a drop in replacement for `Chronos`, as all method names are the same, and all the tests pass.


driesvints:
Thanks @browner12!

PR #829, opened 2020-04-21
[4.x] Update DashboardStatsController.php

CharbelJ:
- Get the count of pending jobs


taylorotwell:
I am closing this pull request because it lacks sufficient explanation, tests, or both. It is difficult for us to merge pull requests without these things because the change may introduce breaking changes to the framework.

Thanks!

CharbelJ:
Thank you Taylor for your feedback. I will recreate this pr with a more thorough explanation and how it may benefit the user.
The reason behind it is that I was updating the Lumen horizon package to support Lumen 7 and found out that metrics are now split. Previously completed/pending jobs were under one stat but now they are split. Feel free to check the first Lumen horizon 4.0 that is fully compatible with Lumen 7. 😊 https://github.com/carsguide/lumen-horizon

PR #838, opened 2020-05-06
show attempts count for completed jobs

smknstd:
Attemps count is displayed on failed jobs list. But when a job finally succeed after retry, it could also be great to get information about retries.


taylorotwell:
No plans to add this right now.

smknstd:
I think it was displayed before "recent jobs" list was split

attempts-success

PR #839, opened 2020-05-18
Trigger MasterSupervisorDeployed event

styxit:
I wanted to listen for an event whenever Horizon is (re)started. I found the `MasterSupervisorDeployed` event exists but is never triggered.

I'm assuming this was the right place to trigger the event?


taylorotwell:
Seems OK.

PR #840, opened 2020-05-19
Add support for setting workers name

themsaid:
```
'supervisor-2' => [
     //...
    'workersName' => 'myworker',
    //...
],
```


src/Console/SupervisorCommand.php
```diff
@@ -17,6 +17,8 @@ class SupervisorCommand extends Command
     protected $signature = 'horizon:supervisor
                             {name : The name of supervisor}
                             {connection : The name of the connection to work}
+                            {--workers-name=default : The name of the workers}
+                            {--balance= : The balancing strategy the supervisor should apply}
```
mfn, line 21: The `balance=` line is now duplicated, probably not intended to be

PR #841, opened 2020-05-22
[5.x] Make sure you are able to override the full job payload

morloderex:
### Motivation
Currectly when you use `Queue::createPayloadUsing` and you try to override or merge existing tags the existing tags was not part of the job payload.

### The Issue
The issue/bug is simple it's because the JobPayload class gets instanticated after any hooks has had a chance to run.

More specifically the methods that generate the job payload e.g `createObjectPayload` and `createStringPayload` also calls any payload hooks. 

But the methods `later` and `pushRaw` both overrides the the payload right before the attempt to push the given job to the queue. But at this point the hooks have already ran. 

### Solution
The proposed solution to this problem is to override `createObjectPayload` and `createStringPayload` respectively and add the missing keys to the payload array there.

This should give the developer the missing context in the hook. 

### Future changes
I had no clever way of reducing the code repitition sense calling `createObjectPayload` and `createStringPayload` on the parent would trigger the hooks to many times. All i wanted is to grap the "raw" payload before any hooks have ran. So maybe we could add a `createObjectPayloadRaw` and `createStringPayloadRaw` that can take overrides on the parent?

### Sources/ Possible closing issues
Closes: https://github.com/laravel/framework/pull/29643, https://github.com/laravel/framework/issues/29665

> We pass you the existing payload, right? Can't you just merge them there?

_Originally posted by @taylorotwell in https://github.com/laravel/framework/pull/29643#issuecomment-522739489_

### DIFF
--- a/src/RedisQueue.php
+++ b/src/RedisQueue.php
@@ -130,7 +130,7 @@ class RedisQueue extends BaseQueue
             'data' => $data,
             'tags' => $payload['tags'],
             'type' => $payload['type'],
-            'pushedAt' => $payload['pushedAt']
+            'pushedAt' => $payload['pushedAt'],
         ]);
     }
 
@@ -153,7 +153,7 @@ class RedisQueue extends BaseQueue
             ],
             'tags' => $jobPayload->tags(),
             'type' => $jobPayload['type'],
-            'pushedAt' => $jobPayload['pushedAt']
+            'pushedAt' => $jobPayload['pushedAt'],
         ]);
 
         return array_merge($payload, [
--- a/src/RedisQueue.php
+++ b/src/RedisQueue.php
@@ -14,13 +14,6 @@ use Laravel\Horizon\Events\JobsMigrated;
 class RedisQueue extends BaseQueue
 {
     /**
-     * The job that last pushed to queue via the "push" method.
-     *
-     * @var object|string
-     */
-    protected $lastPushed;
-
-    /**
      * Get the number of queue jobs that are ready to process.
      *
      * @param  string|null  $queue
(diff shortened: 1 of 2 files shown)


driesvints:
We can't do all of these changes in the current stable release. Will probably need to go to master.

morloderex:
@driesvints Alright I'll rebase against master, then.

morloderex:
So apperently i cannot change the base branch GitHub gives me a 502 error. But the rebasing should be done.

themsaid:
I don't seem to understand the issue. Can you put it in code form? What code do I need to run to see the issue?

morloderex:
@themsaid there should be an illustration of the issue in the or mentioned see https://github.com/laravel/framework/pull/29643#issuecomment-522742476

themsaid:
@morloderex I reviewed that PR and didn't understand the issue yet. So can you please explain the issue in the context of Horizon? What's wrong with tags and payload?

morloderex:
@themsaid I can try to explain the usecase for this instead of the problem. 

Let's imagine that the developer has some global state which is dynamic generated and changes with each application startup when dispatching any given job class.

This state is set using `Queue::createPayloadUsing` to avoid repitition. For the explaination let's say it's some arbitary uuid. 

```php
Queue::createPayloadUsing(function ($connection, $queue, $payload) {
  return array_merge($payload, [
    'uuid' => 'uuid-here'
  ]);
});
```

The developer is so happy because the state is now populated correctly. Then the developer installs Horizon and finds the documentation on [tags](https://laravel.com/docs/7.x/horizon#tags). 

Now the developer decides it would be a great idea to include this uuid in the tags for the jobs as somekind global identifer. 

Now here is this is the issue:

```php
$uuid = 'string'

Queue::createPayloadUsing(function ($connection, $queue, $payload) {
   // assuming only horizon has tags
   if(!empty($payload['tags'])) {
       $payload = array_merge($payload['tags'], $uuid);
       $payload['uuid'] = $uuid;
   }

  return $payload 
});
```

You cannot currently do something along the lines of this because the `tags` key currently is not populated because of the bug explained above. 

I opened this PR in horizon sense the framework itself does not have the concept of `tags`, so the framework itself doesn't have a `tags` key and it never would.

themsaid:
It was never documented that you can add tags manually except from inside a tags() method inside your job class.

I've opened a PR to allow merging tags from payload into Horizon's tags https://github.com/laravel/horizon/pull/842

PR #842, opened 2020-05-26
Merge tags from payload

themsaid:
This PR allows merging existing tags from the raw payload to the payload created by Horizon. Current behaviour in Horizon 4.x is to use the existing raw tags an override Horizon's.


driesvints:
This can go to master I believe?

PR #853, opened 2020-06-18
[4.x] Autoscaling can now be configured to scale the way you want it to

Kinomej:
For medium / large queues the auto scaling feature creates new processes too slow imho.
It works good if you don't need to have your queue worked through as fast as possible.
But if you are processing something a user waits on the frontend it may take awhile.
You may want to disable auto-scaling or use min-processes to always use minimum 100 processes.

But thats a bit resource heavy on some systems thats why i made this pull request. :)

The Autoscaling may now be configured the way the user wishes.

Say you got 1000 Jobs in your queue and you got 20 processes running.
Depending on your configuration you could say: create as many processes as possible but never more than 50 at a time.

Or you could say: i want to start 25% (250) of the processes each iteration but also never more than 100 at a time.

Also the configuration defaults to: never start more than 1 at a time. 
If your upgrading: the default will always be in place so it's fully backward compatible and you don't need to change a thing in your code.

Also i wrote tests to confirm the 'percentage' and 'limit per iteration' are working.

### DIFF
--- a/config/horizon.php
+++ b/config/horizon.php
@@ -153,6 +153,20 @@ return [
 
     'memory_limit' => 64,
 
+    /**
+    |--------------------------------------------------------------------------
+    | Queue Worker Scale Configuration
+    |--------------------------------------------------------------------------
+    |
+    | Here you may define the queue worker scale settings used by your application
+    | in all environments. These setting will let you limit the number of processes scaled at a time
+    | and speed factor on how many processes will start at a time (Percentage of the total number jobs).
+     */
+    'scale' => [
+        'limit' => 1,
+        'factor' => 20,
+    ],
+
     /*
     |--------------------------------------------------------------------------
     | Queue Worker Configuration
--- a/src/AutoScaler.php
+++ b/src/AutoScaler.php
@@ -130,21 +130,68 @@ class AutoScaler
 
         if (ceil($workers) > $poolProcesses &&
             $this->wouldNotExceedMaxProcesses($supervisor)) {
-            $pool->scale($poolProcesses + 1);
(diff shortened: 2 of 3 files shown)


taylorotwell:
Can you explain more what `factor` and `limit` mean? I find it a bit confusing.

Kinomej:
@taylorotwell 

In Short: `factor` is a factor how many processes of `needed` processes will start at a time.
`limit` is the max. number of processes which will start at a time.

Longer explanation:
`limit` will limit the number processes which will be started on each iteration of scale.
Let's say you want to start max 500 Processes. But you want to make sure they don't all start at the same time. If you have a limit of 50, it will start up to max 50 processes each time until either your max. number of processes is reached or the queue is empty in which case the processes will be terminated by max. 50 processes per iteration.

The `factor` is a bit like the `limit`. 
It limits the processes started to a percentage of the needed workers.
Example: You got 100 Jobs in your queue and you got 0 processes running and you want to start more processes.

Sure you may just use the `limit` to limit it to 50 but that may be your hard-cap and you don't need to have 50 started at a time. 25 per Iteration would be enough (not as resource heavy at a time). So you have that `factor` of 25. Now it would be starting 25% of the 100 possible processes.

So it's okay for that scenario to use the factor but if you got 10000 jobs you may want to reach your cap to work through your queue as fast as possible. (25% for 10000 jobs -> 2500 but capped at max 50)

It seems more scalable for my purposes so i think that makes sense. 🗡️

I hope that will clarify my intentions and what `factor` and `limit` mean. 
If not i will try to explain them again. :)

taylorotwell:

Issue #604, opened 2019-05-31
Can't retry failed jobs


curious-jim:
- Horizon Version: 3.2.1
- Laravel Version: 5.8.19
- PHP Version: 7.3.5
- Redis Driver & Version: predis/phpredis 4.0.9 & 1.1.1
- Database Driver & Version: Ver 15.1 Distrib 10.3.15-MariaDB

### Description:

```
$ php artisan queue:retry all

In Tags.php line 107:
  Class  does not exist
```

trace:

```
local.ERROR: Class  does not exist {"exception":"[object] (ReflectionException(code: -1): Class  does not exist at /home/dev/app/vendor/laravel/horizon/src/Tags.php:107)
[stacktrace]
#0 /home/dev/app/vendor/laravel/horizon/src/Tags.php(107): ReflectionClass->__construct('')
#1 /home/dev/app/vendor/laravel/horizon/src/Tags.php(28): Laravel\\Horizon\\Tags::modelsFor(Array)
#2 /home/dev/app/vendor/laravel/horizon/src/JobPayload.php(132): Laravel\\Horizon\\Tags::for(NULL)
#3 /home/dev/app/vendor/laravel/horizon/src/JobPayload.php(91): Laravel\\Horizon\\JobPayload->determineTags(NULL)
#4 /home/dev/app/vendor/laravel/horizon/src/RedisQueue.php(59): Laravel\\Horizon\\JobPayload->prepare(NULL)
#5 /home/dev/app/vendor/laravel/framework/src/Illuminate/Queue/Console/RetryCommand.php(71): Laravel\\Horizon\\RedisQueue->pushRaw('{\"timeout\":null...', 'production')
#6 /home/dev/app/vendor/laravel/framework/src/Illuminate/Queue/Console/RetryCommand.php(37): Illuminate\\Queue\\Console\\RetryCommand->retryJob(Object(stdClass))
#7 [internal function]: Illuminate\\Queue\\Console\\RetryCommand->handle()
#8 /home/dev/app/vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php(32): call_user_func_array(Array, Array)
#9 /home/dev/app/vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php(90): Illuminate\\Container\\BoundMethod::Illuminate\\Container\\{closure}()
[...]

PR #862, opened 2020-07-13
fix for retry fail job issue

minkbear:
Fix for https://github.com/laravel/horizon/issues/604#issuecomment-656988882


driesvints:
Seems like this breaks the test suite.

PR #865, opened 2020-07-22
[4.x] Display actual delay time instead of relative time

kielabokkie:
The "Delayed for x" tooltip on the pending jobs page is a bit confusing at the moment as it shows the relative time until the job will run again. So if you have a job that's delayed for 10 minutes and 3 minutes have passed it will say "Delayed for 7 minutes". This PR will make sure it always show the exact timeframe the job was delayed for, so it shows "Delayed for 10 minutes" no matter how much time has passed.

I thought about adding an extra column that shows the time the job will run again but thought I first submit this PR to see if the current behaviour is expected or not.


driesvints:
Please revert the asset compiling. We'll do this ourselves. See https://laravel.com/docs/7.x/contributions#compiled-assets

kielabokkie:
@driesvints Sorry, missed that one. All done now.

taylorotwell:
I'm not I particularly like this change. It's nice to see how much longer it is delayed for. Maybe we should just update the wording.

Issue #438, opened 2018-11-09
Allow default for config->horizon.environment


kohlerdominik:
I think it would make sense to have possibility of making a "default"-config for the queue worker configuration.

It could be as easy as:

`src/Console/TimoutCommand.php:38`

from
```php
config('horizon.environments.'.$this->argument('environment'), [])
```

to
```php
config('horizon.environments.'.$this->argument('environment'),
    config('horizon.environments.default', []))
```

It would change from directly returning an empty array if the environment doesn't exist to first check if default exists, and then returning the empty array.

driesvints:
I don't think this is necessary as Horizon at least needs an environment to run. How would an empty default help here?

kohlerdominik:
Well i don't meant to discuss about empty default (which, just to say, is current implementation).

It's about possibility to create the "default" environment in the config file, working for example for production, stage and integration. I think, often these three environments would be exactly the same, so why do i need to copy/paste it 3 times instead of using the "default"-environment and only if local, its loading the local-configuration.

driesvints:
Ah I see what you mean. Yeah, could be useful perhaps. Feel free to send in a PR to master.

djtarazona:
@kohlerdominik I'm thinking of giving this a try:

```php
// in config/horizon.php

'environments' => [
    env('APP_ENV') => [
        'supervisor-1' => [
            'connection' => 'redis',
            'queue' => ['default'],
            'balance' => 'simple',
            'processes' => 10,
            'tries' => 1,
        ],
    ],
],
```

hackel:
Another reason it would be very useful to be able to set environment defaults in the config is so we can set the default values once, and only customize what we need per environment.  All of my environments are identical aside from maxProcesses.

It would be very nice if I could quickly spin up a new testing environment without having to commit a temporary change to my config.

driesvints:
I'm attempting a PR for this: https://github.com/laravel/horizon/pull/869

Feedback welcome on it.

driesvints:
This got merged. Here's the new config:

```php
    /*
    |--------------------------------------------------------------------------
    | Queue Worker Configuration
    |--------------------------------------------------------------------------
    |
    | Here you may define the queue worker settings used by your application
    | in all environments. These supervisors and settings handle all your
    | queued jobs and will be provisioned by Horizon during deployment.
    |
    */

    'defaults' => [
        'supervisor-1' => [
            'connection' => 'redis',
            'queue' => ['default'],
[...]

PR #869, opened 2020-08-06
[5.x] Implement default environment

driesvints:
This PR implements a default environment for Horizon. Any other environment will re-use this default environment's settings which can be overridden. Supported environments still need to be defined but this helps with the maintenance burden of having to maintain duplicate settings for multiple environments.

I've updated the config file which immediately shows how this can be useful.

Sent to master because it's obviously breaking for anyone who was relying on the current formatting of the config.

Closes https://github.com/laravel/horizon/issues/438

### DIFF
--- a/config/horizon.php
+++ b/config/horizon.php
@@ -165,23 +165,25 @@ return [
     */
 
     'environments' => [
-        'production' => [
+        'default' => [
             'supervisor-1' => [
                 'connection' => 'redis',
                 'queue' => ['default'],
                 'balance' => 'simple',
-                'processes' => 10,
+                'processes' => 1,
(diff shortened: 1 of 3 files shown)


driesvints:
Additional options for this PR are:

- Move the defaults out of the environments array
- Don't require environments to be defined in the array and use the `default` one when one cannot be found

hivokas:
Thanks @driesvints! That's a very useful feature.

setkyar:
Hi @driesvints , I added the documentation for this PR. Since it's confusing for people who use multiple supervisors and the default queues are on there. Please help check and help me add comments if I can improve on the documentation.

Thanks

Issue #658, opened 2019-09-03
Autoscaler is too slow (or run too seldom)


lechbaczynski:
- Horizon Version: 3.3.2
- Laravel Version: 5.8.34
- PHP Version: 7.2.19
- Redis Driver & Version: predis 1.1.1
- Database Driver & Version: mysql

### Description:

Supervisor with 6 queues and 1800 processes (3 other supervisors running too, but with not much workers and jobs). 
Starts with 300 processes, then the numbers go down.

There is a lot of jobs in one queue, yet the number of workers rises very slowly. It may take over half an hour to an hour to rise. Although all other queues are empty.
(in the image below there were some jobs in other queues, but even if other queues are empty, the number of workers rises slowly - here it rose to 1300, but it took long time)

![image](https://user-images.githubusercontent.com/1127623/64159950-5c649a00-ce3b-11e9-8e98-a33376ac27f8.png)

### Steps To Reproduce:

Make a supervisor like 

```
'supervisor-1' => [
                'connection' => 'redis',
                'queue' => [
                    'one', 'two', 'three',
                    'four', 'five', 'six'
                ],
                'balance' => 'auto',
                'processes' => 1800, // even multiply of number of queues
                'min_processes' => 2,
                'tries' => 5,
                'timeout' => 550,
                'memory' => 512,
            ],
```

The number of workers will decrease from 300.
Add 100.000 jobs in `one` queue, and make each job run for random(5,200) seconds.
The queue workers number will rise slowly.

driesvints:
Not sure what's going on here. Maybe @themsaid can help?

lechbaczynski:
Video with one queue, number of workers and tasks. Other queues were rather empty at that time (animated GIF) 
![workersrisingslowly](https://user-images.githubusercontent.com/1127623/64232830-03583d00-cef3-11e9-87a6-c6ead9622ba2.gif)

lechbaczynski:
Another strange thing: 5 queues, 2000 processes, min_processes 50.

```
            'processes' => 2000, // even multiply of number of queues
            'min_processes' => 50,
```

3 queues basically idle, two have thousands of jobs.
The number of processes in the bigger queue grow, but in the smaller shrinks - although it is still not using those 2000 allowed processes. Shouldn't it rise?

![11 41](https://user-images.githubusercontent.com/1127623/64245024-6acdb700-cf0a-11e9-8ddf-45c0564a2ce0.png)

[...]

PR #874, opened 2020-08-18
[5.x] Make autoscaling rate configurable

hivokas:
This PR makes autobalancing rate configurable for applications that use the `auto` queue balancing strategy.

## Problem

Imagine we have the following supervisor config:

```php
'supervisor-1' => [
    'connection' => 'redis',
    'queue' => ['high', 'low'],
    'balance' => 'auto',
    'minProcesses' => 5,
    'maxProcesses' => 150,
    'tries' => 1,
],
```

Let's say we dispatch 1,000 jobs to the `low` queue for something like an email campaign. Currently, autobalancing will only adjust the number of workers at a maximum rate of ±1 worker every 3 seconds. So, it will take a long time to increase the number of workers from 5 (`minProcesses`) to 150 (`maxProcesses`). There's no way to change this.

## Solution

With this PR we make the autobalancing rate configurable:

```php
/*
|--------------------------------------------------------------------------
| Autoscaling
|--------------------------------------------------------------------------
|
| Here you may define the autoscaling settings used by your application.
|
*/

'autoscaling' => [
    'cooldown' => 3,
    'max_shift' => 1,
],
```

`cooldown` - the number of seconds to wait in between auto-scaling attempts
`max_shift` - the maximum number of processes to increase or decrease per one scaling

## Breaking changes

This PR doesn't introduce any breaking changes since the default values (for `cooldown` and `max_shift`) are consistent with the current values. Even if someone upgrades without adding these config values, it will still use those defaults.

However, I decided that it might be better to point this PR to the `5.x` branch.

## Benefit

Let's perform a test by dispatching 750 jobs at once (each job sleeps for 5 seconds) without this change. We can see it takes [135 seconds](https://pastebin.com/raw/LJ9cHLbK) to process these jobs.

Let's set `'cooldown' => 1` and `'max_shift' => 50` and perform the same test one more time. 

This time, it took only [30 seconds](https://pastebin.com/raw/QZqkMC4C), because autoscaling happens much faster.

## Related issues

https://github.com/laravel/horizon/issues/658

### DIFF
--- a/config/horizon.php
+++ b/config/horizon.php
@@ -163,14 +163,7 @@ return [
     */
 
     'autoscaling' => [
-        /*
-         * The cooldown between autoscaling iterations in seconds.
-         */
         'cooldown' => 3,
-
-        /*
-         * The maximum number of processes to increase/decrease per iteration.
-         */
         'max_shift' => 1,
     ],
 
--- a/config/horizon.php
+++ b/config/horizon.php
@@ -155,20 +155,6 @@ return [
 
     /*
     |--------------------------------------------------------------------------
-    | Autoscaling
-    |--------------------------------------------------------------------------
-    |
-    | Here you may define the autoscaling settings used by your application.
-    |
-    */
-
-    'autoscaling' => [
-        'cooldown' => 3,
-        'max_shift' => 1,
-    ],
-
-    /*
-    |--------------------------------------------------------------------------
     | Queue Worker Configuration
     |--------------------------------------------------------------------------
     |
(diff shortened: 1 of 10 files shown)


taylorotwell:
What if you want the cool_down and max_shift to be different for each supervisor configuration? Any reason it's not part of that configuration?

hivokas:
> What if you want the cool_down and max_shift to be different for each supervisor configuration?

Hm, that sounds like a good idea. 

> Any reason it's not part of that configuration?

The reason is I didn't think about that.

hivokas:
Does the following example of config look good for you @taylorotwell?

```php
'supervisor-1' => [
    'connection' => 'redis',
    'queue' => ['high', 'low'],
    'balance' => 'auto',
    'minProcesses' => 5,
    'maxProcesses' => 150,
    'tries' => 1,
    'maxShift' => 20,
    'cooldown' => 1,
],
```

hivokas:
If yes, I'll work on moving `cooldown` and `max_shift` to the supervisor config.

taylorotwell:
Maybe rename `cooldown` to `balanceCooldown`?

hivokas:
@taylorotwell I've updated the PR. `balanceCooldown` and `autoScaleMaxShift` are the part of a supervisor configuration now.

taylorotwell:
Renamed `autoScaleMaxShift` to `balanceMaxShift`.

hivokas:
Thanks for merging!

PR #876, opened 2020-08-27
[5.x] Enable fast termination by default

halaei:
To me this feature is the least commonly used, yet the most commonly needed feature, and perhaps unknown to many.
The fast_termination is a must to have when you are expected to have long running jobs, e.g. jobs requiring several seconds to be done, alongside real-time jobs that need to be handled with the minimum possible delay, for example less than a second.
The default configuration should be considering this a cases, especially because this config doesn't come with terrible side-effects.
Not having this option have the following consequences during the deployment, when we need to terminate a horizon instance and start a new one:
1. If a long running job is being processed, all the real-time jobs are delayed more than the users can tolerate.
2. If you want to be able to safely restart/stop horizon via supervisord, you must always make sure the value of stopwaitsecs of supervisord is safely greater that the maimum expected run-time of your jobs. To me this consequence is not as important as the first one, mainly because it is probably not the way most of the people restart their workers during deployment.

These are the downsides of fast_termination, as far as I remember. I think they are mostly negligible by the majority of applications:
1. During the termination of a horizon instance and the start of the next one, the number of processes will increase to nearly as twice as normal in the worst case. Hence the memory usage can increase temporarily.
2. If there are queues that needed to be handled only by one worker at the time, this feature should not be used unless jobs are secured by other mechanisms like atomic locks.

For more information, see the related issue (#400) and PRs (#402, #406, #426) implementing fast_termination.

Some documentation will also be good to have, especially in case you decide not to accept this PR.


taylorotwell:
I think I will keep the default as is.

halaei:
In that case, some changes in descriptions isn't bad. Maybe a sentence about it enabling zero-downtime for applications that need to have low latency real time jobs.

barryvdh:
I guess ideally the number of processes wouldn't exceed the configuration. So if the old process (eg a  'queue-long' worker) is running, it should start the new Horizon supervisor with the workers, minus the ones still running on the old supervisor, until that one is also finished.

BTW, thanks for this. Didn't realize this option existed. Ran into this a few times when the default queue wasn't restarting because a single longrunning task was processing.

bmckay959:
I'm running into similar issues with long running tasks. Some of which take 10-15 minutes to run in the background.

This is causing my Envoyer setup to not really be zero downtime because my queues are 'frozen' while the last job finishes up.

Has there been any thoughts/progress on this front? It would be really helpful for my use case to do as @barryvdh  mentioned and start the new workers minus the ones still running.

PR #879, opened 2020-08-27
[4.x] niceness is opposite of priority

halaei:
This partially reverts https://github.com/laravel/horizon/commit/8d0034bd6d72450be8cdba8e874656d3e704306d because I think the original description was better. If I increase the nice value I actually decrease the priority and increase the niceness. But the descriptions suggests the exact opposite.


halaei:
I was thinking of adding some small description about all the possible options in config.php, so the developers can understand the meaning and options faster and make sure they don't miss anything. Are you OK with that?

PR #881, opened 2020-09-03
[5.0] Check if parent is still running

halaei:
Currently if horizon process gets terminated by SIGKILL, supervisor processes must be killed by kill command, or they will be provisioning workers for ever. There is no artisan command to terminate orphan supervisors, so it is a little bit of trouble.

If you want to see if you have orphan supervisors in your production, you can use the following command:
```bash
ps  xao pid,ppid,etime,time,comm,args|grep -P "^\\s+(PID|\\d+\\s+[01]\s.*horizon:)"
```

This PR lets the orphan supervisors check for their parent process id at the beginning of their loop, so they can terminate in case their parent, `php artisan horizon`, is no longer running.

### DIFF
--- a/src/SupervisorOptions.php
+++ b/src/SupervisorOptions.php
@@ -54,6 +54,13 @@ class SupervisorOptions
     public $minProcesses = 1;
 
     /**
+     * The parent process identifier.
+     *
+     * @var int
+     */
+    public $parentId = 0;
+
+    /**
      * The process priority.
      *
      * @var int
@@ -158,6 +165,7 @@ class SupervisorOptions
      * @param  int  $nice
      * @param  int  $balanceCooldown
      * @param  int  $balanceMaxShift
+     * @param  int  $parentId
      */
     public function __construct($name,
                                 $connection,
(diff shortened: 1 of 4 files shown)


src/SupervisorOptions.php
```diff
@@ -176,7 +184,8 @@ public function __construct($name,
                                 $force = false,
                                 $nice = 0,
                                 $balanceCooldown = 3,
-                                $balanceMaxShift = 1)
+                                $balanceMaxShift = 1,
+                                $parentId = 0)
```
driesvints, line 188: @halaei why didn't you implement this dependency? Atm it'll always be zero.

halaei: Sorry @driesvints It seems I made a mistake. Setting parentId in constructor also doesn't work because it is still zero in ensureParentIsRunning.

driesvints: @halaei what do you mean? If we fix this by passing the `$parentId` from the constructor to the property it should work, no? Why do you think it would still be zero?

halaei: @driesvints I just dump() the value! It was 0 for sure. I sent a draft PR

PR #742, opened 2020-01-11
[3.x] Add ability to purge queues

parkourben99:
Add the ability to purge individual queues with a confirmation.

Fixes https://github.com/laravel/horizon/issues/328

image

image

PR #890, opened 2020-09-17
[5.x] Add ability to clear queues from dashboard

### LINKED ISSUE #328: Clear all jobs of one queue
Hi,

In beanstalk there is some button "Delete all ready jobs" is there in horizon also a button for that action?

paras-malhotra:
queues
confirm

This PR adds the ability to clear queues (delete all jobs from specific queues) from the Horizon dashboard.
Resubmission of https://github.com/laravel/horizon/pull/742 with fixes to sorted sets and also uses the in-built queue clear of the Laravel framework based on https://github.com/laravel/framework/pull/34330. Fixes #328. 

**Important Notice**
Clearing of queues will not work on Laravel versions below 8.4 (unreleased). So I'm not sure if we want to send to master or 5x? The `illuminate\queue` repository isn't updated with the recent changes either, so we can't bump up that version yet. Also, I haven't compiled assets with this PR. Let me know if you'd like me to update with compilation of assets as well.

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -5,6 +5,7 @@ use Illuminate\Support\Facades\Route;
 Route::prefix('api')->group(function () {
     // Dashboard Routes...
     Route::get('/stats', 'DashboardStatsController@index')->name('horizon.stats.index');
+    Route::post('/clearQueue', 'QueueClearController')->name('horizon.queue-clear.store');
 
     // Workload Routes...
     Route::get('/workload', 'WorkloadController@index')->name('horizon.workload.index');
--- a/src/Repositories/RedisJobRepository.php
+++ b/src/Repositories/RedisJobRepository.php
@@ -689,13 +689,13 @@ class RedisJobRepository implements JobRepository
     }
 
     /**
-      * Delete pending and reserved jobs for a queue.
-      *
-      * @param  string  $queue
-      * @return void
-      */
-     public function purge($queue)
-     {
+     * Delete pending and reserved jobs for a queue.
+     *
+     * @param  string  $queue
+     * @return void
+     */
+    public function purge($queue)
+    {
         $ids = collect(range(0, ceil($this->nextJobId() / 50)))->transform(function ($_, $index) use ($queue) {
             return $this->getRecent($index === 0 ? null : $index * 50)->filter(function ($recent) use ($queue) {
                 return in_array($recent->status, ['pending', 'reserved']) && $recent->queue === $queue;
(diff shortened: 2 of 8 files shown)


taylorotwell:
It would probably be best to check for the existence of `Illuminate\Foundation\Application` and then read the VERSION constant from it to determine whether to show this in the UI.

paras-malhotra:
Sure @taylorotwell, I think an even better way might be to check if the queue implements `ClearableQueue`. I'll update the PR accordingly.

taylorotwell:
Sure, either way.

paras-malhotra:
@taylorotwell, I've added the check in the UI to display only if the `clear` method exists. Haven't committed freshly compiled assets. Let me know if you'd like me to do that and update the PR. Thanks!

src/Repositories/RedisJobRepository.php
```diff
@@ -688,6 +688,27 @@ public function deleteFailed($id)
...
+     * @param  string  $queue
+     * @return void
+     */
+    public function purge($queue)
+    {
+        $ids = collect(range(0, ceil($this->nextJobId() / 50)))->transform(function ($_, $index) use ($queue) {
+            return $this->getRecent($index === 0 ? null : $index * 50)->filter(function ($recent) use ($queue) {
+                return in_array($recent->status, ['pending', 'reserved']) && $recent->queue === $queue;
+            });
+        })->flatten(1)->pluck('id')->all();
```
taylorotwell, line 703: Please explain what this is doing.

routes/web.php
```diff
@@ -5,6 +5,7 @@
 Route::prefix('api')->group(function () {
     // Dashboard Routes...
     Route::get('/stats', 'DashboardStatsController@index')->name('horizon.stats.index');
+    Route::post('/clearQueue', 'QueueClearController')->name('horizon.queue-clear.store');
```
taylorotwell, line 8: Unclear to me why a delete operation would be a POST. Shouldn't it be something like DELETE /queues/{queue-name} or something?

paras-malhotra: I thought of clearing queues as more of an action rather than deleting a resource. If you'd like I can change it to a delete operation.

src/Repositories/RedisJobRepository.php
paras-malhotra: Here we're collecting the IDs of the pending or reserved jobs on the specified queue. To do so, we get all the recent jobs (50 at a time based on the in-built RedisJobRepository methods) and filter through them to check 1) if they're either pending or reserved and 2) if the queue matches the specified queue. On the filtered set of jobs, we pluck the job IDs and then delete them from the Redis hash and recent/pending jobs sorted sets.

taylorotwell:
Why don't we take this one step at a time. Why don't we have just a CLI command to clear queues in Horizon first.

PR #891, opened 2020-09-19
[5.x] Add ability to see if a failed job is retried and/or a retry

hxnk:
In the current situation, a job that has previously failed and has a successful retry can be identified by the fact that it doesn't show the retry indicator/button on the right side.

If retrying a failed job results in the retry also failing, you'll end up with two failed jobs that can both be retried again: the original failed job, and the failed retry.
Right now it's hard to keep track of what jobs have been retried and/or are (failed) retries, since they all look the same. This PR tries to solve that.

On the index page:
- Failed jobs that have been retried show a "Retried" badge
- Failed jobs that are retries show the (truncated) ID of the retried job that links to the detail page of that job

Horizon_-_Failed_Jobs_Index

On the detail page:
- The number of retries are shown
- If the failed job is a retry, the job ID of the retried job is shown (if the job is not a retry, this will not be shown)
- The Attempts count was missing here, so I added that for good measure

Horizon_-_Failed_Jobs_Details

### DIFF
--- a/resources/js/screens/failedJobs/index.vue
+++ b/resources/js/screens/failedJobs/index.vue
@@ -137,6 +137,22 @@
 
 
             /**
+             * Determine if the given job was retried.
+             */
+            wasRetried(job) {
+                return job.retried_by && job.retried_by.length;
+            },
+
+
+            /**
+             * Determine if the given job is a retry.
+             */
+            isRetry(job) {
+                return job.payload.retry_of;
+            },
+
+
+            /**
              * Refresh the jobs every period of time.
              */
             refreshJobsPeriodically() {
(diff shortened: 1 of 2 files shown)


taylorotwell:
If I'm viewing this page:

Screen Shot 2020-09-21 at 8 28 07 AM

And click on the retry attempt. I'm taken to the detail page. But clicking on the "Retry of ID" ID doesn't update the page... the ID stays the same...

Screen Shot 2020-09-21 at 8 28 11 AM

hxnk:
Fixed! :)

taylorotwell:
Merged.

PR #893, opened 2020-09-22
[5.x] Add ability to delete a failed job

hxnk:
This PR expands upon #891 by further improving managing failed jobs.

It would be nice to be able to delete failed jobs no longer needing attention, which makes it easier to find the ones that do. I've renamed the Retry column on the Index page to Actions, and added a button to delete the job. The same button is added on the detail page as well.

I also fixed the `deleteFailed` method on the RedisJobRepository; it now actually checks if the job is failed before deleting it. Tests are added.

Horizon_-_Failed_Jobs

Screenshot_22-09-2020_21_45

### DIFF
--- a/resources/js/screens/failedJobs/index.vue
+++ b/resources/js/screens/failedJobs/index.vue
@@ -121,6 +121,19 @@
 
 
             /**
+             * Remove the given failed job.
+             */
+            remove(id) {
+                this.$http.delete(Horizon.basePath + '/api/jobs/failed/' + id)
+                    .then((response) => {
+                        this.jobs = _.reject(this.jobs, job => job.id == id)
+                    }).catch(error => {
+                        //
+                    });
+            },
+
+
+            /**
              * Determine if the given job is currently retrying.
              */
             isRetrying(id) {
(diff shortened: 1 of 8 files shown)


paras-malhotra:
Looks like a useful feature. I think we should also delete the failed job from the `queue.failer` / failed job provider similar to the `queue:forget` command.

hxnk:
I see your point. However, I can imagine people using the failed job provider for different purposes.

Also, failed jobs in horizon get trimmed (by default) after 7 days, without touching the failed job provider. 
Deleting a job manually from the horizon UI could be seen as "Don't wait for auto-trim, remove now.", so I think I'm leaning towards having it behave the same.

hxnk:
Thinking about it some more; I think I like it as optional behaviour

We could:

1. Add it as a configurable option in the horizon config
2. Fire an event `FailedJobDeleted` and let the user handle it in whatever way they want

I prefer the second option

paras-malhotra:
Most users would be unaware of how Horizon works behind the scenes. They're probably thinking that the failed jobs shown is anyway from the database. Nowhere in the docs, have we mentioned that recent/failed/pending jobs shown on the Horizon dashboard are stored separately as sorted sets in Redis. That's why I was thinking to just make it simple and deleting a failed job should delete from both Redis and DB.

taylorotwell:
Yeah it should probably delete from both and I don't want another configuration option.

taylorotwell:
Also, this gets into other features like people are going to immediately ask for a modal confirmation window before deleting the job, etc. So I think I will just hold off on this for now.

PR #896, opened 2020-09-24
[5.x] Add horizon:forget command to delete a failed job

paras-malhotra:
Currently, there's no way to delete a failed job from both Redis and the database. The `queue:forget` command in the framework simply deletes the job from the database but the Horizon dashboard still shows the job as failed.

This PR provides a clean way for users using Horizon to delete failed jobs.

### DIFF
--- a/src/Console/ForgetFailedCommand.php
+++ b/src/Console/ForgetFailedCommand.php
@@ -1,0 +1,39 @@
+<?php
+
+namespace Laravel\Horizon\Console;
+
+use Illuminate\Console\Command;
+use Laravel\Horizon\Contracts\JobRepository;
+
+class ForgetFailedCommand extends Command
+{
+    /**
+     * The console command signature.
+     *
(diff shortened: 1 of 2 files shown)


driesvints:
Wondering if we can make this a bit more generic as a `ForgetCommand` that can delete any job but has flags like `--failed` to specifically delete failed jobs. Does that even makes sense?

paras-malhotra:
@driesvints I wanted to make it mirror the `queue:forget` like we have `horizon:clear` and `queue:clear`.

Also, there's no way to delete a pending job using the Framework. The `delete` function on the Job class just deletes reserved jobs and doesn't work for pending jobs. I guess this hasn't been added to the framework as it would be very inefficient for queues like Redis that need a full scan operation after reconstructing the job json payload.

driesvints:
@paras-malhotra gotcha. Makes sense 👍

PR #900, opened 2020-10-01
[5.x] Add secs to y-axis ticks' for clarity

mazedlx:
Since the line chart's y-axis shows ticks without units, this could help clarify that it's in seconds.


driesvints:
Can you maybe show a screenshot on how this would look?

mazedlx:
Bildschirmfoto 2020-10-01 um 11 15 58

It should only be visible in the runtime charts - the other charts are "fine" ;-)

Issue #906, opened 2020-10-09
Delayed job shows always in horizon the current timestamp


stefanbauer:
- Horizon Version: 4.3.5
- Laravel Version: 7.26.1
- PHP Version: 7.4.10
- Redis Driver & Version: predis 1.1.4
- Database Driver & Version: MySQL 5.7

### Description:

If you go to your horizon dashboard, click on the pending jobs item and select a pending job, you'll see an overview like this:

![image](https://user-images.githubusercontent.com/3192662/95564605-54ecc880-0a1f-11eb-84f6-61d9a080fa1b.png)

I'd expect that within the "Delayed Until" field there is the timestamp when the job should be executed. But it shows always the current timestamp. Refresh the page and you can see that it refreshes always to the current timestamp.

Here's a better screenshot. This job was already executed and it's already completed, but still the `delay until` refreshing (in the future which doesn't make sense though)

![image](https://user-images.githubusercontent.com/3192662/95565694-db55da00-0a20-11eb-92fc-6de8087a1fcf.png)

### Steps To Reproduce:

- Queue a delayed job with any delay (e.g. 1 hour)
- Go to your horizon dashboard and check the delayed jobs
- You won't see the correct "delayed until" timestamp (queuing + delay), but you see the current timestamp.

driesvints:
Hey @stefanbauer can you please try this in Horizon 5? We're not maintaining 4.x anymore. Feel free to report back when you've managed to reproduce it in 5.x

PR #907, opened 2020-10-09
[5.x] Fix delayed until on pending jobs screen

paras-malhotra:
Fixes https://github.com/laravel/horizon/issues/906

I've confirmed this is a bug in 5.x as well. The issue is that [PendingDispatch::delay](https://github.com/laravel/framework/blob/86fc8fb07f05d8a8d5f88f452d7beada732909da/src/Illuminate/Foundation/Bus/PendingDispatch.php#L89) accepts an `int` or `DateTime`. The current Horizon code assumes that it's always a date. This PR fixes that. I've checked it on my PC.

Steps to reproduce:
1. Execute `TestJob::dispatch(...)->delay(60*60*24);`
2. Go to the pending jobs screen. On the index screen, it shows "delayed for a few seconds" (as tooltip) and on the job detail screen, it shows Delayed Until as the current time.

### DIFF
--- a/resources/js/screens/recentJobs/job-row.vue
+++ b/resources/js/screens/recentJobs/job-row.vue
@@ -58,8 +58,11 @@
             },
 
             delayed() {
-                if (this.unserialized && this.unserialized.delay) {
+                if (this.unserialized && this.unserialized.delay && this.unserialized.delay.date) {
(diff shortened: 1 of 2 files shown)


stefanbauer:
Thanks for fixing it @paras-malhotra 👍

PR #910, opened 2020-10-12
Update CompletedJobsController.php

CharbelJ:
This controller gets completed jobs, we should get the total completed jobs and not recent jobs


taylorotwell:
@themsaid ?

PR #914, opened 2020-10-23
[5.x] Add commands to pause and continue supervisors

paras-malhotra:
Currently, it is possible to pause and continue the master supervisor. But there is no command readily available to pause and continue specific supervisors. This could be a common use case in production. 

For instance, say we want to pause a supervisor that runs a dedicated queue for unstable APIs. If the API is currently unavailable, pausing that specific supervisor really helps so that the jobs are queued up and can be continued (without exhausting their `tries`) once the API is up and running.

Screenshot 2020-10-23 at 3 48 15 PM

EDIT: I've added separate commands instead of modifying existing `PauseCommand` and `ContinueCommand` (used for master supervisor) so that there's no breaking change and this PR can be included for 5x.

### DIFF
--- a/src/Console/ContinueSupervisorCommand.php
+++ b/src/Console/ContinueSupervisorCommand.php
@@ -34,7 +34,7 @@ class ContinueSupervisorCommand extends Command
     {
         $processId = optional(collect($supervisors->all())->first(function ($supervisor) {
             return Str::startsWith($supervisor->name, MasterSupervisor::basename())
-                    && Str::endsWith($supervisor->name, ':'.$this->argument('name'));
+                    && Str::endsWith($supervisor->name, $this->argument('name'));
         }))->pid;
 
         if (is_null($processId)) {
--- a/src/Console/ContinueSupervisorCommand.php
+++ b/src/Console/ContinueSupervisorCommand.php
@@ -1,0 +1,52 @@
+<?php
+
+namespace Laravel\Horizon\Console;
+
+use Illuminate\Console\Command;
+use Illuminate\Support\Str;
+use Laravel\Horizon\Contracts\SupervisorRepository;
+use Laravel\Horizon\MasterSupervisor;
+
+class ContinueSupervisorCommand extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'horizon:continue-supervisor
+                            {name : The name of the supervisor to resume}';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Instruct the supervisor to continue processing jobs';
+
+    /**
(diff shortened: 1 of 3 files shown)


paras-malhotra:
I've slightly modified the PR so that both full and short names of supervisor work. E.g. both of the below commands work:

`php artisan horizon:pause-supervisor paras-macbook-prolocal-cEEq:supervisor-1` and
`php artisan horizon:continue-supervisor supervisor-1`

The second one is the same name as appears in the `horizon.php` config file and the first one is the full name of the supervisor as appears in the `horizon:supervisors` command.

src/Console/ContinueSupervisorCommand.php
```diff
@@ -0,0 +1,52 @@
...
+        if (is_null($processId)) {
+            $this->error('Failed to find a supervisor with this name');
+
+            return 1;
+        }
+
+        $this->info("Sending CONT Signal To Process: {$processId}");
+
+        if (! posix_kill($processId, SIGCONT)) {
+            $this->error("Failed to kill process: {$processId} (".posix_strerror(posix_get_last_error()).')');
```
mfn, line 49: > Failed to kill process

Just a small feedback that this is not what's happening from the user perspective: we're not trying to send SIGKILL or SIGTERM here, maybe it should be rephrased to "Cant send signal SIGCONT to…" or so?

src/Console/PauseSupervisorCommand.php
```diff
@@ -0,0 +1,52 @@
...
+        if (is_null($processId)) {
+            $this->error('Failed to find a supervisor with this name');
+
+            return 1;
+        }
+
+        $this->info("Sending USR2 Signal To Process: {$processId}");
+
+        if (! posix_kill($processId, SIGUSR2)) {
+            $this->error("Failed to kill process: {$processId} (".posix_strerror(posix_get_last_error()).')');
```
mfn, line 49: > Failed to kill process

Basically the same feedback, something like "Cant send signal SIGUSR2 to…" 🤔

mfn (commented):
Very nice, just a suggestion regarding the wording!

driesvints:
@mfn can you send in a PR for those?

paras-malhotra:
@mfn, this resembles the pause command for the master supervisor. It stems from a funny unix history, where the unix `kill` command is used to send any signal, not just `SIGKILL`. If you wish to send a PR to make it obvious, you can also change the `PauseCommand` and `ContinueCommand` accordingly.

driesvints:
@paras-malhotra can you maybe PR these new commands to the docs?

paras-malhotra:
@driesvints sent a PR to the docs: https://github.com/laravel/docs/pull/6544

driesvints:
@paras-malhotra thanks! :)

PR #915, opened 2020-10-26
[5.x] Added constants for domain and path variables to config

siarheipashkevich:
Currently there's no way to specify the domain and path of Horizon for different environments (for local env I don't wanna use domain but for production and stage I should use domain `horizon.stage.site.com` for `stage` or `horizon.site.com` for production). This PR added ability to set domain and path in env file.


taylorotwell:
Feel free to add these as environment variables in your own application.

PR #918, opened 2020-10-30
[5.x] Improve wording of error message

mfn:
See discussion outcome in https://github.com/laravel/horizon/pull/914#issuecomment-717254110

This also better aligns with the `info` message right above in both places. I neither cases are we "killing" the process, we're just sending a signal.


mfn:
Thank you, @taylorotwell !

PR #929, opened 2020-11-20
[5.x] Ability to see which masters are paused and only show paused if everything is paused

### LINKED ISSUE #927: Show how much "servers" are in paused or running
If you have multiple servers running horizon, you only "pause" one server, the one which you enter the command. It would be helpfull if you can see which "host" is paused. Better should be if the status above show the count of running "hosts" and only shows "paused" when all of them are paused. On each server there could be a status showing it is running or not.

For example:
You have 4 running servers: status = Active (4/4)
You have 3 running servers: status = Active (3/4)
You have 2 running servers: status = Active (2/4)
You have 1 running servers: status = Active (1/4)
You have 0 running servers: status = Paused

If this something that would be a good thing to add, I would try to add this feature in a PR.

glamorous:
This pull request does two things:

- Show on a every "master" if it's paused or running
- Not showing as "paused" if not all master supervisors are paused but show how many are pushed instead.

A couple of images to show what has been changed and what has been kept the same.

status_workers
inactive-not-changed
active_unchanged
paused_unchanged
active_changed

_Little sidenote: I wasn't sure if I should update the javascript files by running npm run production. If necessary, I can make changes._

Closes #927

### DIFF
--- a/tests/Controller/DashboardStatsControllerTest.php
+++ b/tests/Controller/DashboardStatsControllerTest.php
@@ -74,7 +74,7 @@ class DashboardStatsControllerTest extends AbstractControllerTest
         $masters = Mockery::mock(MasterSupervisorRepository::class);
         $masters->shouldReceive('all')->andReturn([
             (object) [
-                'status' => 'running',
+                'status' => 'paused',
             ],
             (object) [
                 'status' => 'paused',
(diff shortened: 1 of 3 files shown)


tests/Controller/DashboardStatsControllerTest.php
```diff
@@ -74,7 +74,7 @@ public function test_paused_status_is_reflected_if_all_master_supervisors_are_pa
         $masters = Mockery::mock(MasterSupervisorRepository::class);
         $masters->shouldReceive('all')->andReturn([
             (object) [
-                'status' => 'running',
+                'status' => 'paused',
```
paras-malhotra, line 77: you may want to change the method name to `test_paused_status_is_reflected_if_any_master_supervisor_is_paused`

Issue #935, opened 2020-11-26
Snapshot miss lock time to time


glamorous:
- Horizon Version: 5.x
- Laravel Version: 8.x
- PHP Version: 7.4
- Redis Driver & Version: predis/phpredis #.#.#

### Description:
Because the cache-lock for metrics is exact 5minutes or 300s you can have cases when cron is running slow/fast, he can't obtain the lock and misses the snapshot for that timing.

### Steps To Reproduce:
Having  fast or slow environment.

Problem lays here:

```
/**
     * Execute the console command.
     *
     * @param  \Laravel\Horizon\Lock  $lock
     * @param  \Laravel\Horizon\Contracts\MetricsRepository  $metrics
     * @return void
     */
    public function handle(Lock $lock, MetricsRepository $metrics)
    {
        if ($lock->get('metrics:snapshot', 300)) {
            $metrics->snapshot();

            $this->info('Metrics snapshot stored successfully.');
        }
```

Proposal is to have this a couple of seconds lower or make it configurable so people can choose to have shapshots every 2 minutes for example.

driesvints:
Sure, feel free to send in a PR.

PR #936, opened 2020-11-26
[5.x] Metrics snapshot config proposal and fix for race condition

glamorous:
- New option in the config so you will be able to save snapshots on smaller timestamps then 5 minutes.
- Fix for a race condition when you have slow or fast servers and lock in redis isn't yes released because the lock was now the same length as the time the scheduler is been called. (bug noticed by @Dries-Vandenneucker and reported by myself in https://github.com/laravel/horizon/issues/935)

### DIFF
--- a/src/Console/SnapshotCommand.php
+++ b/src/Console/SnapshotCommand.php
@@ -31,7 +31,7 @@ class SnapshotCommand extends Command
      */
     public function handle(Lock $lock, MetricsRepository $metrics)
     {
-        if ($lock->get('metrics:snapshot', 300)) {
+        if ($lock->get('metrics:snapshot', config('horizon.metrics.snapshot_lock', 300))) {
             $metrics->snapshot();
 
             $this->info('Metrics snapshot stored successfully.');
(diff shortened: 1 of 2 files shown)


src/Console/SnapshotCommand.php
```diff
@@ -31,7 +31,7 @@ class SnapshotCommand extends Command
      */
     public function handle(Lock $lock, MetricsRepository $metrics)
     {
-        if ($lock->get('metrics:snapshot', 300)) {
+        if ($lock->get('metrics:snapshot', config('horizon.metrics.snapshot_lock', 300) - 30)) {
```
driesvints, line 34: The `-30` here is a slight change in behavior. We may not want to do that on a minor/patch release.

glamorous: That's the bugfix. Can be smaller too, but the point is:
1. Cron executes at 09:00:00
2. Command for metrics runs at 09:00:05 and is taken
3. Metric snapshot is taken
4. Cron executes at 09:05:00
5. Command for metrics runs at 09:05:01 and is not available, because lock is till 09:05:05.

That's the bug. The -30 is "a" solution, not the only one off-course. I'm open for suggestions.

PR #949, opened 2020-12-28
[5.x] Added ability to set local environment

0xb4lint:
With this PR you can set the `local_env` configuration option.
It defaults to the currently hardwired `local` value, so there is no breaking change.


driesvints:
Imo this isn't needed. It's meant to be a sensible default for the 99% use case of apps. If you have a different environment name for your local environment just configure the gate.

0xb4lint:
In this case my apps belong to the 1%. 😉 
I understand I can override this behavior in my `HorizonServiceProvider` (and `Horizon::check()` will follow), but I think I would be nice not to hardwire it.
Of course it's your call.

driesvints:
What's your use case for changing the default local env name?

0xb4lint:
I'm using branch names as environments (some kind of git flow), so in my case the `develop` is the "local" environment.

It wasn't a problem for ages, but I've just checked the Laravel framework code and I see there's 2 reference to `local` environment (`Application::isLocal()` and `TrustHosts::shouldSpecifyTrustedHosts()`). 

So I understand this is the expected behavior in Laravel, sorry for taking your time. 😉

driesvints:
@0xb4lint no sorry needed. Good luck with your app 👍

Issue #950, opened 2021-01-02
Dispatching multiple jobs within transaction (after_commit=true) leads to tags (for all jobs) of last pushed job


graemlourens:
- Horizon Version: 5.6.3
- Laravel Version: 8.19.0
- PHP Version: 7.4.12
- Redis Driver & Version: phpredis 5.3.2

### Description:
We are using the queue config option 'after_commit' set to true, in order that all jobs that are dispatched within transactions, are pushed after the transaction commits.

We use horizon tags, and saw lately that the tags did not match up with what the actual job.
After some digging we found out that when having after_commit set to true, the tags that are generated are taken from the previous job that was pushed to queue, not the actual one that i being pushed.

We followed this down the rabbits hole to `RedisQueue.php` in the method `pushRaw`. When after_commit is set to true, `$this->lastPushed` that is used to 'prepare' the job (extract tags etc) contains the last job that was pushed instead of the actual one that is being prepared.

### Steps To Reproduce:

1) Create 2 different jobs with each either manually tagging or passing a model as a constructor value
2) Set after_commit to true in your queue config.
3) Dispatch both jobs within a transaction:

```
        DB::transaction(function() {
            dispatch(new JobA());
            dispatch(new JobB());
        });
```

If you set a breakpoint in `RedisQueue.php` on the first line of `pushRaw` you will see that in both cases `$this->lastPushed` is equals to the JobB job instance, and therefore is passed along to the `prepare` method which will falsely determine the tags (and other things)

The tags that are determined (before being pushed to queue, and put into payload) will be those of JobB also for JobA.

I have tried to identify where this issue is coming from but can currently can not pinpoint it exactly nor propose a solution.

Hope this will make immediate sense to @themsaid

Kind regards, Graem

driesvints:
Thanks for the report @graemlourens. Mohamed has sent in a PR here: https://github.com/laravel/horizon/pull/951

graemlourens:
Fantastic, thank you for the swift fixing! I'll check the fix out in a couple of hours to confirm our logs are showing the correct tags.

graemlourens:
I can confirm the fix works as intended. Tags are not correctly assigned to the correct jobs. Thx again!

PR #951, opened 2021-01-04
[5.x] Set lastPushed when executing the delayed enqueue closure

themsaid:
This fixes https://github.com/laravel/horizon/issues/950

`$this->lastPushed = $job;` will run when the closure is executed. Before this PR it was set before the closure is executed so multiple jobs could have been dispatched after it.

### DIFF
--- a/src/RedisQueue.php
+++ b/src/RedisQueue.php
@@ -41,9 +41,17 @@ class RedisQueue extends BaseQueue
      */
     public function push($job, $data = '', $queue = null)
     {
-        $this->lastPushed = $job;
+        return $this->enqueueUsing(
+            $job,
+            $this->createPayload($job, $this->getQueue($queue), $data),
+            $queue,
+            null,
+            function ($payload, $queue) use ($job) {
+                $this->lastPushed = $job;
 
-        return parent::push($job, $data, $queue);
+                return $this->pushRaw($payload, $queue);
+            }
+        );
     }
 
     /**


wouter2203:
Horizon doesn't work anymore for me after this update. It throws an exception `Call to undefined method Laravel\Horizon\RedisQueue::enqueueUsing()`.

Without debugging this thoroughly, I noticed you wrapped the same call in a `method_exists` somewhere else in the same file: https://github.com/themsaid/horizon/blob/6c3704e9fb8eba4947a3a5f7e4a2968c0572d215/src/RedisQueue.php#L106.

Is that something this new block of code should have been wrapped in as well?

driesvints:
@wouter2203 Does updating the framework to the latest version help?

wouter2203:
@driesvints Yes, it does, that's something we noticed as well indeed. It probably has to do something with this update on the framework: https://github.com/laravel/framework/pull/35415

I guess that's what we get when we have the framework locked at a specific version and having horizon at `^5.0`. ;)

However, for the sake of clean code, I guess the `method_exists` should still be added. Or remove the other check as well and maybe bump the minimum version of `illuminate/queue` to whatever version the update was added in `illuminate/queue`.

_FYI: we have the framework locked at a specific version because it broke our project once while only a minor version was released. It had something to do with an edge case, but we manually update the framework since then so we can run our test suite and see if everything is still working as expected._

driesvints:
Yeah we should bump the minimum version for illuminate/queue. I'll check in on that but I guess it's sort of too late since we already tagged a new version.

PR #959, opened 2021-01-15
[5.x] Do not override all options provided by connection

edvinaskrucas:
By default when horizon redis connection is created in config it flushes all of extra options. By doing this we are not able to use redis sentinel configuration via predis driver. This PR merges default options with ones that are needed for horizon.


driesvints:
I think this is technically a breaking change as it could suddenly set config options that people might not expect. Perhaps better to be targeted at master?

edvinaskrucas:
Created new PR

PR #960, opened 2021-01-15
[6.x] Do not override all options provided by connection

edvinaskrucas:
By default when horizon redis connection is created in config it flushes all of extra options. By doing this we are not able to use redis sentinel configuration via predis driver. This PR merges default options with ones that are needed for horizon.

https://github.com/laravel/horizon/pull/959

### DIFF
--- a/src/Horizon.php
+++ b/src/Horizon.php
@@ -103,7 +103,9 @@ class Horizon
         }
 
         config(['database.redis.horizon' => array_merge($config, [
-            'options' => ['prefix' => config('horizon.prefix') ?: 'horizon:'],
+            'options' => array_merge($config['options'] ?? [], [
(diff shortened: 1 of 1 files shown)


taylorotwell:
And this fix is tested to fix it for your particular application?

taylorotwell:
Also, can you share an example of what your Redis configuration looks like?

edvinaskrucas:
I am using configuration laid out here [https://github.com/laravel/framework/pull/18850#issue-116339448](https://github.com/laravel/framework/pull/18850#issue-116339448), it works perfectly when not using horizon, as it omits options configuration and overrides it totally with its own when duplicating config group.

taylorotwell:
So, yes? Your changes do indeed work perfectly with your own application using Horizon?

edvinaskrucas:
Yes, it works as expected.

PR #988, opened 2021-03-15
[5.x] Throw exception when minProcesses is set to zero

themsaid:

### DIFF
--- a/src/ProvisioningPlan.php
+++ b/src/ProvisioningPlan.php
@@ -2,6 +2,7 @@
 
 namespace Laravel\Horizon;
 
+use Exception;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Str;
 use Laravel\Horizon\Contracts\HorizonCommandQueue;
@@ -176,6 +177,10 @@ class ProvisioningPlan
 
             return [Str::camel($key) => $value];
         })->all();
+
+        if (isset($options['minProcesses']) && $options['minProcesses'] === 0) {
+            throw new Exception("The value of [{$supervisor}.minProcesses] must be more than 0.");
+        }
 
         $options['parentId'] = getmypid();
 
--- a/src/ProvisioningPlan.php
+++ b/src/ProvisioningPlan.php
@@ -179,7 +179,7 @@ class ProvisioningPlan
         })->all();
 
         if (isset($options['minProcesses']) && $options['minProcesses'] === 0) {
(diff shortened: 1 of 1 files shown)


src/ProvisioningPlan.php
```diff
@@ -177,6 +178,10 @@ protected function convert($supervisor, $options)
             return [Str::camel($key) => $value];
         })->all();
 
+        if (isset($options['minProcesses']) && $options['minProcesses'] === 0) {
```
GrahamCampbell, line 181: `$options['minProcesses'] < 1`?

dillingham:
@themsaid thanks for adding this. Just a thought.. what if this and others described in your book were part of a `horizon:validate` command so they could be discovered sooner? Might also be a nice reference of all the ways we can misconfigure things.

src/ProvisioningPlan.php
driesvints: Updated that btw

ThaDaVos:
Out of curiosity - why was this changed? I've had no issues having it set to 0 - keeps the CPU nice and quiet for me - especially when running locally during development

driesvints:
There always needs to be at least one process. Otherwise it doesn't work.

ThaDaVos:
I've got a whole project setup with multiple queues (it's lumen, so probably used Horizon 3 or 4 I guess) - but all queues are set to minProcesses 0 - it may take a second, but they still do pickup their jobs and process them

bbrala:
I know this is a pretty old issue, but i revisited this again since we have a system which has a queue per tenant who arn't that critical to do extremly quickly, but we do want to balance over the difference queues so all keep progressing. We currently have a [little service provider](https://gist.github.com/bbrala/2509d96ff01d40aad229766720fdea22) that on start adjust the queue variable based on which queues have jobs queued and start those and run the horizon workers on cronjob that starts every minute and terminates after 2. Which kinda does what we want.

I'd rather just have a long living pod though and a lot of times we end up with only a few oif the processes having more than a few jobs queues.

Diving into the code I don't really see why things wouldnt work with scaling to 0. Queue status is not really monitored from the worker right? So don't really see what would break. Any insights on your original comment @driesvints?

I was thinking of using the Loop event in horizon to dynamicly adjust settings and kill ProcessPools at some point. Which could also work, and might work better in this context, but other projects we have would really benefit from a scale to 0.

PR #989, opened 2021-03-16
[5.x] Fix checking parent id

halaei:
I think my previous PR https://github.com/laravel/horizon/pull/881 for checking parent ID was incomplete. I don't quite remember how that happened. I tested this PR in a fresh Laravel application by running horizon and killing it with 9 signal. It works fine in my tests.

@driesvints could you take a look please?


driesvints:
This lgtm but please be aware that this won't be reviewed by Taylor until it's out of draft.

PR #992, opened 2021-03-19
Add Supervisor Rest Option

intrepidws:
This PR allows users to specify a custom value for the `rest` option when configuring a Supervisor in `config/horizon.php`, much like they can do with the `sleep`, `timeout` and `tries` options.

### DIFF
--- a/src/SupervisorOptions.php
+++ b/src/SupervisorOptions.php
@@ -131,6 +131,13 @@ class SupervisorOptions
     public $sleep;
 
     /**
+     * The number of seconds to rest between jobs.
+     *
+     * @var int
+     */
+    public $rest;
+
+    /**
      * The maximum amount of times a job may be attempted.
      *
      * @var int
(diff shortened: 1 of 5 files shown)


src/SupervisorOptions.php
```diff
@@ -160,6 +167,7 @@ class SupervisorOptions
      * @param  int  $memory
      * @param  int  $timeout
      * @param  int  $sleep
+     * @param  int  $rest
```
driesvints, line 170: Adding the new option in the middle of the construct is a major breaking change. Please move it to the end.

intrepidws: Right, sorry - still getting the hang of this and I always forget to consider breaking changes. This is done.

RomanSarvarov:
Guys, should it be documentated?

I'll try to do PR if someone explains how it works - because I can't even find where "rest" option is affects something.

PR #1010, opened 2021-04-19
[5.x] Adds domain and path environment variables

marcreichel:
Adds `HORIZON_DOMAIN` and `HORIZON_PATH` environment variables to config file.


taylorotwell:
These aren't typically things most people would change per environment - you are free to do this in your own application if you wish.

marcreichel:
@taylorotwell Why is this the default for Telescope then? In Telescope there are environment variables for `TELESCOPE_DOMAIN` and `TELESCOPE_PATH` which makes sense imo. For local development (with Valet e.g.) the domain would be `horizon.example.test` and for the production server it would be `horizon.example.com` or do I get something wrong?

PR #1011, opened 2021-04-20
Update PHPDoc variable type

adevade:
I think this is just a typo, since in the constructor further down it's type hinted as int.


driesvints:
Thanks

Issue #1018, opened 2021-05-12
Horizon Dashboard does not work with strict content security policy


NEGits:
- Horizon Version: 5.7.6
- Laravel Version: 8.41.0
- PHP Version: 7.4
- Redis Driver & Version: predis 1.1.7

### Description:

Apparently the Horizon dashboard uses ```eval```, which makes it unusable for us in production. I'm not in a position to disable the CSP for security reasons. And I'm not familiar with Vue, so I'm not sure how to solve this. 

Apart from this it also loads an external font from google fonts which is blocked, but I don't think this breaks functionality. Making ```layout.blade.php``` publishable and editable would enable us to fix/work around this. 

### Steps To Reproduce:

Set a CSP which excludes the use of ```unsafe-eval``` and external style sources, observe the dashboard not loading and producing errors in the javascript console:

```Content Security Policy: The page’s settings blocked the loading of a resource at https://fonts.googleapis.com/css?family=Nunito (“style-src”).```
```Content Security Policy: The page’s settings blocked the loading of a resource at eval (“script-src”).```

driesvints:
> Apparently the Horizon dashboard uses eval

Hmm, I can't find that. Can you point out where exactly?

> google fonts which is blocked

It falls back to the default sans-serif font so this isn't that much of an issue I think.

I don't think we're going to take action here since nothing really is broken. Sorry

NEGits:
@driesvints 

This has been reported before (should have added that, I'm sorry) see https://github.com/laravel/horizon/issues/576

Vue seems to use ```eval``` in some cases. I'm not familiar with Vue, so I'm not sure what the best solution to this is. https://forum.vuejs.org/t/csp-problem-with-eval/17581 Seems to suggest using the vue runtime build fixes this. https://github.com/laravel/horizon/issues/576 implies this issue has been introduced into horizon with version 3 and could be fixed in another way, but having never done anything with Vue I'm not sure.

I would say that this is unwanted and should be considered a bug, as setting a strict CSP seems to be more and more common, but that's just my perspective.

In our production environment the CSP headers are set by the webserver, and I'm not in a position to change them for horizon.

driesvints:
If you can send in a non-breaking pr we could consider it maybe.

NEGits:
I'll try freeing up some time in the coming week to take a stab at it. Would you consider reopening this so someone with experience with Vue might try to fix this?

driesvints:
No sorry. Just send in a pr if you're willing.

Issue #576, opened 2019-04-17
Horizon doesn't work when using content security policy headers


tomgrohl:
- Horizon Version: 3.1.1
- Laravel Version: 5.8.11
- PHP Version: 7.2
- Redis Driver & Version: predis 1.1.1
- Database Driver & Version: Mysql 5.7.25

### Description:

Horizon doesn't work when using content security policy headers. Because of the changes to the view files between versions 1 and 3. i.e going from an empty app:

```
<div id="root"></div>
```

to a div with html elements and a Vue component inside it,

```
<div id="horizon" v-cloak>
...
</div>
```

This causes Vue to use `eval` which is unsafe and not allow by CSP.

### Steps To Reproduce:

1. Add a middleware for content security policy and register in `app.php`:

```php
<?php

namespace App\Http\Middleware;

use Closure;
use Symfony\Component\HttpFoundation\Response;

/**
 * Class ContentSecurityPolicy
 *
 * @package App\Http\Middleware
 */
class ContentSecurityPolicy
{
    /**
     * Handle an incoming request.
     *
     * @param  \Illuminate\Http\Request  $request
     * @param  \Closure  $next
     * @return mixed
     */
    public function handle($request, Closure $next)
    {
        $policy = [
            'default-src' => [
            ],
            'script-src' => [
                '\'unsafe-inline\'',
                'data:',
                '*.googleapis.com',
                '*.twitter.com',
                '*.facebook.net',
                'www.googleadservices.com',
                'www.gstatic.com',
                'www.google.com',
                'google.com',
                'google.co.uk',
                'use.fontawesome.com',
            ],
            'style-src' => [
                '\'unsafe-inline\'',
                'fonts.googleapis.com',
                'tagmanager.google.com',
                'maxcdn.bootstrapcdn.com',
                'cdn-images.mailchimp.com',
                'use.fontawesome.com',
            ],
            'img-src' => [
                '*',
                'data:',
            ],
            'font-src' => [
                'data:',
                'fonts.googleapis.com',
                'fonts.gstatic.com',
                'use.fontawesome.com',
            ],
            'connect-src' => [
                'fonts.googleapis.com',
            ],
            'frame-src' => [
                'www.google.com',
            ],
        ];


        /** @var Response $response */
        $response = $next($request);

        if ('local' === app()->environment()) {
            $policy['script-src'][] = '\'unsafe-eval\'';
        }
        
        array_walk($policy, function(&$item, $key) {
[...]

PR #1022, opened 2021-05-14
Move contents of #horizon to separate component

NEGits:
This PR moves the contents of the `#horizon` div to a separate component. This fixes https://github.com/laravel/horizon/issues/1018 and https://github.com/laravel/horizon/issues/576.

The html in the root Vue element cause Vue to use `new Function`, which is not allowed when using a Content Security Policy that does not include `'unsafe-eval'`, which causes the Horizon Dashboard to not work at all if such a CSP is present. Moving the content into a separate vue component fixes this. (ref. https://github.com/JeffreyWay/laravel-mix/issues/1052#issuecomment-318469036)

As far as I can tell nothing breaks due to this PR. 

I was not sure if I should have included the compiled js in this PR, can remove if wanted.


taylorotwell:
Reverted.

NEGits:
@taylorotwell I'm sorry for that bug, should have caught it. Would you consider a second try at this with a working auto load? Reverting this makes the horizon dashboard unusable with a strict CSP unfortunately.

francoism90:
Any other workaround available?

PR #1037, opened 2021-06-30
Show number of reserved + waiting jobs in dashboard

halaei:
In a stable system, the number of waiting jobs for each queue should be 0 most of the time, while the number of jobs being processed may be greater than 0.
Laravel dashboard currently shows the number of waiting jobs per queue in jobs column which is most on the  time zero. Therefore, it doesn't give us feeling on how crowded each queue is. It also shows the number of workers for each queue, in processes column.
This PR, changes the value showed in "Jobs" column to be `number_of_waiting_jobs + number_of_processing_jobs`. The sum gives more valuable information, and still we can easily know how many jobs are waiting.

Unless someone is using `RedisWorkloadRepository` elsewhere, I don't regard it as a breaking change. Let me know if you want me to send it for 6.x.

image
Before this change, all numbers was constantly 0 all the times. Now I know more, and guess 35 workers is a little bit too many.

### DIFF
--- a/src/RedisQueue.php
+++ b/src/RedisQueue.php
@@ -32,6 +32,28 @@ class RedisQueue extends BaseQueue
     }
 
     /**
+     * Get the number of queue jobs that are processing.
+     *
+     * @param string|null $queue
+     * @return int
+     */
+    public function reservedNow($queue = null)
+    {
+        return $this->getConnection()->zcard($this->getQueue($queue).':reserved');
+    }
+
+    /**
+     * Get the number of queue jobs that are ready to process or processing.
+     *
+     * @param string|null $queue
+     * @return int
+     */
+    public function lengthNow($queue = null)
+    {
+        return $this->readyNow($queue) + $this->reservedNow($queue);
+    }
+
+    /**
      * Push a new job onto the queue.
      *
      * @param  object|string  $job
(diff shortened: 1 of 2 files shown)


themsaid:
hmmm let's say you have 30 workers processing jobs, the column will now show 30 and you have no way of knowing if these are 30 jobs being processed or 30 jobs currently in the queue.

Making this change my confuse people who expect to see this number as 0 most of the time because now it'll start showing jobs being processed.

taylorotwell:

halaei:
@themsaid @taylorotwell  I could show both waiting and processing jobs. But please note that, with blocking pop, jobs almost immediately get processed after being pushed to the queue, therefore the number of waiting jobs will be almost always zero, unless all the workers are busy.
Therefore, with high confidence we can say the following equations are true most of the time:
```
waiting_jobs = max(0, jobs - processes)
processing_jobs = min(jobs, processes)
```
It is not always true, because there is a very small duration from the time worker is aware of a new job to the time it actually reserves the job.
I decided not to show 2 columns in this PR, one for waiting and the other for processing jobs. But if you prefer, I can show both the numbers as well.

I understand this change can be confusing as you wrote. I can add some comments in change log or upgrade guide. Do you think I should send it for 6.x?

### PS
I also find it useful to show 2 other columns for the maximum value of jobs and wait time for each queue. The max values are stored in client side and reset to zero every time the user refreshes the dashboard. Please let me know what you think about it as well. Note that I change the time formats to show seconds and milliseconds instead of using `moment.humanize()`, because in my case, working with jobs that must be immediately processed, milliseconds is actually very important.

image

Source: https://github.com/halaei/horizon/tree/show-max-jobs

PR #1057, opened 2021-08-18
[5.x] Providing directory to Symfony process

piurafunk:
I found that the directory specified in config in `horizon.environments.{environment}.directory` wasn't provided to the Symfony process. It tries to:
```
Process::fromShellCommandline($command, $options->directory ?? base_path())
```
but `directory` is null on that `$options` object. I found it was due to `\Laravel\Horizon\SupervisorOptions::toArray()` not including the `$directory` attribute. Once I added that in, it seems to obey the directory value. It doesn't add it to the `\Laravel\Horizon\SupervisorOptions::toSupervisorCommand()` or `\Laravel\Horizon\SupervisorOptions::toWorkerCommand()` strings, which should maintain compatibility with the `illuminate/queue` package.

This allows developers to specify which directory `artisan queue:work` should be run from. The primary reason (and my reason) is to take advantage of symlinks on a server, where `base_path()` resolves symlinks to real paths.

### DIFF
--- a/src/SupervisorOptions.php
+++ b/src/SupervisorOptions.php
@@ -194,7 +194,8 @@ class SupervisorOptions
                                 $balanceCooldown = 3,
                                 $balanceMaxShift = 1,
                                 $parentId = 0,
-                                $rest = 0)
+                                $rest = 0,
+                                $directory = null)
     {
         $this->name = $name;
         $this->connection = $connection;
@@ -216,6 +217,7 @@ class SupervisorOptions
         $this->balanceMaxShift = $balanceMaxShift;
         $this->parentId = $parentId;
         $this->rest = $rest;
+        $this->directory = $directory;
     }
 
     /**
@@ -309,6 +311,7 @@ class SupervisorOptions
             'balanceMaxShift' => $this->balanceMaxShift,
             'parentId' => $this->parentId,
             'rest' => $this->rest,
+            'directory' => $this->directory,
         ];
     }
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
I still don't understand how it will ever be set? It is never passed to the constructor of `SupervisorOptions` anywhere in the code?

@themsaid

taylorotwell:
I don't think this "directory" feature was ever intended to be used or maybe was an idea during the initial development of Horizon that ended up not being necessary.

piurafunk:
To point out where `directory` gets provided to `SupervisorOptions`, :

`\Laravel\Horizon\Console\HorizonCommand::handle` calls `\Laravel\Horizon\ProvisioningPlan::get`
`\Laravel\Horizon\ProvisioningPlan::get` calls `\Laravel\Horizon\ProvisioningPlan::__construct`, pulling in `config('horizon.environments')` (which can have the `directory` key defined).
`\Laravel\Horizon\ProvisioningPlan::__construct` calls `\Laravel\Horizon\ProvisioningPlan::toSupervisorOptions`
`\Laravel\Horizon\ProvisioningPlan::toSupervisorOptions` calls `\Laravel\Horizon\ProvisioningPlan::convert`
`\Laravel\Horizon\ProvisioningPlan::convert` calls `\Laravel\Horizon\SupervisorOptions::fromArray`

That's how it gets set on the object. Then when passing it to each child process:

`\Laravel\Horizon\MasterSupervisorCommands\AddSupervisor::process` rebuilds the `SupervisorOptions` object from array. This array will have the `directory` key in it with my changes.

---

I get that this isn't going to be a widely used "feature", but it sure would simplify my use case. Our multi-tenancy setup relies on the path being one of a specific set to load configuration correctly, and we use symlinks to maintain a single application that multiple tenants can use (symlink path indicates the tenant key).

With this info (confirming that the `directory` key would get provided to the `SupervisorOptions` and providing a real-world use case), is this something that can be added?

If not, is anyone against me submitting a PR to remove the `directory` references, so that future devs don't think this is a feature and try to use it (as I did, which inpsired this PR).

Issue #365, opened 2018-07-12
Customizable Laravel Horizon dashboard route


jofomah:
I would like to link Laravel Horizon route `/routes` to my admin navigation  bar and also my app admin `logout` and `home` route to `/horizon` navbar so that i can easily navigate from Horizon dashboard to my admin dashboard and also logout from Horizon dashboard.

I.e is there a way to add a custom route to Horizon navbar so that it is easy to link it to ones app routes???

driesvints:
Horizon is a SPA which means everything is rendered client side. Since everything is compiled I don't see how we could offer a way through the package to make this happen.

What you *could* theoretically do is overwrite the base view and build some styling/html around it. Add a file to this location: `resources/views/vendor/laravel/horizon/app.blade.php` with the contents of https://github.com/laravel/horizon/blob/3218e86cea737643fa626fc98d56a3b846f8ba2a/resources/views/app.blade.php and you could build around it as long as you allow the SPA to generate everything to the root div.

abdosaeedelhassan:
may be switching to build every pages using livewire components solve the problem

PR #1072, opened 2021-09-21
Adding site_path to link back to site.

rrpadilla:
This PR adds a new config **site_path** to add a link back to the site. This is optional, no breaking change.

See related issue: #365.

Example: 

```php
'site_path' => env('HORIZON_SITE_PATH', '/dashboard'),
```


Restingo:
Maybe take the base url as default fallback?

taylorotwell:
I would personally prefer a smaller "Back to application" link or something.

rrpadilla:
Hi @taylorotwell ,
what do you think if only the app name is the link in this case?

laravel-horizon

PR #1074, opened 2021-09-22
Clickable app name to go back to application.

rrpadilla:
This PR adds a new config **site_path** to add a link back to the site. This is optional, no breaking change.
In this case only the app name will be the link to go back to the application.
> Taylor says: I would personally prefer a smaller "Back to application" link or something.

If you still prefer it let me know the best place for this.

Example: 

```php
'site_path' => env('HORIZON_SITE_PATH', '/dashboard'),
```

Link:

laravel-horizon

No link / current view:

laravel-horizon-no-link


taylorotwell:
I really think everything is fine as is.

PR #1081, opened 2021-09-30
[5.x] Update forget failed command to use job uuid instead of job id

### LINKED ISSUE #1080: ForgetFailedCommand should have used uuid instead of id when forgetting a failed job
- Horizon Version: 5.7.13
- Laravel Version: 8.6.2
- PHP Version: 7.4.23

### Description:
ForgetFailedCommand uses Job ID to be removed from horizon redis. However, when storing a failed job unto Redis, Horizon always uses UUID instead of ID. This causes the job to be successfully removed from the failed_job database table but not removed from Horizon.

To fix this, ForgetFailedCommand must be updated to remove the failed job from Redis using job UUID as key instead of job ID

### Steps To Reproduce:
1) Create a job which always fail
2) Trigger a job
3) Notice that a new entry is created in the failed_jobs table
4) Try "forgetting" that failed job by running `php artisan horizon:forget <job-id>` as explained in the [docs](https://laravel.com/docs/8.x/horizon#deleting-failed-jobs)
5) Job would be removed from failed_jobs database but running `HGETALL 'horizon:<job-uuid>' against redis shows that the job still remain in Redis.

ocind:
fixes #1080

### Summary 
This PR fixes a bug on `ForgetFailedCommand` where it tries to remove a job from Horizon Redis using Job ID instead of Job UUID.

### Description
`ForgetFailedCommand`uses Job ID to be remove a particular job entry from Horizon's Redis. However, when storing a failed job unto Redis, Horizon always uses UUID instead of ID. 

This causes a bug whereby the job is successfully removed from the `failed_job` database table but not removed from Horizon (since there is no such key with `horizon:job-id`, the correct key is `horizon:job-uuid`)

To fix this, ForgetFailedCommand must be updated to remove the failed job from Redis using job UUID as key instead of job ID

### Steps To Reproduce

1) Create a job which always fail/throw exception
2) Trigger the job
3) Notice that a new entry is created in the `failed_jobs` table
4) Try "forgetting" that failed job by running `php artisan horizon:forget job-id` as explained in the docs
5) Job would be removed from failed_jobs database but running `HGETALL 'horizon:job-uuid'` against redis shows that the job still remain in Redis.

### DIFF
--- a/src/Console/ForgetFailedCommand.php
+++ b/src/Console/ForgetFailedCommand.php
@@ -28,7 +28,11 @@ class ForgetFailedCommand extends Command
      */
     public function handle(JobRepository $repository)
     {
-        $repository->deleteFailed($this->argument('id'));
+        $failedJobPayload = $this->laravel['queue.failer']->find($this->argument('id'))->payload ?? null;
+        if (! is_null($failedJobPayload)) {
+            $jobId = json_decode($failedJobPayload, true)['uuid'] ?? $this->argument('id');
+            $repository->deleteFailed($jobId);
+        }
 
         if ($this->laravel['queue.failer']->forget($this->argument('id'))) {
             $this->info('Failed job deleted successfully!');


driesvints:
@ocind please add a thorough explanation to a PR and not just link to an issue. Otherwise your PR will be closed.

ocind:
@driesvints i copied and pasted the issue description here. Is the explanation clear enough?

taylorotwell:
Just pass the job's UUID to the command if you want this behavior.

ocind:
@taylorotwell passing the job UUID means, clearing the job UUID will fail since in the DB it is stored using job ID

ocind:
@driesvints any idea about this? I think this is clearly an unintended behaviour and I think it needs to be fixed?

driesvints:
@ocind You need to use:

```
php artisan horizon:forget "<the uuid here>"
```

Doesn't that work for you?

driesvints:
Btw, in general, I very much recommend to keep your defaults up to date with the new changes in each Laravel release. Tools like https://laravelshift.com can automate this for you.

ocind:
yeah, it was actually documented on the upgrade guide but we did not know the impact of [this](https://laravel.com/docs/8.x/upgrade#failed-jobs-table-batch-support) last time.

Will definitely checkout https://laravelshift.com/!

PR #1091, opened 2021-11-01
Reduce bundle size

sebdesign:
This PR reduces the bundle size by updating an outdated webpack configuration.
The webpack.IgnorePlugin has changed its syntax since it was added in 2017.

So now the bundle is reduced from 1160kb to 922kb!

See: https://webpack.js.org/plugins/ignore-plugin/#example-of-ignoring-moment-locales

I've submitted the same fix in https://github.com/laravel/telescope/pull/1142

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -6,6 +6,9 @@
 *.md diff=markdown
 *.php diff=php
 
+public/*.css binary
+public/*.js binary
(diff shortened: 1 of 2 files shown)


driesvints:
Thanks @sebdesign

PR #1095, opened 2021-11-16
Add ability to delete a pending job

sgnatyuk:
It is possible to delete jobs which are in status "pending". For this feature was added button "Delete" and checkboxes near each job for selecting, which should be removed. Also it could be selected all jobs on page. After button was pressed, ids of selected jobs send to `PendingJobsController:batchDelete` and removed from Redis.


themsaid:
This just deletes the pending jobs from the horizon cache not from the actual queue, those deleted jobs will still run. They just won't appear in Horizon. Also the batches functionality will show wrong numbers if any of the job batches were deleted.

Issue #897, opened 2020-09-25
minProcesses has no effect with balance=auto


robj:
- Horizon Version: 5.0
- Laravel Version: 8.0
- PHP Version: 7.4

### Description:

with minProcesses=1  & balance=auto:

workers per queue will still scale down to 0

Jobs will be assigned to a queue that has been scaled down to 0 for very long periods (eg. >10mins) before a worker is finally assigned to the queue. It is important to us to have at least 1 worker always assigned to priority queues to prevent delays.

driesvints:
Hey there,

Can you first please try one of the support channels below? If you can actually identify this as a bug, feel free to report back and I'll gladly help you out and re-open this issue.

- [Laracasts Forums](https://laracasts.com/discuss)
- [Laravel.io Forums](https://laravel.io/forum)
- [StackOverflow](https://stackoverflow.com/questions/tagged/laravel)
- [Discord](https://discordapp.com/invite/KxwQuKb)
- [Larachat](https://larachat.co)
- [IRC](https://webchat.freenode.net/?nick=laravelnewbie&channels=%23laravel&prompt=1)

Thanks!

robj:
I don't think rapidly closing all Issues reported to a project with templates makes for a good project ;)

driesvints:
@robj can you please try a support channel first? Happy help you out if you've confirmed that this is a bug.

robj:
@driesvints - Looking at closed tickets, I am the second person to report a similar issue in 48hrs. If you left Issues open instead of 'OCD Inbox Zeroing' then that would have been more visible.

driesvints:
@robj I just tried this and as I suspected this worked just fine for me. I'm suspecting there's something else at fault for causing this to fail for you but that's probably best to be figured out as a support request through one of the links above.

I closed this because I don't have the time to provide support for every single issue. There's about a hundred issues a day that I handle and if I had to tackle each of them myself personally I wouldn't be able to keep up. We have a large community ready to help you out on our support channels. If at some point, you are absolutely sure you've pinpointed a bug then you're free to post a thorough issue with the details on how to recreate it.

In general I want to ask you to be a little bit more considerate for project maintainers the next time you post an issue.

```php
    'defaults' => [
        'supervisor-1' => [
            'connection' => 'redis',
            'queue' => ['default'],
            'balance' => 'auto',
            'minProcesses' => 1,
            'maxProcesses' => 5,
            'tries' => 1,
            'nice' => 0,
        ],
    ],

[...]

PR #1096, opened 2021-11-16
Prevent pool to receive more than available processes on scaling

ferdinandfrank:
As described in https://github.com/laravel/horizon/issues/897 a queue might currently scale down to 0 processes even though a minimum amount of processes per queue is configured. 
I observed the behavior especially when one queue needs to process a lot of jobs over a longer period of time (`system_function` in the screenshot). In that case it sometimes receives all processes from other busy queues (`low` in the screenshot) which leads to the other busy queue (`low`) receiving 0 processes as all others are already busy even though the `minProcesses` in the config is set to 3.
image

To prevent this issue this pull request ensures that a queue will not receive more than the available processes when scaling up by considering the configured `minProcesses` and the amount of configured queues using the following formula: 
`[maxProcesses] - (([amount of queues] - 1) * [minProcesses])`

This ensures that every queue has the configured amount of `minProcesses` available.

### DIFF
--- a/src/AutoScaler.php
+++ b/src/AutoScaler.php
@@ -139,7 +139,7 @@ class AutoScaler
             $pool->scale(
                 min(
                     $totalProcessCount + $maxUpShift,
-                    $supervisor->options->maxProcesses,
+                    $supervisor->options->maxProcesses - (($supervisor->processPools->count() - 1) * $supervisor->options->minProcesses),
                     $desiredProcessCount
                 )
             );


taylorotwell:
Thanks for contributing to Laravel! ❤️

jf-m:
Hi @ferdinandfrank 

I'm still having having the 0 processes issue, but to my research, it does looks like it comes from this pull request.

I am not sure to fully understand what you were trying to do with this formula, so here is break-down of my current situation applied to this PR change : 


AutoScaler config 
```
maxProcesses: 8
minProcesses: 4
maxUpShift: 1

processPools: 4
``` 

Your current formula : `$supervisor->options->maxProcesses - (($supervisor->processPools->count() - 1) * $supervisor->options->minProcesses)`

Is then `8 - ((4 - 1) * 4)` = `-4`.
The `-4` value is the picked as the minimum value, therefore 0 processes are created.


Before this PR, the old parameter would have worked just fine :
```
$pool->scale(
                min(
                    $totalProcessCount + $maxUpShift,
                    $supervisor->options->maxProcesses,
                    $desiredProcessCount
                )
            );
```

Using only `$supervisor->options->maxProcesses` in the `min` function should be enough to guarantee that this number will never be exceeded, right ? 

I might miss something here, hence why I'm poking you  :) !

PR #1115, opened 2022-01-28
[5.x] Add exception context on failed jobs

antoinelame:
This PR adds the exception context when displaying a failed job.

Capture d’écran 2022-01-28 à 02 06 20

### DIFF
--- a/resources/js/screens/failedJobs/job.vue
+++ b/resources/js/screens/failedJobs/job.vue
@@ -198,6 +198,16 @@
             </div>
         </div>
 
+        <div class="card mt-4" v-if="ready">
+            <div class="card-header d-flex align-items-center justify-content-between">
+                <h5>Exception Context</h5>
+            </div>
+
+            <div class="card-body code-bg text-white">
+                <vue-json-pretty :data="prettyPrintJob(job.context)"></vue-json-pretty>
+            </div>
+        </div>
+
 
         <div class="card mt-4" v-if="ready">
             <div class="card-header d-flex align-items-center justify-content-between">
--- a/src/Http/Controllers/FailedJobsController.php
+++ b/src/Http/Controllers/FailedJobsController.php
@@ -117,6 +117,8 @@ class FailedJobsController extends Controller
 
         $job->exception = mb_convert_encoding($job->exception, 'UTF-8');
 
+        $job->context = json_decode($job->context);
+
         $job->retried_by = collect(json_decode($job->retried_by))
                     ->sortByDesc('retried_at')->values();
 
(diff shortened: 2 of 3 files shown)


driesvints:
This is a good pr but we'll need to make sure it's backwards compatible with existing horizon installations. To try this out, install an existing Laravel 8 app with current horizon, log a few jobs and afterwards update to your changes in the PR and log a few new jobs. Make sure the old ones still function.

antoinelame:
👋 @driesvints! I installed it on two existing and well-used Laravel 8 applications, and everything has been working like a charm for the past 5 days. It has been pretty useful for us to have the exception context directly in Horizon (like Telescope already does), rather than having to read it in the log file.

driesvints (approved):

driesvints:
Cool, thanks @antoinelame. Taylor will review this.

taylorotwell:
Does this only work for custom exceptions in your application that define a `context` method? `context` is not a method that is present on all exceptions.

antoinelame:
Yep, it only works for exceptions that define a `context` method.

Do you think it's an issue? The idea of the PR is to unify the use of the `context` method between the Handler, Telescope and Horizon, given that:

- Illuminate Handler specifically adds this method data, if it exists, to the logs ([see here](https://github.com/laravel/framework/blob/8.x/src/Illuminate/Foundation/Exceptions/Handler.php#L289))
-  Telescope also adds this method data to the "Exception" screen ([see here](https://github.com/laravel/telescope/blob/4.x/resources/js/screens/exceptions/preview.vue#L108))

What do you think?

taylorotwell:
This does not unify that though. Note that in Telescope the context is derived from the Logging system - where context can be applied globally to all logged messages. This PR only pulls context from the exception's `context` method if it exists. So, they are not entirely the same implementation and the data will not always match between these Horizon and Telescope in this case.

antoinelame:
Indeed, I see your point.

I don't see a simple way to implement the functionality in the same way as Telescope, since Horizon doesn't rely on `MessageLogged` event. I will see if I can change the implementation. Maybe we should close the PR in the meantime.

taylorotwell:
I don't think it's a bad PR and it might be better than nothing.

flexchar:
Hi @antoinelame! An amazing addition to the ecosystem.

I'm looking forward to using it. However struggling to find how exactly the context is added. I myself checked Laravel documentation - doesn't mention anything on the horizon page.

I'd greatly appreciate if you throw an example job code or reference a documentation article!

antoinelame:
Hi @flexchar, you just have to declare the `context()` method in your exception. You can find an example [in the documentation](https://laravel.com/docs/9.x/errors#exception-log-context).

PR #1119, opened 2022-02-09
[6.x] Update to bootstrap 5.1 and drop jQuery

mmachatschek:
This PR updates bootstrap from 4.6 to 5.1 and applies the migration steps from https://getbootstrap.com/docs/5.1/migration/

Additionally I managed to remove jQuery and replace its usage with vanilla JS.

This does mainly the same thing like https://github.com/laravel/telescope/pull/1179 and additionally removes the popper.js dependency.


browner12:
can we just update to Tailwind instead?

mmachatschek:
@browner12 would be doable. I think the only problems that came to my mind was modals which would need a custom implementation. Luckily the tailwindcss ecosystem provides headlessui which gives us modals out of the box.

I would be open to this. depends on Taylor

taylorotwell:
I've never made a big effort to change to Tailwind because that end user doesn't know or care how it is implemented and therefore there is no "return on investment" in changing it - just the risk of introducing regressions.

browner12:
I think there are 2 benefits I'd advocate for in switching to Tailwind which lend to an ROI.

1. Consistency across the ecosystem. There seems to be a mix across the different products currently, but the trend definitely is towards Tailwind. Consistency might encourage more contribution, and make maintenance for the internal team easier.

```
Bootstrap
- Envoyer
- Horizon
- Telescope

Tailwind
- laravel.com
- Forge
- Vapor
- Spark
- Nova
- Default Pagination
```

2. Ease of upgrade. Comparatively, Tailwind has so far been leaps and bounds easier to upgrade than Bootstrap. Yes, the initial transition here would be time consuming, but future upgrades would theoretically be easier. Per your comment about the "risk of regressions", upgrading Bootstrap definitely gives me more cause for concern on regressions, because I've done that upgrade and know it can be difficult.

---

Obviously I'm not the one who has to spend the time coding it, so just my opinion. For the business I understand how it's difficult to justify.

driesvints:
@mmachatschek totally forgot about this PR and we're now getting a whole bunch of merge conflicts... I feel like this PR could have perfectly gone into 5.x as well. If you're up for it, feel free to send it that way!

mmachatschek:
@driesvints I try to resend it to the 5.x branch. but that probably needs some additional checks as jess has made a lot of changes to the design. i need to go through the migration guide again

PR #1126, opened 2022-02-21
Clear timers during each queue worker iteration

### LINKED ISSUE #1125: Memory leak due to job duration recording
- Horizon Version: `v5.9.1`
- Laravel Version: `v9.1.0`
- PHP Version: `8.0.16`
- Redis Driver & Version: `phpredis v5.3.6`
- Database Driver & Version: `sqlsrv v5.9.0`

### Description:
Using Horizon to monitor queue jobs causes a memory in the queue workers. The memory increases with the number of jobs being executed.

The [`Stopwatch` class](https://github.com/laravel/horizon/blob/ff534ce86fa27623d49da8a4272d3301b7998e6f/src/Stopwatch.php#L5) is [registered as singleton via `$serviceBindings`](https://github.com/laravel/horizon/blob/ff534ce86fa27623d49da8a4272d3301b7998e6f/src/HorizonServiceProvider.php#L174), which causes a memory leak in its `$timers` array.
The [`start($key)` method is called for started jobs](https://github.com/laravel/horizon/blob/14b0a3cabb195540dc5c35cb98cb1d0fbdd0d848/src/Listeners/StartTimingJob.php#L36) and [`check($key)` for stopped jobs](https://github.com/laravel/horizon/blob/ff534ce86fa27623d49da8a4272d3301b7998e6f/src/Listeners/UpdateJobMetrics.php#L50). There is no code which removes entries from the `$timers` array.

### Steps To Reproduce:

- Add some logging to [`Stopwatch::check($key)`](https://github.com/laravel/horizon/blob/ff534ce86fa27623d49da8a4272d3301b7998e6f/src/Stopwatch.php#L31-L36) which counts the `$timers` to see them increasing (per queue worker) or use some other tool to monitor the array (xdebug):
    ```php
    class Stopwatch
    {
        public function check($key)
        {
            Log::debug('Stopwatch', ['count' => count($this->timers)]);

            if (isset($this->timers[$key])) {
                return round((microtime(true) - $this->timers[$key]) * 1000, 2);
            }
        }
    }
    ```
- Run a few queue jobs (it doesn't matter what they do)
- See the `$timers` increasing:
    > {"message":"Stopwatch","context":{"count":1},"level":100,"level_name":"DEBUG","datetime":"2022-02-20T11:14:37.782784+01:00"}
    > {"message":"Stopwatch","context":{"count":2},"level":100,"level_name":"DEBUG","datetime":"2022-02-20T11:14:37.782784+01:00"}
    > {"message":"Stopwatch","context":{"count":3},"level":100,"level_name":"DEBUG","datetime":"2022-02-20T11:14:37.784126+01:00"}
    > {"message":"Stopwatch","context":{"count":4},"level":100,"level_name":"DEBUG","datetime":"2022-02-20T11:14:37.785836+01:00"}

### Why this Issue?

I'm wondering whether I overlooked something or if this really is a 5 year old bug (last changes to `Stopwatch` were in 2017).

Also are there many ways to solve this issue, e.g.:
- by removing the entry from `$timers` in `check($key)`,
- by removing the entry from `$times` with a separate method after calling `check($key)`,
- by using a single `$timer` in `Stopwatch` (why have multiple anyway?) or
- by resetting the `Stopwatch::$timers` in `Queue::looping($callback)` using a new `Stopwatch::reset()` method.

I'm happy to provide a PR if you let me know which fix is the desired one.

Namoshek:
Fixes #1125 by clearing the `Stopwatch::$timers` while the queue worker is looping. Using the `Looping` event ensures the timers are reset before taking on the next job.

I've created [a repository with a minimal reproducible example](https://github.com/Namoshek/laravel-horizon-memory-leak).

### Alternatives

If using the `Looping` event in Horizon is not desired or seems excessive due to the listener also being run if no job has been executed, it may also work to clear the timers from listeners for the `JobDeleted`, `JobFailed` and maybe `JobReleased` events. Since I'm not entirely sure when which of these events is run, I opted for the safest solution to always clear the timers.

### DIFF
--- a/src/EventMap.php
+++ b/src/EventMap.php
@@ -67,5 +67,9 @@ trait EventMap
         Events\LongWaitDetected::class => [
             Listeners\SendNotification::class,
         ],
+
+        \Illuminate\Queue\Events\Looping::class => [
+            Listeners\ClearJobTimers::class,
+        ],
     ];
 }
--- a/src/Listeners/ClearJobTimers.php
+++ b/src/Listeners/ClearJobTimers.php
@@ -1,0 +1,36 @@
+<?php
+
+namespace Laravel\Horizon\Listeners;
+
+use Laravel\Horizon\Stopwatch;
+
+class ClearJobTimers
+{
+    /**
+     * The stopwatch instance.
+     *
+     * @var \Laravel\Horizon\Stopwatch
+     */
+    public $watch;
+
+    /**
+     * Create a new listener instance.
+     *
+     * @param  \Laravel\Horizon\Stopwatch  $watch
+     * @return void
+     */
+    public function __construct(Stopwatch $watch)
+    {
+        $this->watch = $watch;
+    }
+
+    /**
+     * Handle the event.
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        $this->watch->clear();
+    }
+}
--- a/src/Stopwatch.php
+++ b/src/Stopwatch.php
@@ -34,4 +34,14 @@ class Stopwatch
             return round((microtime(true) - $this->timers[$key]) * 1000, 2);
         }
     }
+
+    /**
+     * Clear all registered timers.
+     *
+     * @return void
+     */
+    public function clear()
+    {
+        $this->timers = [];
+    }
 }


taylorotwell:
Decided to take this approach: https://github.com/laravel/horizon/commit/69486d5e4fad8b6ee02c97ed0bea39ef241df956

Thanks!

Namoshek:
@taylorotwell Is this safe in case of exceptions? I added a [branch with your fix in my demo repo](https://github.com/Namoshek/laravel-horizon-memory-leak/tree/test-taylors-fix) and from my understanding, this doesn't work in case of exceptions:

> [2022-02-21 20:13:27] local.DEBUG: The first job is run.
[2022-02-21 20:13:27] local.DEBUG: Stopwatch {"timers":0}
[2022-02-21 20:13:27] local.DEBUG: The first job is run.
[2022-02-21 20:13:27] local.DEBUG: Stopwatch {"timers":0}
[2022-02-21 20:13:30] local.DEBUG: The first job is run.
[2022-02-21 20:13:30] local.DEBUG: Stopwatch {"timers":0}
[2022-02-21 20:13:30] local.DEBUG: The second job is started.
[2022-02-21 20:13:30] local.DEBUG: The third job throws an exception.
[2022-02-21 20:13:30] local.ERROR: Oh no! {"exception":"[object] (Exception(code: 0): Oh no! at /home/marvin/projects/laravel-horizon-memory-leak/app/Jobs/ThirdJob.php:22)"}
[2022-02-21 20:13:30] local.DEBUG: The second job is started.
[2022-02-21 20:13:30] local.DEBUG: The third job throws an exception.
[2022-02-21 20:13:30] local.ERROR: Oh no! {"exception":"[object] (Exception(code: 0): Oh no! at /home/marvin/projects/laravel-horizon-memory-leak/app/Jobs/ThirdJob.php:22)"}
[2022-02-21 20:13:33] local.DEBUG: The first job is run.
[2022-02-21 20:13:33] local.DEBUG: Stopwatch {"timers":2}
[2022-02-21 20:13:33] local.DEBUG: The second job is started.
[2022-02-21 20:13:33] local.DEBUG: The third job throws an exception.
[2022-02-21 20:13:33] local.ERROR: Oh no! {"exception":"[object] (Exception(code: 0): Oh no! at /home/marvin/projects/laravel-horizon-memory-leak/app/Jobs/ThirdJob.php:22)"}
[2022-02-21 20:13:33] local.DEBUG: The first job is run.
[2022-02-21 20:13:33] local.DEBUG: Stopwatch {"timers":3}

The demo logs the `$timers` using `Queue::after()`, so it is understandable why there is no log after exceptions. But the `$timers` may also not increase in that case because 1) we cannot know the job is being run on the same worker again (which would potentially remove the timer) and 2) the job might be failed entirely.

Namoshek:
The safest and also simplest solution might be to just use a single `$timer` which gets overwritten at the start of new jobs. If backwards compatibility is not a thing (IMO it is an internal class and there is no useful way to inherit from it... so why not).

PR #1133, opened 2022-03-16
change chart.js intersect mode to false

swist666:
Hovering over the right place on the chart to display a tooltip (in the metrics section) is quite difficult. The dot is small so I changed the intersection mode to false (chart.js docs - https://www.chartjs.org/docs/2.9.4/configuration/tooltip.html).

The tooltip will allways be displayed to the nearest point (when hovering over chart).

### DIFF
--- a/resources/js/components/LineChart.vue
+++ b/resources/js/components/LineChart.vue
@@ -17,6 +17,9 @@
             this.chart = new Chart(this.context, {
                 type: 'line',
                 options: {
+                    tooltips: {
+                        intersect: false,
+                    },
                     legend: {
                         display: false,
                     },


taylorotwell:
Can you provide before / after screenshots and mark this PR as ready for review when you have done so? Thanks.

swist666:
Before:
https://user-images.githubusercontent.com/4541261/158612511-bdd56ed6-ebfb-48d9-a404-f83c798dca24.mov

After:
https://user-images.githubusercontent.com/4541261/158612562-b6c529c2-8c57-4f87-9a59-28b03b66dfdd.mov

PR #1141, opened 2022-04-07
[5.x] Revert "Make horizon work with strict content-security-policy"

driesvints:
Reverts laravel/horizon#1138


francoism90:
@driesvints Why?

PR #1154, opened 2022-05-30
Automatically compile and commit assets

driesvints:
Implements the reusable workflow from https://github.com/laravel/.github/pull/5.

As a bonus, we can remove the special release notes for this repo as the steps are now done automatically 🎉


driesvints:
Ah that has nothing to do with this. You'll still need to do that.

Issue #1171, opened 2022-08-04
LongWait notification specifies exact queue


jasonmccreary:
When specifying multiple queues for a process, the `LongWait` notifications must be configured using the queue names concatenated with a comma (`,`). While it made sense after the fact, my initial inclination was to specify the individual queue inline with the example (e.g. `connection:queue`).

More importantly, when receiving the notification, the queue uses the concatenated name. This doesn't immediately let me know which queue has the long wait. Ideally, the notification would specify the individual queue which has the long wait.

**Current behavior**

Example configuration:
```php
    'waits' => [
        // ...
        'redis:automation,subscriber,paid,workbench' => 1800,
    ],

    // ...

    'environments' => [
        'worker' => [
            'shifts' => [
                // ...
                'queue' => ['automation', 'subscriber', 'paid', 'workbench'],
                'balance' => false,
            ],
        ],
    ],
```

The notification states:

> The "automation,subscriber,paid,workbench" queue on the "redis" connection has a wait time of 62 seconds.

**Desired behavior**

Ideally you could configure the `waits` by the individual queue:
```php
    'waits' => [
        // ...
        'redis:automation' => 1800,
    ],

    // ...

    'environments' => [
        'worker' => [
            'shifts' => [
                // ...
                'queue' => ['automation', 'subscriber', 'paid', 'workbench'],
                'balance' => false,
            ],
        ],
    ],
```

And the notification would specify the wait time for the specific queue:

> The "automation" queue on the "redis" connection has a wait time of 1803 seconds.

---

If this is an acceptable change, I'm glad to help work on it as I already dug around the code a bit to better understand how to configure the `waits`.

taylorotwell:
Hey @jasonmccreary - sure feel free to send something over for us to review if you have time.

PR #1174, opened 2022-08-11
Monitor specific queue for long wait if configured

jasonmccreary:
This is a proposed solution for #1171 to monitor a specific queue if configured.

Currently, when running multiple queues an event is emitted for that "shared" queue. So there's no way to be notified when a specific queue within a "shared" queue has a long wait.

This PR allows you to receive an event for a specific queue, even if running within a multiple queue, when configured via `waits`.

For example:

```php
    'waits' => [
        // ...
        'redis:automation' => 1800,
    ],

    // ...

    'environments' => [
        'worker' => [
            'shifts' => [
                // ...
                'queue' => ['automation', 'subscriber', 'paid', 'workbench'],
                'balance' => false,
            ],
        ],
    ],
```

In an effort to be backwards compatible, I did not change the underlying `WaitTimeCalculator`. Instead, I isolated the change to the listener. It now also monitors any queues configured via `waits`. This actually provides more flexibility as you may now monitor individual and multiple queues, but potentially combinations of queues which may not run under the same supervisor (would need more testing).

Combined with #1172, I hope this makes the `waits` configuration more robust for developers needs.

### DIFF
--- a/src/Listeners/MonitorWaitTimes.php
+++ b/src/Listeners/MonitorWaitTimes.php
@@ -46,10 +46,20 @@ class MonitorWaitTimes
             return;
         }
 
-        // Here we will calculate the wait time in seconds for each of the queues that
-        // the application is working. Then, we will filter the results to find the
-        // queues with the longest wait times and raise events for each of these.
-        $results = app(WaitTimeCalculator::class)->calculate();
+        // Here we will calculate the wait time in seconds for each of the queues
+        // the application is working or monitoring. Then we will compare the
+        // results against the configured times for any long wait times.
+        $waitTime = app(WaitTimeCalculator::class);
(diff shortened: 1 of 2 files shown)


jasonmccreary:
Please help me with my cascading comment and collection pipeline. 😅

jasonmccreary:
@taylorotwell, yes, that may be correct. I was trying to avoid changing the `WaitTimeCalculator` for fear of a breaking change.

In the end, this desired feature is really know which queue, in a multiple queue, might have a long wait time. I may need to add a new method to determine that.

jasonmccreary:
After toying around a bit, there isn't a way to determine the wait time for a specific queue in a shared queue or multiple supervisor configuration.

As such, I am closing this PR.

PR #1175, opened 2022-08-11
[5.x] Added localization

andrey-helldar:
Added the ability to localize the interface.

For example, for French:

image

image


andrey-helldar:
Sorry, I can't fix codestyle errors because I don't have access to the site ☹️

image

andrey-helldar:
VPN helped solve the problem. It was only necessary to find the right country.

taylorotwell:

andrey-helldar:
@taylorotwell, in fact, it would be cool to use all parts of the multilingual application, including Horizon, but without this functionality, this step is impossible, unfortunately.

If you are confused by writing code, then it is made by analogy with Nova, and the mixin file is directly taken from it.

darakanoit:
Please consider including this in the package

arman-arif:
It would be great if this feature is implemented!

PR #1184, opened 2022-09-17
Add home page link to the horizon dashboard

roksprogar:
I've added a homepage link to the horizon dashboard, because after accessing horizon from the website (admin) menu, I think it allows for an easy option to get back to the website (and access the menus again).

### DIFF
--- a/resources/views/layout.blade.php
+++ b/resources/views/layout.blade.php
@@ -29,7 +29,7 @@
             </svg>
 
             <h4 class="mb-0 ml-2">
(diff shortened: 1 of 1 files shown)


ankurk91:
Don't use `{!!` in blade unless you really want to, it may cause XSS issues

taylorotwell:

roksprogar:
> Don't use `{!!` in blade unless you really want to, it may cause XSS issues

You think it could be a problem since it only reads from the config files? Any alternative solutions?

PR #1187, opened 2022-09-22
Update .gitattributes

ankurk91:
Hi,

No one is using these files from `vendor`

```bash
680K    vendor/laravel/horizon/src
764K    vendor/laravel/horizon/package-lock.json

```


driesvints:
`src` is definitely used?

ankurk91:
Whoops, sending another PR with correction.

PR #1190, opened 2022-09-26
Add exit codes to status command

mbardelmeijer:
This PR adds exit codes to the `horizon:status` artisan command.

The use-case for this is to monitor the Horizon status via monitoring tools, which we run on top of the normal supervisor configuration. 

We've noticed in production usage that supervisor didn't restart the Horizon worker in a specific edge-case where Redis crashed. The status command however gave the correct status of `inactive`. We wish to monitor the `horizon:status` output to ensure the queue is always running. 

Our current workaround is checking the output string, but we would prefer to use exit code checking.

This PR also adds colored output to the status:
- `error` for inactive (red)
- `warn` for paused (orange)
- `info` for running (green)

### DIFF
--- a/src/Console/StatusCommand.php
+++ b/src/Console/StatusCommand.php
@@ -25,27 +25,26 @@ class StatusCommand extends Command
      * Execute the console command.
      *
      * @param  \Laravel\Horizon\Contracts\MasterSupervisorRepository  $masterSupervisorRepository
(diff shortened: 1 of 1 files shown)


mbardelmeijer:
Sent again via personal fork via https://github.com/laravel/horizon/pull/1191

PR #1191, opened 2022-09-26
Add exit codes to status command

mbardelmeijer:
This PR adds exit codes to the `horizon:status` artisan command.

The use-case for this is to monitor the Horizon status via monitoring tools, which we run on top of the normal supervisor configuration. 

We've noticed in production usage that supervisor didn't restart the Horizon worker in a specific edge-case where Redis crashed. The status command however gave the correct status of `inactive`. We wish to monitor the `horizon:status` output to ensure the queue is always running. 

Our current workaround is checking the output string, but we would prefer to use exit code checking.

I was unsure which exit code to assign to the `paused` status. I've landed on exit code 1. Happy to receive input on that.

This PR also adds colored output to the status:
- `error` for inactive (red) - exit code 1
- `warn` for paused (orange) - exit code 1
- `info` for running (green) - exit code 0

### DIFF
--- a/src/Console/StatusCommand.php
+++ b/src/Console/StatusCommand.php
@@ -25,27 +25,26 @@ class StatusCommand extends Command
      * Execute the console command.
      *
      * @param  \Laravel\Horizon\Contracts\MasterSupervisorRepository  $masterSupervisorRepository
-     * @return void
+     * @return int
      */
     public function handle(MasterSupervisorRepository $masterSupervisorRepository)
     {
-        $this->line($this->currentStatus($masterSupervisorRepository));
-    }
-
-    /**
-     * Get the current status of Horizon.
-     *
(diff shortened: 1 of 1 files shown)


PHLAK:
@mbardelmeijer I know this has already been merged but I was wondering what your reasoning was for using an exit code of `1` for warn/paused instead of a different code (e.g. `2`)? I would think a different exit code would be useful for determining the difference between inactive/paused while still being able to do a "not equal to `0`" check for checking if horizon was not running regardless of condition.

I don't have an immediate use for this or anything but was curious if this could be improved.

mbardelmeijer:
@PHLAK sure! I was unsure what exit code exactly to use there too. 

I argued that exit code 2 is mostly used for a more severe error in bash scripts (generally). As the status is `paused`, I would argue that exit code 1 would still be the best fit. Happy to revisit though.

PR #1202, opened 2022-10-27
Link to Horizon docs from dashboard

shalvah:
Small quality-of-life improvement: link to the Horizon docs from the dashboard.

image

I often need to consult the docs to verify how something works or should be configured. This makes that use case simpler.


taylorotwell:
No plans to add this right now.

PR #1204, opened 2022-10-31
Add suggestion for laravel-horizon-watcher package from Spatie

phpfour:
This adds a composer suggestion for the https://github.com/spatie/laravel-horizon-watcher package from Spatie. 

This handy package can watch for local file changes and automatically restart horizon so that the new changes take effect.


taylorotwell:
We don't usually recommend "nice to have" packages here. Only things that affect core functionality.

Issue #1153, opened 2022-05-25
Exclude certain jobs from Completed Jobs list


hailwood:
Right now Horizon displays all jobs that have completed in the `Completed Jobs` list.
This is great, unless you've got jobs running that you don't really care about - 

For example we use OhDearApp so if the job fails we'll know about it from OhDearApp notifying us.
The issue we run into is that it fires one job per scheduled task per execution - and this is drowning out the less often jobs we *do* care about.

It would be great if either in the horizon config we could have a way to say 'hide these jobs'
```php
'hide_completed_jobs' => [
    PingOhDearJob::class,
],
```
Or a closure based approach like (complete pseudocode)
```php
Horizon::filterJob(function(Job $job) {
   return $job->className !== PingOhDearJob::class || !$job->isComplete;
});
```
![horizon_jobs_completed](https://user-images.githubusercontent.com/709773/170377582-4f8b3c77-8c00-4cf4-9810-ed0e8f38fb8f.png)

driesvints:
I like the idea here. Maybe the following config?

```php
'jobs' => [
    'exclude' => [
        PingOhDearJob::class,
    ],
],
```

I'll ping @taylorotwell to see if we can accept a PR for this.

taylorotwell:
I'm fine with this if someone wants to make a PR. 👍

driesvints:
Feel free to PR something 👍

driesvints:
@hailwood I've started working on something here: https://github.com/laravel/horizon/pull/1230

PR #1230, opened 2022-12-20
[5.x] Exclude jobs from completed overview

driesvints:
Closes https://github.com/laravel/horizon/issues/1153

This PR implements a way to exclude certain jobs from the Horizon completed overview. This is handy to filter frequently shown jobs like `PingOhDearJob`.

```php
'jobs' => [
    'exclude' => [
        PingOhDearJob::class,
    ],
],
```


driesvints:
I managed to crack the changes needed to `getJobsByType` to properly filter for jobs but now I'm running into an issue with pagination and `countJobsByType`. Because it returns the count of *all* jobs, not just the ones filtered by job name, pagination allows you to continue to click to next pages because the total pages count doesn't matches the result set.

We could maybe change `countJobsByType` to the same inner workings as `getJobsByType` to retrieve all ids (without mapping them to jobs) to do a full count in PHP of all matching ID's. But this could maybe add a lot of load if there's tons and tons of results.

I'm not sure what the correct way forward is now. If this were a database result set we could just simply do a `whereIn` 😅

taylorotwell:
@driesvints we may be approaching this from the wrong angle entirely which is making our life harder than it needs to be.

Another option to achieve the same results would be to either:

A) Not store excluded jobs that complete at all. See the `completed` method of `RedisJobRepository`. This is definitely the easiest approach; however, people may still want some way view information about them, which brings me to:

B) Store excluded jobs in a new category of jobs called "Silenced" or something, with a new tab in the UI. This would also likely be a pretty easy change by modifying the `completed` method and adding a couple of new methods on the `RedisJobRepository`.

driesvints:
@taylorotwell great suggestion! I sent this in: https://github.com/laravel/horizon/pull/1232

PR #1236, opened 2023-01-16
[5.x] Implement Silenced interface

driesvints:
This PR implements a new `Silenced` contract that, once applied to a job, will automatically silence it and move it to the Silenced tab. This is handy for packages so they can silence their jobs by default.

### DIFF
--- a/src/Contracts/Silenced.php
+++ b/src/Contracts/Silenced.php
@@ -1,0 +1,7 @@
+<?php
+
+namespace Laravel\Horizon\Contracts;
+
+interface Silenced
+{
+}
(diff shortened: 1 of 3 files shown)


freekmurze:
Instead of an interface, it could also be a property or function that controls this behaviour. That way package users could override the package default (you can't override an interface).

lioneaglesolutions:
@freekmurze @driesvints I just added https://github.com/laravel/horizon/pull/1238 - no idea if it's close enough but tests pass... happy to take feedback or close it if it isn't where it needs to be for the method override. I would assume you could do the same thing with a public property easily enough.

PR #1237, opened 2023-01-17
[5.x] Increase `perPage` variable for monitored tags

joshuaruesweg:
Currently, all pages show 50 jobs per page. Only the Monitored Tags page is limited to 3 jobs per page. In my opinion, this makes little sense and especially if you want to monitor a lot more jobs, it is very difficult to find the right jobs or to get an overview of the last jobs that have been run.


taylorotwell:
Looks like left over debug code.

PR #1238, opened 2023-01-20
[5.x] Add silence method override

lioneaglesolutions:
Hacked a little bit on this to get some feedback on being able to silence a job using a method.

Was mentioned in https://github.com/laravel/horizon/pull/1236

Wasn't sure about the best way to test this but was hoping this was close enough - but I don't have the biggest hope it's correct as Dries said we need to use reflection so maybe I"m missing something by storing the silenced result in the payload.

This PR enabled you to silence a job using a method instead of an interface or the config file;
```php
class SilencedJob implements ShouldQueue
{
    public function silenced()
    {
        return true;
    }
}
```

I wasn't sure if the method name should be `silenced` or `silence` or if the method should override the interface or config. At the moment if **any** of the Interface is there, class is in config or silenced method returns true, then the job will be silenced.

Happy to PR the docs if this is merged.

### DIFF
--- a/src/JobPayload.php
+++ b/src/JobPayload.php
@@ -79,6 +79,16 @@ class JobPayload implements ArrayAccess
     }
 
     /**
+     * Determine if the job is silenced.
+     *
+     * @return bool
+     */
+    public function isSilenced()
+    {
+        return $this->decoded['silenced'] ?? false;
+    }
+
+    /**
      * Prepare the payload for storage on the queue by adding tags, etc.
      *
      * @param  mixed  $job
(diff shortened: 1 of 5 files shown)


driesvints:
This lgtm! Although I'm not quite sure we actually need such a way to modify the behavior.

lioneaglesolutions:
> This lgtm! Although I'm not quite sure we actually need such a way to modify the behavior.

The only benefit I can think of is being able to conditionally silence a job (if that would be every necessary). Perhaps you could choose not to silence it under certain conditions (gaining insights a certain tenant's jobs for example).

driesvints (approved):

taylorotwell:
Not necessary... you can tag jobs and monitor them even if they are silenced... effectively giving you a way to surface jobs for a given tenant exactly as your use case describes.

CleanShot 2023-01-20 at 10 13 16@2x

PR #1244, opened 2023-02-07
[5.x] Fix missing variable use in anonymous function

joshuaruesweg:


ankurk91:
The bug must have been caught by a test case

Issue #1245, opened 2023-02-09
Lodash library special characters are breaking horizon dashboard


wesleyvieira12:
- Horizon Version: 5.14.1
- Laravel Version: 9.36.2
- PHP Version: 8.0.2
- Redis Driver & Version: predis 2.1
- Database Driver & Version: Not

**Description:**
I followed the correct steps to install horizon, but the browser was showing the error: "Uncaught SyntaxError: Invalid or unexpected token (at app.js?id=766b568d1924aa83b2a1f279d2c5ab9e:2:484352)".

We analyzed the app.js file and saw that it had the following error code:

[...]

PR #1246, opened 2023-02-10
[5.x] Remove Lodash

jessarcher:
The original intention of this PR was to change the Lodash imports to just pull in the functions we're actually using in order to avoid importing the unused function causing #1245 and to reduce the bundle size.

It then became apparent that we didn't actually need Lodash in most places, so I removed it altogether. This shaved 70KB from the _minified_ bundle.


browner12:
thanks for taking care of this! we were happy to see the bundle savings in our projects, as well.

want to handle it for Telescope too?

https://github.com/laravel/telescope/blob/4.x/package.json

PR #1255, opened 2023-03-06
[5.x] Add cancelled status to batch overview

markvaneijk:
Currently it's not visible if a batch is cancelled and instead it still shows `Pending` even when the batch is cancelled and not running anymore.

I found this quite confusing when cancelling batches for a specific use case where only one batch may run at once and the previous needs to be cancelled when there is already one pending.

So this adds the functionality to show the status of cancelled batches:

CleanShot 2023-03-06 at 16 28 55@2x

### DIFF
--- a/resources/js/screens/batches/index.vue
+++ b/resources/js/screens/batches/index.vue
@@ -171,14 +171,17 @@
                         </router-link>
                     </td>
                     <td>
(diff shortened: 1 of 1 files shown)


markvaneijk:
Choice for badge could also be `bade-warning`.

PR #1268, opened 2023-04-25
[5.x] add staging environment in config

ankurk91:
Hi,

I believe all of us has a `staging` environment.

If someone forgets to add the environment in `config/horizon.php`, their Horizon dashboard will show status as "Inactive".

There is no easy way to us to know what is wrong, specially for a beginner.

### DIFF
--- a/config/horizon.php
+++ b/config/horizon.php
@@ -204,6 +204,12 @@ return [
             ],
         ],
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
You can add this to your own config if necessary.

PR #1272, opened 2023-05-05
Add timestamp to horizon start message, like other queue messages

rudiedirkx:
All queue (work) messages have timestamps, but not the Horizon start message:

> Horizon started successfully.

Could be useful too!:

> 2023-05-05 22:51:10 Horizon started successfully.

It doesn't have to be flashier. Works for every Laravel.


taylorotwell:

PR #1275, opened 2023-05-18
Respect the prefers-color-scheme setting

ramonrietdijk:
I've changed the layout file to include media attributes on the CSS files in order to respect the `prefers-color-scheme`-setting. The style of Horizon will therefore be determined by the preference of the user. Developers who prefer Horizon in light mode are not affected by the `Horizon::night()` method anymore.

Note that the `night` method has not been removed to prevent a breaking change.


driesvints:
I think this is a good step but maybe we should take it further and add a toggle like on the Laravel docs so people can choose either light, dark or system preferences?

ramonrietdijk:
Agreed, it would offer a lot more flexibility

driesvints:
Can you add a screenshot how this looks?

ramonrietdijk:
This is how it currently looks:
image

driesvints:
I'd personally remove the background but other than that lgtm. Thanks!

driesvints (approved):

taylorotwell:
LGTM - thanks 👍

lloricode:
Maybe adding `@deprecated` to a `night()` will will stop using this in future?

PR #1284, opened 2023-06-13
Add event for when supervisor process died

PrinsFrank:
Following #1282, we are still figuring out why our supervisors are disappearing. As they are not terminated but die, this PR adds events so third-party code can process these occurrences.

### DIFF
--- a/src/Events/SupervisorProcessDied.php
+++ b/src/Events/SupervisorProcessDied.php
@@ -1,0 +1,37 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Laravel\Horizon\Events;
+
+use Laravel\Horizon\SupervisorProcess;
+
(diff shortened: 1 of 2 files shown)


src/Events/SupervisorProcessDied.php
```diff
@@ -0,0 +1,36 @@
+<?php
+declare(strict_types=1);
```
driesvints, line 2: Remove this please. We don't use it in any Laravel code.

PrinsFrank: Sorry, it's in my file template in my IDE. Removed!

taylorotwell:
Since the last PR didn't help you I don't want to randomly add code to Horizon to try to debug this issue. Try to add this to your own copy of Horizon first and see if it solves your issue.

PR #1289, opened 2023-06-22
Check that minProcesses times the nr of queues is less than maxProcesses

PrinsFrank:
When changing the number for minProcesses, maxProcesses or the amount of queues for a supervisor, there might be 0 processes assigned for all queues within a supervisor. This happens when minProcesses * count(queues) is more than maxProcesses:
```php
'environments' => [
    'production' => [
        'supervisor-1' => [
            'minProcesses' => 2,
            'maxProcesses' => 5,
            'queues' => ['queue-1', 'queue-2', 'queue-3']
        ]
    ]
],
```

This PR adds an exception to prevent this from silently happening.

If maxProcesses is less than the number of queues, the maxProcesses is ignored and a process is added for each queue silently. This will still work after this change, but might be unexpected as well.

### DIFF
--- a/src/ProvisioningPlan.php
+++ b/src/ProvisioningPlan.php
@@ -182,6 +182,10 @@ class ProvisioningPlan
             throw new Exception("The value of [{$supervisor}.minProcesses] must be greater than 0.");
         }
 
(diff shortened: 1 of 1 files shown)


PrinsFrank:
Sorry, this is a breaking change and currently only results in supervisors with zero processes but horizon still starts. I'll fix the underlying issue in the autoscaler.

PR #1294, opened 2023-07-12
Fix: Make sure that missing supervisors are correctly displayed when using multiple environments

PrinsFrank:
When displaying active/inactive supervisors for a master supervisor that is started with a different environment, the provisioning in the overview is done with the configuration of the environment the horizon endpoint runs on, causing a potential mismatch when using different environments. This PR passes the environment to the master and uses that variable to display a correct overview. (See discussion in PR #1286 for a reproduction)

### DIFF
--- a/src/MasterSupervisor.php
+++ b/src/MasterSupervisor.php
@@ -22,6 +22,13 @@ class MasterSupervisor implements Pausable, Restartable, Terminable
     use ListensForSignals;
 
     /**
+     * The environment that was used to provision this master supervisor.
+     *
+     * @var string|null
+     */
+    public $environment;
+
+    /**
      * The name of the master supervisor.
      *
      * @var string
@@ -57,22 +64,17 @@ class MasterSupervisor implements Pausable, Restartable, Terminable
     public static $nameResolver;
 
     /**
-     * The environment that is used to provision this master supervisor.
-     *
-     * @var string|null
-     */
-    public $environment;
-
-    /**
      * Create a new master supervisor instance.
      *
+     * @param  string  $environment
      * @return void
      */
     public function __construct(string $environment = null)
     {
+        $this->environment = $environment;
+
         $this->name = static::name();
         $this->supervisors = collect();
-        $this->environment = $environment;
 
         $this->output = function () {
             //
--- a/src/MasterSupervisor.php
+++ b/src/MasterSupervisor.php
@@ -57,14 +57,22 @@ class MasterSupervisor implements Pausable, Restartable, Terminable
     public static $nameResolver;
 
     /**
+     * The environment that is used to provision this master supervisor.
+     *
+     * @var string|null
+     */
+    public $environment;
+
+    /**
      * Create a new master supervisor instance.
      *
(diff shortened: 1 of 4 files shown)


src/MasterSupervisor.php
```diff
@@ -56,15 +56,23 @@ class MasterSupervisor implements Pausable, Restartable, Terminable
...
+     */
+    public $environment;
+
     /**
      * Create a new master supervisor instance.
      *
      * @return void
      */
-    public function __construct()
+    public function __construct(?string $environment = null)
```
driesvints, line 71: The `?` is not needed here because the argument is already nullable.

```suggestion
    public function __construct(string $environment = null)
```

driesvints (changes requested):

src/Repositories/RedisMasterSupervisorRepository.php
```diff
@@ -81,9 +81,10 @@ public function get(array $names)
 
             return ! $record[0] ? null : (object) [
                 'name' => $record[0],
-                'pid' => $record[1],
-                'status' => $record[2],
-                'supervisors' => json_decode($record[3], true),
+                'environment' => $record[1],
+                'pid' => $record[2],
+                'status' => $record[3],
+                'supervisors' => json_decode($record[4], true),
```
driesvints, line 87: Does this not break existing Horizon installs who have the records saved in a different sequence?

PrinsFrank: I checked this scenario specifically, where I started a mastersupervisor and then switched to this branch. HMGET returns the keys as requested in the order at line 75. [For any keys that don't exist, null is returned](https://redis.io/commands/hmget/). As there might be master supervisors already running with missing environments, the key might return null which will be properly handled as the fallback environment will be used in the masterSuperVisorController.

driesvints: I mean, doesn't the order of the keys and values get switched? I think the mapping of these values will be wrong?

PrinsFrank: I can put the environment variable at the end as well, but according to the [redis docs](https://redis.io/commands/hmget/) it doesn't matter:
> **Return** [Array reply](https://redis.io/docs/reference/protocol-spec#resp-arrays): list of values associated with the given fields, in the same order as they are requested.

> For every field that does not exist in the hash, a nil value is returned.

driesvints: I think I see what you mean. Seems good to me!

taylorotwell:
Why would the environments be different?

PrinsFrank:
> Why would the environments be different?

In the linked PR @bjarn had an example where there were multiple environments with different supervisor configurations. 

I've seen this setup at more larger companies, and it provides another dimension of scalability. For example, you can use 3 default configured pods running the default production supervisors, but then when there is a workload on a specific queue that is suffocating the other queues, you can add a dedicated pod for a different environment running only that queue.

PR #1301, opened 2023-08-14
Added silenced expiration

MartiSalaMorral:
This feature allows users to configure an expiration time for silenced jobs. This allows having different expiration times between completed and silenced jobs.

If silenced expiration is not present, it uses the completed expiration time by default.

### DIFF
--- a/config/horizon.php
+++ b/config/horizon.php
@@ -102,6 +102,7 @@ return [
         'recent' => 60,
         'pending' => 60,
         'completed' => 60,
+        'silenced' => 60,
         'recent_failed' => 10080,
         'failed' => 10080,
         'monitored' => 10080,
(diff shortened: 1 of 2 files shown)


MartiSalaMorral:
Hi @taylorotwell, thanks for your answer.

Although this code does not correct a bug, it adds a quality of life improvement that I think is quite interesting.

This allows to set an expire time for the silenced jobs without affecting the completed jobs. I made these changes because the importance of tracking silenced and completed jobs can differ a lot and setting a lower time for the silenced jobs helps to reduce the memory usage.

mgralikowskifrontkom:
I was actually thinking about this change now. Since silenced jobs are typically less critical or significant, providing this flexibility is especially valuable.

mgralikowski:
I was actually thinking about this change now. Since silenced jobs are typically less critical or significant, providing this flexibility is especially valuable.

PR #1308, opened 2023-08-29
[5.x] Throws `Laravel\Horizon\Exceptions\ForbiddenException` on unauthorized access

joelbutcher:
It is common in applications that restrict access to another users' resource to return a 404 to indicate that resource does not exist. I would like to give developers to option to apply this functionality to Horizon to hide it's presence within an application from unauthenticated users.

This PR allows developers to define what status code should be returned when Horizon's authentication fails, via a new `horizon.unauthorized_status` config option (happy for this to be renamed, if anyone has any other suggestions).

I've restricted the accepted status codes to 403 and 404, falling back to 403 if this is not the case.

### DIFF
--- a/src/Exceptions/UnauthorizedException.php
+++ b/src/Exceptions/UnauthorizedException.php
@@ -1,0 +1,10 @@
+<?php
+
+namespace Laravel\Horizon\Exceptions;
+
+use RuntimeException;
+
+class UnauthorizedException extends RuntimeException
+{
+    //
+}
--- a/src/Exceptions/UnauthorizedException.php
+++ b/src/Exceptions/UnauthorizedException.php
@@ -2,9 +2,9 @@
 
 namespace Laravel\Horizon\Exceptions;
 
-use RuntimeException;
+use Symfony\Component\HttpKernel\Exception\HttpException;
 
-class UnauthorizedException extends RuntimeException
+class UnauthorizedException extends HttpException
 {
     //
 }
--- a/src/Http/Middleware/Authenticate.php
+++ b/src/Http/Middleware/Authenticate.php
@@ -15,6 +15,26 @@ class Authenticate
      */
     public function handle($request, $next)
     {
-        return Horizon::check($request) ? $next($request) : abort(403);
+        if (! Horizon::check($request)) {
+            abort($this->statusCode());
+        }
+
+        return $next($request);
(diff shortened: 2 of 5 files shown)


joelbutcher:
@crynobone a custom exception would work – are you thinking to extend a `RuntimeException` or something else?

taylorotwell:
I think now we would throw a 500 error?

src/Exceptions/UnauthorizedException.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace Laravel\Horizon\Exceptions;
+
+use RuntimeException;
+
+class UnauthorizedException extends RuntimeException
```
crynobone, line 7: Should extends https://github.com/symfony/http-kernel/blob/6.3/Exception/UnauthorizedHttpException.php

You should be able to use renderable and typehint directly to Horizon exception to intercept the response.

crynobone (changes requested):

src/Exceptions/UnauthorizedException.php
crynobone: as @taylorotwell  mentioned, without it now the exception would render 500 error code.

joelbutcher: @crynobone that exception requires a challenge for the `WWW-Authenticate` header, which I'm assuming we don't want / need to set? I've extended the base `HttpException` and updated the status code to 401

src/Http/Middleware/Authenticate.php
```diff
@@ -15,6 +16,10 @@ class Authenticate
      */
     public function handle($request, $next)
     {
-        return Horizon::check($request) ? $next($request) : abort(403);
+        if (! Horizon::check($request)) {
+            throw new UnauthorizedException(401);
```
crynobone, line 20: Any reason not to use `403` same as before?

crynobone: I probably pasted the incorrect exception on my last review. Sorry

joelbutcher: Well, the 403 status code !== unauthorized. We can use 403, but we should probably change the name of the exception to use "Forbidden" instead :)

crynobone (approved):

PR #1309, opened 2023-08-31
Suppress status output option

tomcastleman:


tomcastleman:
Created PR here by mistake. Will follow up with one when ready

PR #1313, opened 2023-09-06
Callback function for notifications

ercsctt:
As of right now, there's pretty limited customization for Horizon notifications, without overriding a bunch of classes. I've introduced a change that allows for a Closure to be defined in the `HorizonServiceProvider`, effectively allowing you to send notifications anywhere.

An example of this is LogSnag, which I use pretty heavily for monitoring and alerts on my application. This is how that would look:

```php
class HorizonServiceProvider extends HorizonApplicationServiceProvider
{
    public function boot(): void
    {
        ...

        Horizon::routeGenericNotificationsTo(function ($notification, $message) {
            Logsnag::log(
                channel: 'horizon',
                event: 'Queue wait times alert',
                description: $message,
                icon: '🚨',
                notify: true
            );
        });
    }
}
``` 

I've also added some tests to verify notifications are being sent, as it seemed to not cover this.


taylorotwell:
Hmm - we only have one notification (for long waits). Can you not just listen for the `LongWaitDetected` event and do whatever you want?

Issue #1063, opened 2021-08-25
Pending job count not correct on batch with failures allowed


bilfeldt:
- Horizon Version: 5.7.11
- Laravel Version: 8.56.0
- PHP Version: 7.4
- Redis Driver & Version: predis/phpredis: Unknown
- Database Driver & Version: MySQL 8.0

### Description:

When using batches with [failures allowed](https://laravel.com/docs/8.x/queues#allowing-failures) then the Batch information shown in Horizon can be quite confusing/buggy.

Example (can be found in github repo below):
```php
\Illuminate\Support\Facades\Bus::batch([
        new \App\Jobs\TestJob(1), // Will succeed on first attempt
        new \App\Jobs\TestJob(2), // Will succeed on second attempt
        new \App\Jobs\TestJob(3), // Will succeed on third attempt
        new \App\Jobs\TestJob(4), // Will succeed on fourth attempt (which will never happen since $maxExceptions = 3)
        new \App\Jobs\TestJob(0), // Will never succeed
    ])->allowFailures()->dispatch();
```

This batch should have a total of 5 jobs and 3 of them should eventually succeed (60%). Now the end result in Horizon shows 2 pending jobs + 2 failed jobs:

![Screenshot 2021-08-25 at 14 58 05](https://user-images.githubusercontent.com/30228807/130794722-38ee3ce7-5638-41a0-8384-11a899ad8bba.png)

I think this is a bug since only jobs not marked as "failed" or "completed" should appear as pending (since they will be processed soon). This is also confirmed by no pending jobs appearing:

![Screenshot 2021-08-25 at 14 58 08](https://user-images.githubusercontent.com/30228807/130794807-03265be0-678d-4cea-99a6-676c78443498.png)

While there are in fact three completed jobs:

![Screenshot 2021-08-25 at 14 58 11](https://user-images.githubusercontent.com/30228807/130794849-2f4877d3-572e-42c2-9e70-8cfede31c43f.png)

and two failed:

![Screenshot 2021-08-25 at 14 58 13](https://user-images.githubusercontent.com/30228807/130794876-35b7e716-f7cb-46f5-a997-c1e721864b3e.png)

### Steps To Reproduce:

A Github repository showcasing the error can be seen here: https://github.com/bilfeldt/bug-report

1. Start Horizon (remember to use redis as queue driver): `php artisan horizon`
2. Go to `/batch` which will dispatch the batch of jobs
3. Go to the horizon dashboard and see the images also attached above

PR #1315, opened 2023-09-12
Use available `batch.processedJobs` value from `Illuminate\Bus\Batch`

### LINKED ISSUE #1311: Pending job number seems incorrect
### Horizon Version

5.19.1

### Laravel Version

10.19.0

### PHP Version

8.2.9

### Redis Driver

Predis

### Redis Version

1.1.10

### Database Driver & Version

_No response_

### Description

It appears the "Pending Jobs" number may be incorrect or misleading when there are failed jobs. This batch is complete and 160 did fail. However, no more are pending.

Screenshot 2023-09-05 at 7 47 17 PM

I would expect this number to be 0 since the batch is complete. Or, if it is to indicate the number of "Incomplete Jobs", then maybe that should have a separate title. Depending on the direction, the "Completed Jobs" and its subtitle might need to be adjusted as currently I would expect it to be `484`.

I see the same thing for all my recent batched jobs with failures - the "Pending Jobs" and "Failed Jobs" numbers are the same.

### Steps To Reproduce

Run a batched job with some failures. You should see "Pending Jobs" and "Failed Jobs" are the same number.

crynobone:
fixes #1311


driesvints (approved):

taylorotwell:
I see this marks #1311 as complete but that issue is primarily about the "pending" job count?

crynobone:
@taylorotwell Based on the previous PR to "Completed Jobs" the value of "Pending" and "Failed" with `allowFailures()` does make sense as discussed in https://github.com/laravel/horizon/issues/1063#issuecomment-906753640

Issue #1314, opened 2023-09-07
Monitoring tags consumes more Redis RAM than expected


benholmen:
### Horizon Version

5.20.0

### Laravel Version

10.22.0

### PHP Version

8.2.9

### Redis Driver

PhpRedis

### Redis Version

Predis 2.2.1

### Description

We ran into out of memory issues with Redis and after investigation found that several zsets were unusally large:

```
Biggest   zset found 'hubventory_horizon:media-import' has 16922051 members
...
Biggest   zset found 'hubventory_horizon:media-import' has 2379151023 bytes
```

Those correspond to a `media-import` tag that we are monitoring in Horizon. The oldest score/timestamp in that zset was 7 months old, which is much older than we expected.

Our Horizon config includes the following `trim` settings:
```php
    'trim' => [
        'recent' => 60,
        'pending' => 60,
        'completed' => 60,
        'recent_failed' => 10080,
        'failed' => 10080,
        'monitored' => 1440,
    ],
```

I did some source diving and found [`RedisJobRepository@trimMonitoredJobs`](https://github.com/laravel/horizon/blob/bea664a3cd78eee9729ede94713b1f4edc98b738/src/Repositories/RedisJobRepository.php#L603-L604) that seems to intend to clean this up using `config('horizon.trim.monitored')` but it only works on the `monitored_jobs` key.

Is it up to us to trim these monitored jobs or is there a Horizon method we can use to keep them from growing over time and consuming so much RAM?

Thanks so much!

### Steps To Reproduce

1. Monitor a tag via the UI.
2. Complete many jobs with that tag.

benholmen:
Similar issues I found: [715](https://github.com/laravel/horizon/issues/715), [333](https://github.com/laravel/horizon/issues/333)

Related PR: [484](https://github.com/laravel/horizon/pull/484)

driesvints:
Hi @benholmen. Thanks for reporting this. Which exact key would we also target with this trim?

driesvints:
@benholmen have you had time to look at my question above?

benholmen:
@driesvints [PR submitted](https://github.com/laravel/horizon/pull/1317)!

benholmen:
For anyone with a similar issue, @taylorotwell has clarified that the intended use case for monitoring is short term, not permanent monitoring (see #1317).

If you do monitor a tag indefinitely, your Redis memory consumption will grow over time. You can manually trim those monitored tags using a redis command like the following:

```redis
[...]

PR #1317, opened 2023-09-14
Trim monitored tag sorted sets, not `monitored_jobs`

benholmen:
This PR addresses [issue 1314](https://github.com/laravel/horizon/issues/1314): for each monitored tag, the redis sorted set containing the timestamp and job ID would grow without trimming. `monitored_jobs` was trimmed, but seems unused so I removed it to avoid confusion.

### DIFF
--- a/src/Repositories/RedisJobRepository.php
+++ b/src/Repositories/RedisJobRepository.php
@@ -417,8 +417,6 @@ class RedisJobRepository implements JobRepository
     public function remember($connection, $queue, JobPayload $payload)
     {
         $this->connection()->pipeline(function ($pipe) use ($connection, $queue, $payload) {
-            $this->storeJobReference($pipe, 'monitored_jobs', $payload);
-
             $pipe->hmset(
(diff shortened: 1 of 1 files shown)


driesvints:
@benholmen seems the tests fail here. Can you fix those before proceeding?

benholmen:
Thanks @driesvints @taylorotwell!

I like being able to monitor recent jobs - we use it for troubleshooting recent product imports, webhook processing, etc, which is more effective if it's always available. If we turned it on as needed we'd be missing some context.

How do you want to proceed? Do you want me to add tests to this PR and keep it moving forward, or do you want to go a different direction?

driesvints:
Thanks @benholmen. Remember to mark a PR as ready for review if you want an answer from @taylorotwell.

benholmen:
Thanks @driesvints! I'll do that next time.

taylorotwell:
Personally I would rather not break existing use cases so going to table this for now.

Issue #1152, opened 2022-05-16
Obsolete Jobs/Queues in Metrics


Cellard:
- Horizon Version: 5.9.7
- Laravel Version: 9.12.2
- PHP Version: 8.1
- Redis Driver & Version: predis 1.1.10
- Database Driver & Version:

### Description:

Metrics is showing jobs and queues, that were already removed or renamed.

### Steps To Reproduce:

1. Create and run a job
2. Rename job class and run it again

1. Define a queue, run a job through it
2. Rename queue and run a job again

In both cases old names remained in Horizon Metics panel. I hoped they will be removed after `metrics.trim_snapshots`  exceeded, but they wasn't.

driesvints:
This is expected. The metrics show historical data. Renaming doesn't changes the data that was snapshotted.

Cellard:
Is there a way to completely reset metrics?

erickneverson:
I recognize this issue is closed, but am also curious about how to reset this particular metric.  Do we need to create some sort of custom command until one is provided by the framework?

Thanks

trevorgehman:
I'm also wondering the same thing. Is there some kind of command to just completely wipe the Horizon metrics that are stored in Redis?

trevorgehman:
Here's something I came up with to clear all Horizon metrics:

```php

use Laravel\Horizon\Repositories\RedisMetricsRepository;

$redis = resolve(RedisMetricsRepository::class)->connection();

$redis->del('measured_jobs');
$redis->del('measured_queues');
$redis->del('last_snapshot_at');

foreach($redis->keys('snapshot:*') as $key) {
  $redis->del(str($key)->after(config('horizon.prefix')));
}

foreach($redis->keys('metrics:*') as $key) {
  $redis->del(str($key)->after(config('horizon.prefix')));
}

trevorgehman:
> Here's something I came up with to clear all Horizon metrics:
> 
> ```
> use Laravel\Horizon\Repositories\RedisMetricsRepository;
> 
> $redis = resolve(RedisMetricsRepository::class)->connection();
> 
> $redis->del('measured_jobs');
> $redis->del('measured_queues');
> $redis->del('last_snapshot_at');
> 
> foreach($redis->keys('snapshot:*') as $key) {
>   $redis->del(str($key)->after(config('horizon.prefix')));
> }
> 
> foreach($redis->keys('metrics:*') as $key) {
>   $redis->del(str($key)->after(config('horizon.prefix')));
[...]

PR #1318, opened 2023-09-15
Add `horizon:reset-metrics` command

trevorgehman:
# Problem

If a Job is renamed or moved, it will persist in Horizon under Metrics and never be removed. This ends up cluttering the Metrics with obsolete jobs. The problem is described in https://github.com/laravel/horizon/issues/1152

CleanShot 2023-09-15 at 00 40 06@2x

# Solution

The simplest solution IMO is to just add a command to reset the metrics for Horizon entirely. I've given this command the signature `horizon:reset-metrics`.

It will remove all metric data in Horizon's Redis database, including:

- List of jobs and queues that have metrics: `measured_jobs` / `measured_queues`
- Current metrics (average runtime & throughput) for those jobs and queues: `job:*` / `queue:*`
- Snapshots of metrics: `snapshot:*`
- Timestamp of the last snapshot: `last_snapshot_at`
- Snapshot lock key: `metrics:snapshot`

I believe I have covered everything relating to metrics without touching anything else. 

Additionally, because I have to search Redis by wildcard to find some of the metrics data, I used SCAN instead of KEYS which is what Redis recommends to prevent locking issues: 

> Warning: consider KEYS as a command that should only be used in production environments with extreme care. It may ruin performance when it is executed against large databases. This command is intended for debugging and special operations, such as changing your keyspace layout. Don't use KEYS in your regular application code. If you're looking for a way to find keys in a subset of your keyspace, consider using [SCAN](https://redis.io/commands/scan)...

> https://redis.io/commands/keys/

# After Running Command

CleanShot 2023-09-15 at 01 01 09@2x

CleanShot 2023-09-15 at 01 02 39@2x

CleanShot 2023-09-15 at 01 02 42@2x

### DIFF
--- a/src/Console/ResetMetricsCommand.php
+++ b/src/Console/ResetMetricsCommand.php
@@ -1,0 +1,37 @@
+<?php
+
+namespace Laravel\Horizon\Console;
+
+use Illuminate\Console\Command;
+use Laravel\Horizon\Contracts\MetricsRepository;
+
+class ResetMetricsCommand extends Command
+{
+
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'horizon:reset-metrics';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Delete metrics for all jobs and queues';
+
+    /**
+     * Execute the console command.
+     *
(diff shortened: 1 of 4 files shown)


wit3:
if i try to use the command ``
i receive this error : `Call to undefined method Laravel\Horizon\Repositories\RedisMetricsRepository::reset()`

Maybe the command must call `::clear()` instead of `::reset()`

??

PR #1327, opened 2023-09-25
Base-encode job metrics slug to prevent slashes or backslashes pre-ur…

goestijn:
…l-decoders (e.g. azure appservices) can interpret as request_uri segment


taylorotwell:
Is it possible the slug would contain non-Latin characters which would fail on `btoa`?

driesvints:
Another thought: would existing urls still work?

goestijn:
The latest commit has replaced the depreciated `btoa` function in js with the `Buffer` class which does also supports non-Latin characters.

taylorotwell:
I don't know that I really want to mess with this on patch release.

PR #1332, opened 2023-10-06
feat: Adding Horizon Flush Command to delete all keys on horizon conn

jihadadulrazaqjasim:
This feature is useful in local environments to truncate all horizon-related information from Redis, It includes confirmation when running the command in production.

### DIFF
--- a/src/Console/FlushCommand.php
+++ b/src/Console/FlushCommand.php
@@ -3,10 +3,12 @@
 namespace Laravel\Horizon\Console;
 
 use Illuminate\Console\Command;
+use Illuminate\Console\ConfirmableTrait;
 use Illuminate\Contracts\Redis\Factory as RedisFactory;
 
 class FlushCommand extends Command
 {
+    use ConfirmableTrait;
     /**
      * The name and signature of the console command.
      *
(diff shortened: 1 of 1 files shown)


src/Console/FlushCommand.php
```diff
@@ -0,0 +1,37 @@
...
+   * @param RedisFactory $redis
+   * @return void
+   * @throws \RedisException
+   */
+    public function handle(RedisFactory $redis)
+    {
+      if(app()->environment('local')) {
+        $redis->connection('horizon')->client()->flushAll();
+        $this->info('All queue jobs flushed successfully.');
+      }
```
crynobone, line 35: Why do we need to check for `local` environment here?

crynobone: We have special trait to add confirmation whenever we need to run a command on `production` envrionment.

jihadadulrazaqjasim: @crynobone, I just added the confirmation trait. Do you think we should include a `--force` option? In my view, it's better not to include it, as the action is quite destructive, and requiring confirmation would be safer. What do you think?

taylorotwell:

PR #1333, opened 2023-10-09
Fix the job repository class based on service bindings

daison12006013:
Basically when extending the `RedisJobRepository.php` it doesn't work properly when calling `horizon:clear`
I've noticed that we're not using the one inside the `ServiceBindings.php`, the code currently calling the original `RedisJobRepository`

PS: In our project, I was able to make the redis-cluster worked, the only remaining issue is this horizon:clear

Extended the class
image

Our redis-cluster production
image


taylorotwell:
The `purge` method doesn't appear on the interface?

PR #1334, opened 2023-10-11
Fix the job repository class based on service bindings

daison12006013:
Continuation from my previous PR #1333


taylorotwell:
Sorry, we can't add a method to an interface on a patch release. You would need to inject the interface and do a method_exists call.

PR #1336, opened 2023-10-16
Support for laravel 11

ahsankhatri:
- Added support for Laravel 11 (include `composer.json` changes only)

Note: Since **main** branch isn't up to date, it shows an immense amount of commits to merge. Not sure which branch it should get merged on.


driesvints:
Seems like something went wrong with branching here

Issue #1348, opened 2023-11-17
Disable all jobs lists


andreladocruz:
Would be great if we had a way to disable all the lists (recent, pending, completed, failed, etc).

This consumes too much memory and in a heavily used system, this is not good at all.

driesvints:
I think in this case you'd want a more capable server to serve these.

andreladocruz:
@driesvints ,

We have all the logs we need in Google Cloud.

This kind of reply is unacceptable.

![Xnip2023-11-18_18-34-09](https://github.com/laravel/horizon/assets/3315596/e3b7f021-0e42-462a-973f-c05429d7f653)


![Xnip2023-11-18_18-28-51](https://github.com/laravel/horizon/assets/3315596/ad9abfd5-6a09-4386-bdc4-525351a54a35)
![Xnip2023-11-18_18-29-19](https://github.com/laravel/horizon/assets/3315596/54b5eb9b-bef1-4d7d-aa0a-09692380fc70)
![Xnip2023-11-18_18-29-40](https://github.com/laravel/horizon/assets/3315596/97c73f29-0744-4528-8d6d-4373f4e71aef)

driesvints:
I’m sorry if I offended you with my reply, that wasn’t my intention. If you want, you can always try a PR to see if it gets accepted 👍

andreladocruz:
@driesvints ,

Your response was completely disproportionate and rude.

Regarding the issue of making a PR to "see" if it will be accepted, it is also inelegant.

I completely understand the issue of having logs available for smaller systems with fewer jobs.

However, for systems like ours, they don't make any sense precisely because they take up memory, making our operational costs higher than necessary.

To give you an idea, these logs make us maintain 6Gb of Redis only for the queues when our average usage is less than 1Gb.

The difference only applies to peak moments where, in addition to a large volume of jobs in our queues, we also have to manage useless histories and statistics.

I tried setting the "Jobs Triming Time" to zero, but the result was even worse and they were not removed.

We have been using the Laravel ecosystem in our project since 2017 and whenever we try to contribute our experience, we are disrespected by you.

What a beautiful "community" you are building.

driesvints:
I tried to apologise here but it seems you're not interested in that. So feel free to move on.

andreladocruz:
@driesvints ,

So it's like this, you respond rudely and rudely and I don't have the right to make that clear?

Don't I also have the right to explain the entire situation better, providing even more details about the issue?

[...]

PR #1349, opened 2023-11-22
Only creates a reference if minutesForType > 0

andreladocruz:
This avoids consuming too much memory with references.

Conversation here: https://github.com/laravel/horizon/issues/1348


andreladocruz:
closing as the solution is not the right one

PR #1350, opened 2023-11-22
Delete completed jobs if expires equal to zero

andreladocruz:
Immediately removes the job if the expiration time is equal to zero, thus optimizing memory usage.

Context: https://github.com/laravel/horizon/issues/1348


taylorotwell:
Don't like the way you talk to us. Sorry. 👋

PR #1367, opened 2024-01-15
Add cache check before registering routes

serpentblade:
If routes are already cached, it is not necessary to re-register Horizon routes on every request.

I've found that many Laravel packages don't check for already cached routes and will needlessly re-register routes that already exist in the cache anyway.  This reduces boot times in production by ~0.5 ms which adds up when applied to multiple other first-party Laravel packages. 

I can open PRs in other projects if this is approved.

### DIFF
--- a/src/HorizonServiceProvider.php
+++ b/src/HorizonServiceProvider.php
@@ -3,6 +3,7 @@
 namespace Laravel\Horizon;
 
 use Illuminate\Contracts\Events\Dispatcher;
+use Illuminate\Contracts\Foundation\CachesRoutes;
 use Illuminate\Queue\QueueManager;
 use Illuminate\Support\Facades\Route;
 use Illuminate\Support\ServiceProvider;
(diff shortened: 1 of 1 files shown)


driesvints:
`loadRoutesFrom` already checks for this. Please check the underlying code.

serpentblade:
I see that `loudRoutesFrom` does handle this check already.  But that doesn't account for the fact that when routes are cached `Routes::group(...)` is called only to have nothing happen.

If adding this code is not considered a valid solution, perhaps moving `Route::groups(...)` into the routes file instead is more acceptable?  Ultimately my aim here is to not have this code run at all if it is not required to run.

driesvints:
Thanks. I'll leave the decision with @taylorotwell but I don't feel this is necessary.

siarheipashkevich:
@driesvints @serpentblade will be better to move 'Route::group' inside `routes` files for avoiding extra logic.

PR #1379, opened 2024-01-26
Ignore IDE local workspace folders

alirezadp10:
To work more efficiently on the project and contribute effectively, I believe it's a good idea to add local IDE workspace folders to the .gitignore file.


taylorotwell:
Put them in your global gitignore file.

Issue #1295, opened 2023-07-13
Why start at max processes for a new supervisor?


marcoskubis:
I noticed that when raising a new horizon instance, the horizon instantly scales the workers to the value defined in max processes. This causes server overhead whenever horizon is restarted, especially when we have a high number of workers and queues.

Wouldn't it be much better to start the workers using the value that's defined in min-processes, letting the balancer only scale if necessary?

Check this line. I suggest change max-processes to min-processes. What do you all think about it?

[https://github.com/laravel/horizon/blob/5.x/src/Console/SupervisorCommand.php#L98](https://github.com/laravel/horizon/blob/5.x/src/Console/SupervisorCommand.php#L98)

```
$supervisor->scale(max(
    0, $this->option('min-processes') - $supervisor->totalSystemProcessCount()
));
```

### More info? Ok.

Just setup a supervisor in `horizon.php`:

1. Define `maxProcesses=1000` and `minProcesses=1`
2. Define `balancer=auto`
3. Run `php artisan horizon`
4. See your laptop crash.

taylorotwell:
Can't speak to whether it's better or worse really - but in theory you should not set `max-processes` greater than what your machine can reasonably handle. If settings your max processes to 1000 causes your machine to crash, then that's a possibility regardless of whether it scales to that number initially or not. It could always scale up to that number 5 minutes later when your queue is busy.

marcoskubis:
Therefore, it would not be a good idea to make this configurable in `horizon.php`?

cyppe:
I agree on this one. For me it's not about if server crashes or not. It's just a bit waste of resources to scale up to 100% just to scale down to minimum a few seconds later if there is no jobs. Would be nice to have it configurable.

cyppe:
Added PR to address this. https://github.com/laravel/horizon/pull/1384

cyppe:
Declined as expected. Sorry, I tried.

Hope they change default instead to be minimum, because current method to scale to max makes zero sense. If you run in kubernetes each scale up can just trigger a chain of scale ups as memory and cpu goes to max on startup. Just one example of why its good idea to default to minimum.

PR #1384, opened 2024-02-04
Implement 'scaleToMin' configuration option for supervisor startup scaling method

cyppe:
This pull request introduces a new configuration option, scaleToMin, as a response to the community discussion in [Issue #1295](https://github.com/laravel/horizon/issues/1295). 

The addition of this feature enhances the flexibility and efficiency of Horizon's supervisor scaling behavior, particularly for environments where resource conservation is critical.

**Motivation:**
Currently, Horizon supervisors scale directly to the maximum number of processes (maxProcesses) upon startup. This behavior can lead to unnecessary resource usage (CPU/memory) in scenarios where the workload does not immediately require such capacity. The community has expressed a need for a more resource-efficient approach to scaling, especially important for applications with variable job loads.

**Implementation:**
The scaleToMin option allows developers to specify that supervisors should initially scale to the minimum number of processes (minProcesses, defaulting to 1) and then use the configured auto-scaling strategy to adjust the number of workers based on workload. This approach optimizes resource utilization by starting with a conservative number of worker processes and scaling up only as required.

**Configuration:**
To enable this behavior, set 'scaleToMin' => true in the relevant supervisor configuration within the horizon.php config file. For example:

```
'supervisor-1' => [
    ...
    'scaleToMin' => true,
    ...
],
```

When scaleToMin is enabled, initiating Horizon with `php artisan horizon` will start supervisors with minProcesses instead of maxProcesses.

**Benefits:**

- Resource Efficiency: Reduces initial resource consumption by scaling supervisors based on actual demand.

- Cost-Effective Scaling: Particularly beneficial for cloud-hosted applications where resource usage directly impacts hosting costs.

- Enhanced Flexibility: Offers developers more control over the scaling behavior of Horizon supervisors, allowing for tailored configurations that match their application's specific needs.

I look forward to your feedback and hope this contribution proves valuable.


taylorotwell:

marcoskubis:
It appears to be an automated message.

> Thanks for your pull request to Laravel!
> 
> Unfortunately, I'm going to delay merging this code for now. To preserve our ability to adequately maintain the framework, we need to be very careful regarding the amount of code we include.
> 
> If possible, please consider releasing your code as a package so that the community can still take advantage of your contributions!
> 
> If you feel absolutely certain that this code corrects a bug in the framework, please "**@**" mention me in a follow-up comment with further explanation so that GitHub will send me a notification of your response.

ryzr:
Hope this is reconsidered in future - starting at maxProcesses makes autoscaling on memory usage quite tricky.

mbardelmeijer:
Would love to see support for this as well. We're running into same issues were we can support peak load of some queues of maxProcesses, but having all the queues at max processes on start-up gives peak loads.

Perhaps Horizon can be modified to start up on the average value between min & max processes by default?

dbpolito:
I completely agree that we should start using min instead of max... and i completely feel you guys how frustrating it is sometimes when they close the PR / Issue like this.

If feels their OCD for not having opened issues / PRs is more important...

Like this one: https://github.com/laravel/horizon/pull/1442#issuecomment-2127826748

If you have no time, burying it instead of leaving it open for when the time comes feels pretty bad and waste of the time spent....

I ended up overwriting the SupervisorFactory and Supervisor in my project and overwrote the scale method with:

```
    {
        if (debug_backtrace()[1]['class'] === SupervisorCommand::class) {
            return;
        }

        parent::scale($processes);
    }
```

To ignore this call: https://github.com/laravel/horizon/blob/5.x/src/Console/SupervisorCommand.php#L99-L101

Issue #1068, opened 2021-09-08
Call redis.replicate_commands()


davidvandertuijn:
- Horizon Version: 5.7
- Laravel Version: 8.40
- PHP Version: 7.4.22
- Redis Driver & Version: predis/phpredis 3.2.12 & predis 1.1
- Database Driver & Version:

### Description:

`ERR Error running script (call to f_281c23c60c642d42fdedcc1acc8adc7317dd75a5): @user_script:18: @user_script: 18: Write commands not allowed after non deterministic commands. Call redis.replicate_commands() at the start of your script in order to switch to single commands replication mode.`

[Replicating commands instead of scripts](https://redis.io/commands/EVAL#replicating-commands-instead-of-scripts)
Howto call redis.replicate_commands() ?

### Steps To Reproduce:

`php artisan horizon:clear --queue=supplier_feeds_queue`

driesvints:
Hi there,

Thanks for reporting but it looks like this is a question which can be asked on a support channel. Please only use this issue tracker for reporting bugs with the library itself. If you have a question on how to use functionality provided by this repo you can try one of the following channels:

- [Laracasts Forums](https://laracasts.com/discuss)
- [Laravel.io Forums](https://laravel.io/forum)
- [StackOverflow](https://stackoverflow.com/questions/tagged/laravel)
- [Discord](https://discordapp.com/invite/KxwQuKb)
- [Larachat](https://larachat.co)
- [IRC](https://web.libera.chat/?nick=laravelnewbie&channels=#laravel)

However, this issue will not be locked and everyone is still free to discuss solutions to your problem!

Thanks.

logan-jobzmall:
I am having this issue as well - only against Elasticache however, local is fine -  @davidvandertuijn @driesvints  - was there any resolution to this?

markieo1:
This is still an issue. I have created an PR to fix it.

PR #1389, opened 2024-02-16
Add redis.replicate_commands() to make purging of queue work

markieo1:
It is impossible to clear a queue using the `horizon:clear` artisan command. This is also seeen in issue #1068 
The following exception is then thrown:

```
ERR Error running script (call to f_281c23c60c642d42fdedcc1acc8adc7317dd75a5): @user_script:18: @user_script: 18: Write commands not allowed after non deterministic commands. Call redis.replicate_commands() at the start of your script in order to switch to single commands replication mode.
```

This resolves the issue by adding the `redis.replicate_commands()` call in the Lua script.

### DIFF
--- a/src/LuaScripts.php
+++ b/src/LuaScripts.php
@@ -48,6 +48,8 @@ LUA;
     public static function purge()
     {
         return <<<'LUA'
+
+            redis.replicate_commands()
             
             local count = 0
             local cursor = 0


taylorotwell:
This was removed in Redis 7:

CleanShot 2024-02-16 at 10 08 22@2x

PR #1390, opened 2024-02-16
Use class binding in HorizonCommand

v1nc3nt67:
This pull request modify the horizon command to permit class binding to the MasterSupervisor.

Reason: Modifying Laravel\Horizon\MasterSupervisor require changes to the HorizonCommand which will cause unnecessary issues when upgrading

### DIFF
--- a/src/Console/HorizonCommand.php
+++ b/src/Console/HorizonCommand.php
@@ -31,17 +31,17 @@ class HorizonCommand extends Command
      */
     public function handle(MasterSupervisorRepository $masters)
     {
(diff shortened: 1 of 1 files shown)


taylorotwell:

v1nc3nt67:
@taylorotwell do you think modifying the HorizonCommand class to use class injection instead of calling the class directly would be a better approach? I don't think it's a 5.x modification as this could be breaking but for 6.x
If yes, how should I send in a PR for 6.x

Issue #375, opened 2018-08-05
Horizon dies after few seconds


spwin:
Hey,
I am trying to set up the local environment. I can see horizon dashboard with `Inactive` status.
When I am running `php artisan horizon` the status changes to `Active` for a second and then changes back to `Inactive` again. I tried using `supervisor` to run horizon but the same problem. From the supervisor logs I can see that it's running for a second, then restarting and starting again all the time. Is there any way to debug horizon and check why is it dying? I tried running `php artisan horizon -vvv` but every time the output is the same: Horizon started successfully. and then terminating the process. I can also confirm the queue is working fine when I run it using `php artisan horizon:work`.

**Environment details:**
Using Apache as a server on
Ubuntu 16.04.
Laravel version:  5.6.14
PHP version: 7.2
Redis server version: 4.0.11
Horizon version: 1.3

**Configuration**
config/queue.php
```
'redis' => [
            'driver' => 'redis',
            'connection' => 'default',
            'queue' => 'default',
            'retry_after' => 90,
            'block_for' => null,
        ],
```

config/horizon.php
```
'local' => [
            'supervisor-1' => [
                'connection' => 'redis',
                'queue' => ['default'],
                'balance' => 'auto',
                'processes' => 3,
                'tries' => 3,
            ],
        ],
```

.env
```
APP_ENV=local
QUEUE_DRIVER=redis
REDIS_HOST=127.0.0.1
REDIS_PASSWORD=null
REDIS_PORT=6379
```

**Any suggestions?**

spwin:
Changed OS to Ubuntu 18.04 - this solved the problem. But some logging on horizon would be useful anyway.

spwin:
@mfn I couldn't find other way than just comment `MonitorMasterSupervisorMemory.php` line 20 in vendor folder:
```
public function handle(MasterSupervisorLooped $event)
    {
        $master = $event->master;

        if ($master->memoryUsage() > 64) {
            // $master->terminate(12);
        }
    }
```
Works fine on local machine. If it's happening in production for you I think it's not a solution and you should increase memory on the server. 
Cheers!

mfn:
Hardcoded memory usage checks are never awesome; I think it would be very interesting what `$master->memoryUsage()` reports in your case. Do you think you will be able to get this value, e.g. via `\Log::debug($master->memoryUsage());`?

spwin:
[...]

PR #1391, opened 2024-02-20
Output console error when terminating due to memory usage

SanderMuller:
In #463 the memory limit was made configurable, however, when the memory limit is set too low (e.g. in my case more memory was being used than the default 64MB), Horizon will just fail silently.

For months Horizon was "broken" in one of my projects and I couldn't figure out why. No console output, no log output. The same is the case for #375.

This PR adds clear console output when `php artisan horizon` terminates due to using more memory than configured (or the default 64MB)

Before this PR:
image

After this PR: 
image

I first just added the output without the change in `src/Console/HorizonCommand.php`, that is still an option.
The output is then:

image


driesvints (approved):
Great PR 👍

PR #1397, opened 2024-03-17
Add job retry panel to under job information

CharlesBilbo:
This pull request relocates the retry widget shifting it from the bottom of the page to a more prominent position beneath the job information section. While seemingly minor, this adjustment addresses a significant issue. Myself and several colleagues, were unaware of its existence due to its placement at the page's bottom. Particularly with larger payloads, it was effectively concealed.

image

### DIFF
--- a/resources/js/screens/failedJobs/job.vue
+++ b/resources/js/screens/failedJobs/job.vue
@@ -176,36 +176,6 @@
             </div>
         </div>
 
-        <div class="card overflow-hidden mt-4" v-if="ready">
-            <div class="card-header d-flex align-items-center justify-content-between">
-                <h2 class="h6 m-0">Exception</h2>
-            </div>
-            <div>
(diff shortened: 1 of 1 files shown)


CharlesBilbo:
I wasnt sure if i needed to recompile the app.js or if thats done when the code is merged

lukasyelle:
Definitely would find this useful! 👍

driesvints:
> I wasnt sure if i needed to recompile the app.js or if thats done when the code is merged

That's all done automatically.

taylorotwell:
Doesn't this just mean now that the sections you moved to the bottom would be "undiscovered"?

CharlesBilbo:
not really since those components are present at all time while this one only appears when the button is clicks and is easily missed since there's not an indication that the job was retried.

PR #1403, opened 2024-03-28
Use morph-name instead of class-name for a tag

Cellard:
If job has no explicit tags, Horizon builds tags with every model, passed to a job constructor. Model's tag is a full class name and primary key.

I think, that we may use model's morph name intead of class name. It would be more laconic. 

You can see the difference:
image


taylorotwell:

PR #1407, opened 2024-04-03
Command to forget all failed jobs command

jaroslavstefanec:
This pull request introduces a new console command, `horizon:forget-all-failed-jobs`, which allows for the removal of all failed jobs.

**Benefits to End Users:**
This addition resolves various discussions regarding how to delete failed jobs and prevents users from resorting to using `flushdb` to delete only failed jobs.


taylorotwell:
Should just be `horizon:forget --all` if added at all.

PR #1408, opened 2024-04-03
[5.x] Customize the lock time for the notification to be re-sent.

ekateiva:
In order to be able to "silent" the notifications about the same long wait times, there could be a setting.

For example, we are aware about the situation for long waiting time in queue and would like to get notifications not so often (now it can not be customized as is sent every 5 minutes). Example in Slack:
Screenshot 2024-04-03 at 15 01 01

This setting could be customizable the same way as `waits`, where the combination of connection / queue could have it's own setting.

### DIFF
--- a/config/horizon.php
+++ b/config/horizon.php
@@ -89,6 +89,23 @@ return [
 
     /*
     |--------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


taylorotwell:

PR #1415, opened 2024-04-13
[5.x] feature: this adds the vite integrity plugin to support assetsAreCurrent check

mmachatschek:
This adds the manifest integrity plugin for vite to support checks if the user published the current version of the horizon assets.

### DIFF
--- a/resources/views/layout.blade.php
+++ b/resources/views/layout.blade.php
@@ -1,8 +1,20 @@
 @php
 use Illuminate\Support\Facades\Vite;
+use Illuminate\Foundation\Vite as ViteFoundation;
+
+$viteDataSchemeLight = new ViteFoundation();
+$viteDataSchemeLight->useStyleTagAttributes([
+'data-scheme' => 'light',
+]);
+
+$viteDataSchemeDark = new ViteFoundation();
+$viteDataSchemeDark->useStyleTagAttributes([
+'data-scheme' => 'dark',
+]);
 @endphp
 <!DOCTYPE html>
 <html lang="en">
+
 <head>
     <!-- Meta Information -->
     <meta charset="utf-8">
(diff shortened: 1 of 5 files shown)


resources/views/layout.blade.php
```diff
@@ -1,8 +1,20 @@
...
+
+$viteDataSchemeLight = new ViteFoundation();
+$viteDataSchemeLight->useStyleTagAttributes([
+'data-scheme' => 'light',
+]);
+
+$viteDataSchemeDark = new ViteFoundation();
+$viteDataSchemeDark->useStyleTagAttributes([
+'data-scheme' => 'dark',
+]);
```
mmachatschek, line 13: if there is a better solution to this please tell me. currently its fairly hard (impossible) to add these kind of style tag attributes with the @vite blade directive

resources/views/layout.blade.php
```diff
@@ -1,8 +1,20 @@
...
 @endphp
 <!DOCTYPE html>
 <html lang="en">
+
```
driesvints:
```suggestion
```

driesvints:
@mmachatschek what would they see if the assets aren't up to date? I also thought we already had a check for this 🤔

mmachatschek:
@driesvints there is a file compare check here https://github.com/laravel/horizon/blob/3c6661d19013d108143ac204bc5f0e4f3313bf68/src/Horizon.php#L195 that checks if the manifest.json in the published folder is the same as the manifest.json from the package.

with laravel mix we used versioning that added a hash `?id=` for invalidating https://github.com/laravel/horizon/blob/32ca28dcb91fa2f4fa650f21440144c4bc2846dd/public/mix-manifest.json#L2 the built files.

the integrity package readds this missing feature after my change from mix to vite

driesvints (approved):

PR #1421, opened 2024-04-19
[5.x] Readd vite support

### LINKED ISSUE #1416: Manifest.json cause 404 error if file in resources is missing
### Horizon Version

5.24.1

### Laravel Version

11.4.0

### PHP Version

8.3.4

### Redis Driver

PhpRedis

### Redis Version

7.2.4

### Database Driver & Version

_No response_

### Description

This update cause a bug when opening horizon dashboard with error like:
`GET http://localhost:5173/resources/js/app.js net::ERR_ABORTED 404 (Not Found)`
`GET http://localhost:5173/resources/sass/styles.scss net::ERR_ABORTED 404 (Not Found)`

This is because in some of our project we don't use js anymore but ts files also we don't use styles.scss as entrypoint anymore but another file. This issue will also break any project without those files as entrypoint.

After reading commit history, this file seems to be the culprit: `public/build/manifest.json`

### Steps To Reproduce

1. Create new laravel project
2. Install horizon
3. Remove or rename file `resources/js/app.js` and `resources/sass/styles.scss`
4. Open horizon dashboard and this issue will happen

mmachatschek:
Fixes #1416

This readds the vite implementation to bundle the assets with a fix to support `npm run dev` mode on the side of consuming laravel apps.

Previously, the horizon page was broken because the Vite blade helper and the Vite Facade helper do not take into account that prebuilt assets by "third party" (in the sense of assets loaded by packages) packages should not be treated as uncompiled assets from the same laravel app.

The workaround here is suboptimal as I think the Vite blade helpers and the Vite Facade don't take into account that packages may also use these helper methods to render tags.


taylorotwell:
Hey @mmachatschek - can you explain what about this solution is "suboptimal"?

mmachatschek:
@taylorotwell suboptimal in the sense of having to pass in a non existent vite `hotfile` name [in the view](https://github.com/mmachatschek/horizon/blob/33eb7c38e4e8b33e54f7dcd4c102d54bee9caf7b/resources/views/layout.blade.php#L8) to work around the vite helper not being able to detect that the `vendor/horizon` assets were already published, as it's been pointed out here: https://github.com/laravel/horizon/issues/1416#issuecomment-2060505259

jonathan-bird:
Please note that this caused an issue: https://github.com/laravel/horizon/issues/1439

Should at least use `base_path` or similar rather than a root dir slash in layout.blade.php to find your non existent file.

PR #1422, opened 2024-04-21
Share an idea I use on other projects to clear all Horizon jobs

alnutile:
As a user I sometimes just want to clear all Horizon queues and not one by one by name.

This command does that.

```bash
php artisan horizon:clear_all
```


taylorotwell:

alnutile:
Good idea about the package. Was that ai that replied though or the real thing? 🤔

PR #1427, opened 2024-04-24
[5.x] Fix setting custom ASSET_URLs and support cache busting without integrity check

mmachatschek:
Fixes #1425
Fixes #1426

There were already tries to integrate Vite with [telescope which have been postponed](https://github.com/laravel/telescope/pull/1391) because of the linked issues and other concerns. I think this PR now solves all of these issues.

@timacdonald I would love to hear your opinion as well as you were involved in a lot of these discussions as well:

- https://github.com/laravel/framework/issues/48964
- https://github.com/laravel/framework/pull/49329 (missing cleanup functionality for packages exporting vite files that include the file hash in the filename) because of this, we need to manually craft cache busting ids to the filename in the manifest instead of using the normal hashed files of vite e.g. `app.129hg41.js`. Without the cleanup, users would end up with a lot of outdated files published by packages like laravel/horizon and laravel/telescope
- https://github.com/laravel/telescope/pull/1391 -> has been reverted because of the broken `npm run dev` commands which this version of the vite implementation resolves

also related:
- https://github.com/laravel/framework/pull/45257
- https://github.com/laravel/vite-plugin/pull/251 (cleanup on the javascript side for the laravel-vite-plugin)

Screenshots:

Running vite dev mode renders the "none" vite url
image

Setting a custom ASSET_URL works:

image

PR #1428, opened 2024-04-25
Make the non-existing file in layout.blade within project

marcin-jozwikowski:
If the PHPs `open_basedir` value is configured, any check for a file outside of given scope causes  `is_file(): open_basedir restriction in effect. File(/vendor/horizon/nonExistentFile) is not within the allowed path(s):` error, which prevents the UI from loading.

### DIFF
--- a/resources/views/layout.blade.php
+++ b/resources/views/layout.blade.php
@@ -2,7 +2,7 @@
 use Illuminate\Support\Facades\Vite;
 use Illuminate\Foundation\Vite as ViteFoundation;
 
-$nonExistentFileName = '/vendor/horizon/nonExistentFile';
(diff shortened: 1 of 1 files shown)


driesvints:
@mmachatschek how does this one compare to your own PR? https://github.com/laravel/horizon/pull/1427

mmachatschek:
@driesvints this should be fixed with #1427 as I use the public path now `$nonExistentFileName = public_path('/vendor/horizon/nonExistentFile');` 

https://github.com/mmachatschek/horizon/blob/a6720ac880159d161aa30d3fff45c85de8d65f9a/resources/views/layout.blade.php#L2

driesvints:
Closing this one in favour of @mmachatschek's solution.

PR #1438, opened 2024-05-02
Simplify asset story

timacdonald:
Horizon was migrated to Vite + `laravel-vite-plugin`. The `laravel-vite-plugin` is not meant for packages. There was some additional work done to support the asset URL. We ended up having a lot of workarounds to make things work.

This PR rethinks Horizon's asset story completely to try and simplify things.

### Inline assets on the dashboard

CSS, JS, and the favicon are now delivered inline. The horizon dashboard is a single-page app, so there is no real-world drawbacks here IMO.

The total transfer size is ~280kb over the wire. Yes we lose browser caching of the assets, but we also had to introduce a lot of manual work to add cache busting and 280kb is nothing.

This just works™️, assets are _always_ up to date, we no longer have to publish assets whenever we change them, and applications that use a CDN for assets are no longer having to deliver assets via their web servers - which was an issue that has been raised.

### Remove `laravel-vite-plugin` and usages of the `Vite` facade.

As I mentioned, the `laravel-vite-plugin` is meant for Laravel apps. It doesn't really offer anything for packages and in some ways makes things harder.

We just use raw Vite now when working on Horizon.

```
# Build for production
npm run build

# Rebuild on changes
npm run watch
```

## Minor changes

- Build assets moved from `/public/build` to `/dist`

## Considerations

- This implementation doesn't currently support HMR. It can be easily added, but I'm not sure if we are developing Horizon enough to need it? Just run `npm run watch` and it will rebuild the assets whenever they change.


.styleci.yml
```diff
@@ -1,4 +1,34 @@
...
+  double-quotes: false
+  finder:
+    exclude:
+      - "dist"
+      - "node_modules"
+      - "vendor"
+    name:
+      - "*.css"
+      - "*.scss"
+      - "*.less"
```
timacdonald, line 34: StyleCI was having issues with the files in `dist`. 

I've copied this config from their docs (this is what is configured with `js: true` or `css: true` and added `dist` to the exclude.

package.json
```diff
@@ -1,25 +1,21 @@
...
         "axios": "^1.6.8",
         "bootstrap": "~5.1.3",
         "chart.js": "^2.9.4",
         "highlight.js": "^10.7.3",
-        "laravel-vite-plugin": "^1.0.2",
         "md5": "^2.3.0",
         "moment": "^2.30.1",
         "moment-timezone": "^0.5.45",
         "phpunserialize": "^1.3.0",
-        "resolve-url-loader": "^5.0.0",
```
timacdonald, line 22: This is a webpack plugin. We don't need it anymore as we use Vite.

resources/js/app.js
```diff
@@ -8,8 +8,6 @@ import Routes from './routes';
 import Alert from './components/Alert.vue';
 import SchemeToggler from './components/SchemeToggler.vue';
 
-import.meta.glob(['../img/**']);
-
```
timacdonald, line 12: This was here for the Favicon. We don't need it anymore as we inline the favicon.

resources/views/layout.blade.php
```diff
@@ -127,12 +118,6 @@
             </div>
 
             <div class="col-10">
-                @if (! $assetsAreCurrent)
-                    <div class="alert alert-warning">
-                        The published Horizon assets are not up-to-date with the installed version. To update, run:<br/><code>php artisan horizon:publish</code>
-                    </div>
-                @endif
-
```
timacdonald, line 135: Assets are always up to date ✌️

vite.config.mjs
```diff
@@ -1,66 +0,0 @@
...
-            "resources/sass/styles.scss",
-            "resources/sass/styles-dark.scss",
-            "resources/js/app.js",
-        ]),
-        vue2(),
-        manifestQueryParam(),
-    ],
-    resolve: {
-        alias: {
-            vue: "vue/dist/vue.esm.js",
```
timacdonald, line 52: Removed this alias and used it inline instead to make it more explicit. This was kind of hidden otherwise.

timacdonald:
@driesvints, if you have a release script for this you will need to update the build command to `npm run build`.

driesvints:
@timacdonald our repos with asset compilation use the compile assets workflow. I've made the necessary adjustments but just FYI it's this: https://github.com/laravel/horizon/commit/8d31ff178bf5493efc2b2629c10612054f31f584

timacdonald:
Thanks!

PR #1448, opened 2024-05-24
Add --fail Option to Horizon Command for Deployment Failure on Unsuccessful Termination

comes:
If you use deployment tools like Envoyer, there is a good chance that you would like the entire deployment process to fail if Horizon was not terminated. This might happen if you change your Redis connection or for another reason.

Currently, the command works successfully with an info message that no process was restarted. This PR does not change that but lets you specify that the command should actually fail in that case.

It does not break any existing feature because the new `--fail` option is an addition. It stops the artisan command by sending a UNIX cli error code 1 (`self::FAILURE`), which indicates that something went wrong.


taylorotwell:
I dunno, I'm just not sure I love this method.

Issue #1447, opened 2024-05-20
Feature Request: Warn the user if no supervisor configuration is present in the selected environment


yudikubota:
I just spent an awful lot of time trying to debug why horizon was not running just to realize I was on staging environment and there was no staging supervisors... 

I think it would be nice to warn the user if no supervisors are configured.

driesvints:
But doesn't the dashboard show no supervisors in the list in that scenario?

yudikubota:
> But doesn't the dashboard show no supervisors in the list in that scenario?

Yeah, but I didn't check the dashboard, only the CLI as it was running inside a docker container for staging environment.

This is a feature request for the CLI. Something like:

```bash
php artisan horizon

INFO  Horizon started successfully.  

WARN no supervisor configuration for environment "staging"
```

I ask you to consider reopening this issue. Maybe other people stumble upon this exact mistake.

driesvints:
Ah right, I understand. Yeah I can see that being useful, feel free to attempt a PR 👍

yudikubota:
> Ah right, I understand. Yeah I can see that being useful, feel free to attempt a PR 👍

Just made my first PR to this repo.

PR #1451, opened 2024-05-25
[5.x] Feature: Warn if no supervisor config (#1447)

yudikubota:
As discussed in https://github.com/laravel/horizon/issues/1447, users may run `php artisan horizon` in a different environment and forget to configure the supervisors for this new environment. This feature warns if there are no supervisors configured for the current environment.

This PR implements:

- Warning when no supervisor configuration is present;
- Integration tests to check the warning;

Obs.

It's possible to warn from the MasterSupervisor output but I think the HorizonCommand is more appropriate.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "editor.formatOnSave": false
-}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "editor.formatOnSave": false
+}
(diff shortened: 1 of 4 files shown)


driesvints (approved):
This is better but I also wonder if we should start Horizon at all if no supervisor config was found. Will let @taylorotwell wage in here.

taylorotwell:
No plans to take any action here.

driesvints:
@yudikubota looks like we're going to leave things be for now, sorry

PR #1453, opened 2024-05-28
Updated Redis Prefix Comment

darkspock:
Updated Redis Prefix Comment to ensure it ends in :, may is better to ensure by code this.


taylorotwell:
Doesn't follow format of rest of docs.

Issue #1450, opened 2024-05-24
Horizon worker sometimes stops too soon, making the shutdown ungraceful


nckrtl:
### Horizon Version

5.24.4

### Laravel Version

11.8.0

### PHP Version

8.3.7

### Redis Driver

PhpRedis

### Redis Version

5.3.7

### Database Driver & Version

sqlite, 3043002

### Description

When there is 1 job being processed by an horizon worker while a horizon:terminate is initiated the worker process often stops before the job finishes. The job will get stuck in the pending state and will finally get marked as failed after the retry_after threshold. The Job won’t get retried after you’ll restart the horizon worker, although the horizon ui will show its been retried multiple times once it's marked as failed.

This behaviour only seems to occur when there is 1 job being processed and a horizon:terminate is being initiated. When there are multiple jobs being processes of if there are multiple jobs in the queue the early shutdowns never seem to occur.

I also made a video demonstrating the behaviour:
https://www.youtube.com/watch?v=GRgw2LWyLto

[0:00](https://www.youtube.com/watch?v=GRgw2LWyLto&t=0s) Run 1: job gets stuck
[0:27](https://www.youtube.com/watch?v=GRgw2LWyLto&t=27s) Run 2: job finishes
[1:04](https://www.youtube.com/watch?v=GRgw2LWyLto&t=64s) Run 3: job finishes
[1:47](https://www.youtube.com/watch?v=GRgw2LWyLto&t=107s) Run 4: job gets stuck
[2:17](https://www.youtube.com/watch?v=GRgw2LWyLto&t=137s) Run 5: job finishes
[2:55](https://www.youtube.com/watch?v=GRgw2LWyLto&t=175s) Run 6: job gets stuck, also shows it wont continue after turning horizon back on
[3:49](https://www.youtube.com/watch?v=GRgw2LWyLto&t=229s) Run 7: job finishes
[4:33](https://www.youtube.com/watch?v=GRgw2LWyLto&t=273s) Run 8: job gets stuck

I tested if it was related to the cache driver, but it seems to happen with any cache driver. Als the database driver seems irrelevant. I think there is some race condition that's causing this behaviour as it seems completely random.

I'm aware that this is an edge case, as I don't think it will happen often that horizon is processing exactly one job while shutting down. However, I think it would be a problem worth solving as it would make Horizon more reliable.

[...]

PR #1454, opened 2024-05-31
Update runningProcesses to also include terminating processes

nckrtl:
This PR is an alternative for https://github.com/laravel/horizon/pull/1433 which both solve https://github.com/laravel/horizon/issues/1450. The difference with this PR is that terminating processes will be included in the `runningProcesses()` function, as suggested by @crynobone. Whenever there is a need to check for running processes elsewhere in the project the terminating processes will be included by default, instead of having to keep adding an additional check for terminating process like the other PR does https://github.com/laravel/horizon/pull/1433.

### DIFF
--- a/src/ProcessPool.php
+++ b/src/ProcessPool.php
@@ -293,7 +293,11 @@ class ProcessPool implements Countable
      */
     public function runningProcesses()
     {
-        return collect($this->processes)->filter(function ($process) {
+        $terminatingProcesses = $this->terminatingProcesses()->map(function ($process) {
+            return $process['process'];
+        });
+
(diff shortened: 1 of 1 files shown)


nckrtl:
Tests are passing locally, are the unsuccessful checks a matter of re-running them?

driesvints:
@nckrtl seems to specifically fail on the PHP 8.3 and L11 combo. Wondering why 🤔

crynobone:
It probably has something to do with Symfony Process 7.1.0 (which just released few hours ago).

driesvints:
@nckrtl this is now fixed on 5.x, can you rebase?

crynobone (approved):

taylorotwell:
We'll give it a shot. 👍

PR #1455, opened 2024-05-31
Update phpunit.xml.dist

nckrtl:
This solves the following message when running the tests:
```
There was 1 PHPUnit test runner deprecation:

1) Your XML configuration validates against a deprecated schema. Migrate your XML configuration using "--migrate-configuration"!
```

The content is the result of running `./vendor/bin/phpunit --migrate-configuration` as suggested by the message. The deprecation message is now resolved.


crynobone:
The new configuration is not compatible with PHPUnit 9 and still required to run the tests with older Laravel versions.

PR #1464, opened 2024-06-11
Fix worker scaling when balance mode is auto

### LINKED ISSUE #1295: Why start at max processes for a new supervisor?

### LINKED ISSUE #51765: Laravel Horizon is scaling the wrong way (auto + max-processes)
### Laravel Version

any 

### PHP Version

any

### Database Driver & Version

any

### Description

Since nobody cares, I am trying to get some attention here.

everything is explained here https://github.com/laravel/framework/discussions/48431

TLDR; when horizon starts max worker get started, even they have nothing to do and get scaled down afterwards.
this concept is terribly wrong. a "balancer" should always scale up, and not down.

i hope that someone understands this simple logic and even this is not a software bug, it´s a "failed by concept" thing
and because of that, it needs to be changed.

thanks!

### Steps To Reproduce

start horizon with 500 worker + auto + max-processes balance logic. 
what happens ? horizon starts 500 worker even nothing is in the queue. 
then all workers will be stopped till only one is running.

what should happen?
actually NONE worker needs to be started (or at least one) if horizon starts.
based on the queue horizon needs to scale up.

again: the concept of a scaler should be: scale UP, not scale down 😏

michabbb:
fix https://github.com/laravel/framework/issues/51765 
fix https://github.com/laravel/horizon/issues/1295

as explained: https://github.com/laravel/framework/discussions/48431

if horizon is in mode balance=auto, there is no need to start hundreds of workers if the queue is empty.
it's okay to start with only 1 worker. if the queue fills, horizon scales up. that's the way it should work.

### DIFF
--- a/src/Console/SupervisorCommand.php
+++ b/src/Console/SupervisorCommand.php
@@ -96,9 +96,13 @@ class SupervisorCommand extends Command
 
         $supervisor->working = ! $this->option('paused');
 
-        $supervisor->scale(max(
-            0, $this->option('max-processes') - $supervisor->totalSystemProcessCount()
-        ));
+        if ($this->option('balance') === 'auto') {
+            $supervisor->scale(1);
+        } else {
+            $supervisor->scale(max(
+                0, $this->option('max-processes') - $supervisor->totalSystemProcessCount()
+            ));
+        }
 
         $supervisor->monitor();
     }


taylorotwell:
I think this is just an opinion of how it should work and probably not worth changing on a patch release.

michabbb:
@taylorotwell it's very frustrating to see that nobody at laravel understands the simple fact, that workers should scale up, and not down. If the queue is empty, why should the server run on max? Why? That makes no sense. This whole thing is glass clear. I have a project where 200 workers can scale up. Why the heck should all 200 workers start from the beginning and get ended after a few minutes, when horizon finally noticed: "oh, there is absolutely nothing to do for me, so let's kill all the works we didn't need in the first place" 🙈

So, what's needs to be done to fix that? 😏 Please, let us know.🙏

michabbb:
Besides that: with all changes laravel had in the past, there was always the will to evolve, keep things simple, remove overhead, make things slim and minimal.

Saying, "it's just an opinion" that it makes sense to start something that is not needed (tons of workers when the queue is empty) - speaks completely against your own principles.

I appeal to your reason and ask you: please think again. Thank you.

vesper8:
@michabbb maybe if you make a benchmark that shows a slower launch or degraded performance as a result of the way it's currently implemented then they'll think it's more than just an opinion. I haven't tried it myself but I would guess that spawning this many workers and then de-spawning them would create a potentially huge memory spike for no justifiable reason.. if that's what's happening in your case. If not then.. I guess it doesn't really matter then.

michabbb:
@vesper8 thanks for your willingness to help. I really think that everything is already said, multiple times.

Of course you have a huge spike, cpu, memory, maybe database connections, everything - and everything is completely Unnecessary. It's frustrating to see someone who invented laravel but at the same time is unwilling to understand such a simple fact 😔

 It's NOT an option, as Taylor said - it's just wrong. The concept is wrong By design. Scaling in the world of queues just simply does not work like that. Since laravel is not a democracy, it's a one-man-decision show, we have to live with that.

At least it's open source, so in my ci process I will deal for the rest of my life with a tiny hack that changes one single line in horizon - it is, what it is. 🤷

PR #1465, opened 2024-06-13
Test wildcard environments config

karkowg:
I might be looking at this totally wrong but I believe `ProvisioningPlanTest`::[test_supervisors_are_added_by_wildcard](https://github.com/laravel/horizon/blob/14eed5662588cd115cd4b6b3c4248b540357be42/tests/Feature/ProvisioningPlanTest.php#L39) isn't testing anything related to wildcard environments. It just looks like a duplicate of `ProvisioningPlanTest`::[test_supervisors_are_added](https://github.com/laravel/horizon/blob/14eed5662588cd115cd4b6b3c4248b540357be42/tests/Feature/ProvisioningPlanTest.php#L13).

I came by this because each member of my team is using a different APP_ENV and I was looking for a simple way to declare that in the Horizon config without having to list them all. I thought adding a wildcard environment wasn't possible, but it is, only missing documentation. If my assumptions are correct, here's a fix to that test. I've also sent https://github.com/laravel/docs/pull/9706 for documenting the wildcard usage.

Thanks!

### DIFF
--- a/tests/Feature/ProvisioningPlanTest.php
+++ b/tests/Feature/ProvisioningPlanTest.php
@@ -62,6 +62,32 @@ class ProvisioningPlanTest extends IntegrationTest
         $this->assertSame(20, $command->options['maxProcesses']);
     }
 
+    public function test_supervisors_are_added_as_fallback_for_wildcard_environments()
+    {
+        $plan = [
+            '*' => [
+                'supervisor-1' => [
+                    'connection' => 'redis',
+                    'queue' => 'first',
+                    'max_processes' => 10,
+                ],
+            ],
+        ];
+
+        $plan = new ProvisioningPlan(MasterSupervisor::name(), $plan);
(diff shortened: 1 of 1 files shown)


taylorotwell:

taylorotwell:
Can you add an entirely new test that you think is properly setup?

karkowg:
> Can you add an entirely new test that you think is properly setup?

What would be the benefit of having a new test?

If you diff `test_supervisors_are_added` and `test_supervisors_are_added_by_wildcard`, the only difference is the function name.

Don't you think the change is proper in that case? If not, I'm not seeing what that test is doing in its current form.

driesvints:
@karkowg could you please just add the new test as Taylor requests? It's better to leave the existing tests alone.

karkowg:
@driesvints yep, no problem. I was just trying to understand what I would have to do differently if adding a new test.

Issue #1461, opened 2024-06-06
Incorrect status on duplicate worker


jasonmccreary:
### Horizon Version

5.24.5

### Laravel Version

10.48.12

### PHP Version

8.3.1

### Redis Driver

Predis

### Redis Version

2.2.2

### Description

A bit of an edge-case here, but I have duplicate workers that spin up which have the same configuration - supervisor and horizon.

I noticed that if I run `horizon:pause` on _worker-1_, then spin up a new worker (_worker-2_) `horizon:status` still says "Paused". However, the Horizon web dashboard correct shows the "play" icon for _worker-2_ and "paused" icon for _worker-1_.

I'm glad to dig into the code, but don't have time now. So I wanted to get this tracked somewhere. My guess is the logic the `horizon:status` command uses to check status is not the same as the logic the web dashboard uses.

---
I've attached some screenshots for a visual diff of the issue where the web dashboard show the status correctly, but CLI does not.

**Horizon Dashboard**
<img width="1209" alt="Horizon web dashboard" src="https://github.com/laravel/horizon/assets/161071/2e64dc0c-d9c2-4e27-82c6-4145c5f90783">

**CLI**
<img width="618" alt="horizon:status command output" src="https://github.com/laravel/horizon/assets/161071/8075b051-c333-4de4-8a73-5a70b754ac64">

### Steps To Reproduce

- Start two identical workers with the same supervisor and horizon config
- Run `horizon:pause` on one
- Start the other worker
- Run `horizon:status` on the other

mfrieswyk:
The `horizon:list` command checks all registered supervisors, and if just one has a "paused" status it outputs "Horizon is paused". What would be the preferred output here @jasonmccreary? A table similar to the dashboard?

jasonmccreary:
@mfrieswyk, I would expect if I am on the worker and run `horizon:status` it would show the current status for that worker.

In lieu of that, then yeah, a list of all worker statuses would be better, imo, than saying "paused" when the current worker is not the one that is paused.

mfrieswyk:
@jasonmccreary I went ahead and made [this PR](https://github.com/laravel/horizon/pull/1467) to address this issue. It's not a full list of all worker statuses because I felt like changing the behavior of the main status command might be a bit much. Instead I made a new command in the style of the pause/continue commands that requires a name of the supervisor you would like the status for. Hope this helps.

driesvints:
@mfrieswyk do you have time to answer Taylor's question on that PR?

[...]

PR #1467, opened 2024-06-24
Create status command for individual supervisors

mfrieswyk:
Following the pattern of the `horizon:pause-supervisor` and `horizon:continue-supervisor` commands, this PR aims to address [Issue 1461](https://github.com/laravel/horizon/issues/1461) by introducing a `horizon:status-supervisor` command.

The existing `horizon:status` command only checks for the presence of a single paused supervisor before reporting that "Horizon is paused". This new command will allow users with multiple configurations to see accurate statuses for each supervisor.

### DIFF
--- a/src/Console/StatusSupervisorCommand.php
+++ b/src/Console/StatusSupervisorCommand.php
@@ -1,0 +1,51 @@
+<?php
+
+namespace Laravel\Horizon\Console;
+
+use Illuminate\Console\Command;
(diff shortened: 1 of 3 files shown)


taylorotwell:
How did you test this command manually? It's never registered with Artisan?

mfrieswyk:
Sorry for the delay, here's the command being successfully invoked locally after registering it:
image

PR #1469, opened 2024-07-05
Feature separate path UI api

jokosulistyo-iris:
Separate path route of ui and api


driesvints:
This is a breaking change for anyone using the `HORIZON_PATH` env variable.

PR #1472, opened 2024-07-09
[5.x] Add clear all queues option

TWithers:
This PR adds an option to clear all queues: `artisan horizon:clear --all-queues`

This is ideal for development as Laravel/Horizon apps might have multiple queues and this would provide a quick and easy way to clear all the queues at once instead of calling the `horizon:clear` command multiple times.

It grabs the default queue, and all the horizon queues from the config and issues the clear command individually to each queue. 

This PR also adds a test for the command as there wasn't previously a test.


taylorotwell:

PR #1473, opened 2024-07-10
[5.x] Adding scaling to balance = false

dbpolito:
This introduces a new strategy for balance: single

What this does is, it still has scaling similar to `auto`, but instead of having one worker per queue, it still keeps all queues into the same process, following the order specified, similar to what you would have using `queue:work --queue=a,b,c`

TBH this was the behavior i expected when i started using horizon a long time ago...

I'm using it in my project and thought it would be useful for others...

I can open a PR on the Docs in case this gets accepted / merged.

**Maybe the name isn't that clear, totally open to change.**

### DIFF
--- a/src/AutoScaler.php
+++ b/src/AutoScaler.php
@@ -77,11 +77,18 @@ class AutoScaler
     protected function timeToClearPerQueue(Supervisor $supervisor, Collection $pools)
     {
         return $pools->mapWithKeys(function ($pool, $queue) use ($supervisor) {
-            $size = $this->queue->connection($supervisor->options->connection)->readyNow($queue);
+            $queues = collect(explode(',', $queue))->map(function ($_queue) use ($supervisor)  {
+                $size = $this->queue->connection($supervisor->options->connection)->readyNow($_queue);
+
+                return [
+                    'size' => $size,
+                    'time' => ($size * $this->metrics->runtimeForQueue($_queue)),
+                ];
+            });
 
             return [$queue => [
-                'size' => $size,
-                'time' => ($size * $this->metrics->runtimeForQueue($queue)),
+                'size' => $queues->sum('size'),
+                'time' => $queues->sum('time'),
             ]];
         });
     }
--- a/src/AutoScaler.php
+++ b/src/AutoScaler.php
@@ -77,7 +77,7 @@ class AutoScaler
     protected function timeToClearPerQueue(Supervisor $supervisor, Collection $pools)
     {
         return $pools->mapWithKeys(function ($pool, $queue) use ($supervisor) {
-            $queues = collect(explode(',', $queue))->map(function ($_queue) use ($supervisor)  {
+            $queues = collect(explode(',', $queue))->map(function ($_queue) use ($supervisor) {
                 $size = $this->queue->connection($supervisor->options->connection)->readyNow($_queue);
 
                 return [
--- a/src/Supervisor.php
+++ b/src/Supervisor.php
@@ -87,7 +87,7 @@ class Supervisor implements Pausable, Restartable, Terminable
      */
     public function createProcessPools()
     {
-        return $this->options->balancing()
+        return $this->options->balancing() && ! $this->options->single()
                         ? $this->createProcessPoolPerQueue()
                         : $this->createSingleProcessPool();
     }
--- a/src/Supervisor.php
+++ b/src/Supervisor.php
@@ -87,7 +87,7 @@ class Supervisor implements Pausable, Restartable, Terminable
      */
     public function createProcessPools()
     {
-        return $this->options->balancing() && ! $this->options->single()
+        return $this->options->balancing()
                         ? $this->createProcessPoolPerQueue()
                         : $this->createSingleProcessPool();
     }
--- a/src/SupervisorOptions.php
+++ b/src/SupervisorOptions.php
@@ -265,7 +265,17 @@ class SupervisorOptions
      */
     public function autoScaling()
     {
-        return $this->balance === 'auto';
+        return in_array($this->balance, ['auto', 'single']);
+    }
+
+    /**
+     * Determine if auto-scaling single is applied.
+     *
+     * @return bool
+     */
+    public function single()
+    {
+        return $this->balance === 'single';
     }
 
     /**
--- a/src/SupervisorOptions.php
+++ b/src/SupervisorOptions.php
@@ -265,17 +265,7 @@ class SupervisorOptions
      */
     public function autoScaling()
     {
-        return in_array($this->balance, ['auto', 'single']);
-    }
-
-    /**
-     * Determine if auto-scaling single is applied.
-     *
-     * @return bool
-     */
-    public function single()
-    {
-        return $this->balance === 'single';
+        return $this->balance !== 'simple';
     }
 
     /**
--- a/tests/Feature/SupervisorTest.php
+++ b/tests/Feature/SupervisorTest.php
@@ -95,10 +95,9 @@ class SupervisorTest extends IntegrationTest
         );
     }
 
-    public function test_supervisor_starts_one_pool_when_single_balancing()
+    public function test_supervisor_starts_pools_with_queues_when_balancing_is_off()
     {
         $options = $this->supervisorOptions();
-        $options->balance = 'single';
         $options->queue = 'first,second';
         $this->supervisor = $supervisor = new Supervisor($options);
 
@@ -177,6 +176,7 @@ class SupervisorTest extends IntegrationTest
     public function test_supervisor_information_is_persisted()
     {
         $this->supervisor = $supervisor = new Supervisor($options = $this->supervisorOptions());
+        $options->balance = 'simple';
         $options->queue = 'default,another';
 
         $supervisor->scale(2);
@@ -207,7 +207,8 @@ class SupervisorTest extends IntegrationTest
 
     public function test_processes_can_be_scaled_up()
     {
-        $this->supervisor = $supervisor = new Supervisor($this->supervisorOptions());
+        $this->supervisor = $supervisor = new Supervisor($options = $this->supervisorOptions());
+        $options->balance = 'simple';
 
         $supervisor->scale(2);
         $supervisor->loop();
@@ -221,6 +222,7 @@ class SupervisorTest extends IntegrationTest
     public function test_processes_can_be_scaled_down()
     {
         $this->supervisor = $supervisor = new Supervisor($options = $this->supervisorOptions());
+        $options->balance = 'simple';
         $options->sleep = 0;
 
         $supervisor->scale(3);
@@ -491,6 +493,7 @@ class SupervisorTest extends IntegrationTest
     {
         SystemProcessCounter::$command = 'worker.php';
         $this->supervisor = $supervisor = new Supervisor($options = $this->supervisorOptions());
+        $options->balance = 'simple';
 
         $supervisor->scale(3);
         $supervisor->loop();
@@ -504,6 +507,7 @@ class SupervisorTest extends IntegrationTest
     {
         SystemProcessCounter::$command = 'worker.php';
         $this->supervisor = $supervisor = new Supervisor($options = $this->supervisorOptions());
+        $options->balance = 'simple';
 
         $supervisor->scale(3);
 
--- a/tests/Feature/SupervisorTest.php
+++ b/tests/Feature/SupervisorTest.php
@@ -95,6 +95,29 @@ class SupervisorTest extends IntegrationTest
         );
     }
 
+    public function test_supervisor_starts_one_pool_when_single_balancing()
+    {
+        $options = $this->supervisorOptions();
+        $options->balance = 'single';
+        $options->queue = 'first,second';
+        $this->supervisor = $supervisor = new Supervisor($options);
+
+        $supervisor->scale(2);
+        $this->assertCount(2, $supervisor->processes());
+
+        $host = MasterSupervisor::name();
+
+        $this->assertSame(
+            'exec '.$this->phpBinary.' worker.php redis --name=default --supervisor='.$host.':name --backoff=0 --max-time=0 --max-jobs=0 --memory=128 --queue="first,second" --sleep=3 --timeout=60 --tries=0 --rest=0',
+            $supervisor->processes()[0]->getCommandLine()
+        );
+
+        $this->assertSame(
+            'exec '.$this->phpBinary.' worker.php redis --name=default --supervisor='.$host.':name --backoff=0 --max-time=0 --max-jobs=0 --memory=128 --queue="first,second" --sleep=3 --timeout=60 --tries=0 --rest=0',
+            $supervisor->processes()[1]->getCommandLine()
+        );
+    }
+
     public function test_recent_jobs_are_correctly_maintained()
     {
         $id = Queue::push(new Jobs\BasicJob);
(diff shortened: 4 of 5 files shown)


fabriciojs:
👍🏻  this is nice, as this way we don't get bloated with too many workers, but we can still make our queues more granular for straight visibility into each of them.

taylorotwell:
What is the main difference between this and `'balance' => false`?

dbpolito:
Well, `off` won't auto scale, right? It will basically stick with the process or max process... This will scale based on the jobs, but each worker will have all queues, so it will scale up to x and each worker with `queue=a,b,c`

taylorotwell:
Gotcha. I do wonder if we really need a new strategy for this if `off` behaves similar to this but just using maxProcesses. You have to build your server to be able to handle your queue worker load at maxProcesses anyways so why not just let it run at max all the time using the `off` strategy?

dbpolito:
Well while this is true, there are several benefits of not spending the memory just to run the workers unnecessarily:

1. Reserving all this memory just to keep seeking for jobs (~100mb per worker)
2. Having more free memory for the jobs to process
3. Scaling Horizon Instances on K8S based on resources usage

Also this brings the functionality of queue priority, so it needs to process on the order.

I think the complexity added vs functionality might worth to more people.

I'm already using this strategy on my project and only changing to it reduced 30% of memory usage in my case... Of course it will change depending on how your queues are designed.

----

Another thing that would help on this matter which was discussed on other issues is starting workers on `min` instead of `max`, to avoid unnecessary memory peak / waste.

taylorotwell:
Should this just be the default behavior of `balance => false`? I don't think the name `single` really conveys much.

dbpolito:
Well, that is what i expected to be TBH... i agree, but isn't it a "BREAKING CHANGE"? Won't break anything but will change the behavior for everyone who uses it.

I'm totally ok with that, or come up with a different name...

taylorotwell:
I'm not sure if it's a breaking change. We never actually document the scaling behavior of `balance => false` and it shouldn't "break" your application as it's mainly an internal optimization for Horizon.

dbpolito:
well makes sense... i will try to change the PR later today or tomorrow tops.

taylorotwell:
Thanks! Just mark as ready for review when you want me to take another look.

dbpolito:
@taylorotwell updated 👍

taylorotwell:
Thanks

ChristopheBorcard:
It might have been wise to rename the PR.

Because now the change log on the Tag is talking about a new Single balance strategy that doesn't exist:

https://github.com/laravel/horizon/releases/tag/v5.26.0

dbpolito:
I renamed the PR but it won't update the release i think

driesvints:
I updated all references. Thanks

vilhelmjosander:
A bit late to the party here but this seems to have broken my use of Horizon with balance = false.

I have a fairly large application which sometimes handles upwards of 200k jobs per hour on a single VPS. After updating Horizon to >= 5.26.0 it suddenly stopped processing jobs (or processed them EXTREMELY slow). Downgrading to 5.25.0 and everything is back to normal again.

This is my setup:

```php
 'environments' => [
        'production' => [
            'supervisor-live' => [
                'connection' => 'redis-live-battles',
                'queue' => ['live'],
                'balance' => 'false',
                'minProcesses' => 1,
                'maxProcesses' => 5,
                'balanceMaxShift' => 15,
                'balanceCooldown' => 1,
                'tries' => 3,
                'timeout' => 80,
            ],
            'supervisor-default' => [
                'connection' => 'redis',
                'queue' => ['urgent', 'high', 'default', 'low'],
                'balance' => 'false',
                'minProcesses' => 1,
                'maxProcesses' => 50,
                'balanceMaxShift' => 15,
                'balanceCooldown' => 1,
                'tries' => 3,
                'timeout' => 80,
            ],
        ],
```
Unfortunately I am not familiar enough with the internals of Horizon to be able to debug this myself but i've narrowed it down to this PR.

This is what my horizon dashboard looks like after deploying >= 5.26.0. Basically jobs just keep piling up without being processed across the different queues.

CleanShot 2024-08-23 at 22 41 30@2x

dbpolito:
@vilhelmjosander 

Well, i also got a fairly big project and haven't noticed any difference on performance, processing ~15k jobs as usual...

Did you check the job throughput? Is that different on versions?

TBH i can't see a reason to impact performance, the only behavior difference is that it scales now... before it would always be at 50 process in your case... now it will scale from 1 to 50... but the way it process and how it process is exactly the same.

Looking at your screenshot, it seems it scaled properly and it's using 50 processes...

I would look at the job throughput / runtime... as maybe the jobs may be different.

trevorgehman:
Just to add another comment. This affected our Horizon workers because we had an old config file where only `processes` was specified, instead of `minProcesses` and `maxProcesses`. Here is our config:

```
'supervisor-1' => [
    'connection' => 'redis',
    'queue' => ['critical'],
    'balance' => false,
    'processes' => 6,
    'tries' => 10,
    'sleep' => 2,
    'timeout' => 3500,
    'memory' => 1500,
    'maxJobs' => 1000,
],
```

Horizon treats `processes` as `maxProcesses`. And since we didn't have a value `minProcesses` it defaulted to 1.

https://github.com/laravel/horizon/blob/4d021d2adef966433b018ff67c718ceab7209076/src/ProvisioningPlan.php#L174

So basically it enabled scaling for all these workers which previously didn't scale.

It didn't break anything for us, just took us a couple weeks to notice that our queues processing jobs a bit slower.

sts-ryan-holton:
Just jumping in here, I tried out setting `'balance' => false` for my supervisors in my Horizon config, not completely sure what the benefit here would be? Does this technically mean it processes each queue synchronously, so each queue in order first? What's the benefits?

vilhelmjosander:
I'm at a loss at what to do here. As soon as I upgrade to >= 5.26.0 horizon basically stops processing jobs with balance = false. It works flawlessly with 5.25.0.  

Currently this also stops me from updating laravel/framework since horizon 5.25.0 gives the following errors with Laravel >= 11.27.0:  

Declaration of Laravel\Horizon\RedisQueue::pop must be compatible with Illuminate\Queue\RedisQueue::pop. 

Any ideas? :)

sts-ryan-holton:
@vilhelmjosander A few things to consider:

1. Not sure if it's causing you a problem, but it should be `'balance' => false` instead of `'balance' => 'false'`. From your screenshot, your "Balancing" column says "False", but when I use `'balance' => false` _(note the boolean not string)_ then under "Balancing" it'll say "Disabled", so check this
2. After looking at your screenshot, your queues, `urgent`, `high`, `default` and `low` in that order, I'm pretty sure it'll try to process everything on the `urgent` queue first, then move on to the `high` queue, then `default`, then `low`. This actually means that, if for example you have a tonne of jobs coming in on the `urgent` queue, or `high` queue, then any queue later on simply will just build up and never really process anything unless you allocate more workers/processes.

I noticed this in my own platform, to try to illustrate it better, if you have 5 workers and each job takes 1 second and you have new jobs filling the urgent queue as quick as they're being processed, no queues below it will have enough workers.

Does this help?

dbpolito:
@vilhelmjosander it's hard to help without a way to reproduce...

I tried to reproduce here... played with the tests, and everything seems to be working as it should... 

Can you check the processes that are running? with `ps aux` or something like that, to check how all those 50 processes looks? You should have 50 processes with `... --queue="urgent,high,default,low" ...`

And again, are you sure it's not processing? Because as stated above, it will process on this queue order... so all urgent first, then high, then default, then low...

None of the queues are processed? or only the `urgent` ones?

Of course we need to fix it, if something is wrong... but have you tried changing the balance? Maybe for your use case you do need to change to simple or auto

vilhelmjosander:
@dbpolito 

At once after deploying >= 5.26.0 it starts to process the queues extremely slow. I now tried to change the setting to false (boolean) instead of the string 'false' (which works fine on 5.25.0).
CleanShot 2024-10-19 at 20 34 39@2x

This is the only thing I see which is related to horizon with `ps aux`. Same for `ps aux | grep queue` or similar. I do not have 50 processes (which I do have with 5.25.0).

CleanShot 2024-10-19 at 20 42 16@2x

I did note that after deploying >= 5.26.0 `ps aux` did show 20-50 processes with `... --queue="urgent,high,default,low" ...` for a while, then it stopped and no processes at all were visible which basically meant that jobs were not processed.

Here is my current config:

CleanShot 2024-10-19 at 20 43 49@2x

Switching to balance = 'auto' seems to work. The reason why I've had balance = false is that I do want the default laravel behaviour where the jobs are processed in the queue order. It is important that the urgent queue has higher priority than the low queue and always processes asap.

dbpolito:
@vilhelmjosander 

the only thing i noticed which shouldn't impact anything, but who knows... is these supervisors uses different redis connections...

so when you run it you only have the supervisor processes running, and no actual worker running? that's weird... as horizon doesn't support scaling to 0 workers...

are you sure nothing pops into your logs? errors or anything?

Really hard to help, i tried changing tests, trying to reproduce but not a lot of luck.

Maybe you can setup a reproducible repo so we can jump in

SteveEdson:
I have the same issue, my logs are flooded with:

```
[2024-11-01 12:13:29] production.ERROR: Declaration of Laravel\Horizon\RedisQueue::pop($queue = null) must be compatible with Illuminate\Queue\RedisQueue::pop($queue = null, $index = 0) {"exception":"[object] (Symfony\\Component\\ErrorHandler\\Error\\FatalError(code: 0): Declaration of Laravel\\Horizon\\RedisQueue::pop($queue = null) must be compatible with Illuminate\\Queue\\RedisQueue::pop($queue = null, $index = 0) at /home/forge/shop.spacest.ar/vendor/laravel/horizon/src/RedisQueue.php:131)
[stacktrace]
#0 {main}
"}
```

Config looks like this:

```php
<?php

use Illuminate\Support\Str;

return [

    /*
    |--------------------------------------------------------------------------
    | Horizon Domain
    |--------------------------------------------------------------------------
    |
    | This is the subdomain where Horizon will be accessible from. If this
    | setting is null, Horizon will reside under the same domain as the
    | application. Otherwise, this value will serve as the subdomain.
    |
    */

    'domain' => null,

    /*
    |--------------------------------------------------------------------------
    | Horizon Path
    |--------------------------------------------------------------------------
    |
    | This is the URI path where Horizon will be accessible from. Feel free
    | to change this path to anything you like. Note that the URI will not
    | affect the paths of its internal API that aren't exposed to users.
    |
    */

    'path' => 'horizon',

    /*
    |--------------------------------------------------------------------------
    | Horizon Redis Connection
    |--------------------------------------------------------------------------
    |
    | This is the name of the Redis connection where Horizon will store the
    | meta information required for it to function. It includes the list
    | of supervisors, failed jobs, job metrics, and other information.
    |
    */

    'use' => 'default',

    /*
    |--------------------------------------------------------------------------
    | Horizon Redis Prefix
    |--------------------------------------------------------------------------
    |
    | This prefix will be used when storing all Horizon data in Redis. You
    | may modify the prefix when you are running multiple installations
    | of Horizon on the same server so that they don't have problems.
    |
    */

    'prefix' => env(
        'HORIZON_PREFIX',
        Str::slug(env('APP_NAME', 'laravel'), '_').'_horizon:'
    ),

    /*
    |--------------------------------------------------------------------------
    | Horizon Route Middleware
    |--------------------------------------------------------------------------
    |
    | These middleware will get attached onto each Horizon route, giving you
    | the chance to add your own middleware to this list or change any of
    | the existing middleware. Or, you can simply stick with this list.
    |
    */

    'middleware' => ['web'],

    /*
    |--------------------------------------------------------------------------
    | Queue Wait Time Thresholds
    |--------------------------------------------------------------------------
    |
    | This option allows you to configure when the LongWaitDetected event
    | will be fired. Every connection / queue combination may have its
    | own, unique threshold (in seconds) before this event is fired.
    |
    */

    'waits' => [
        'redis:default' => 60,
    ],

    /*
    |--------------------------------------------------------------------------
    | Job Trimming Times
    |--------------------------------------------------------------------------
    |
    | Here you can configure for how long (in minutes) you desire Horizon to
    | persist the recent and failed jobs. Typically, recent jobs are kept
    | for one hour while all failed jobs are stored for an entire week.
    |
    */

    'trim' => [
        'recent' => 10080,
        'pending' => 10080,
        'completed' => 10080,
        'recent_failed' => 10080 * 2,
        'failed' => 10080 * 2,
        'monitored' => 10080 * 2,
    ],

    /*
    |--------------------------------------------------------------------------
    | Metrics
    |--------------------------------------------------------------------------
    |
    | Here you can configure how many snapshots should be kept to display in
    | the metrics graph. This will get used in combination with Horizon's
    | `horizon:snapshot` schedule to define how long to retain metrics.
    |
    */

    'metrics' => [
        'trim_snapshots' => [
            'job' => 24,
            'queue' => 24,
        ],
    ],

    /*
    |--------------------------------------------------------------------------
    | Fast Termination
    |--------------------------------------------------------------------------
    |
    | When this option is enabled, Horizon's "terminate" command will not
    | wait on all of the workers to terminate unless the --wait option
    | is provided. Fast termination can shorten deployment delay by
    | allowing a new instance of Horizon to start while the last
    | instance will continue to terminate each of its workers.
    |
    */

    'fast_termination' => false,

    /*
    |--------------------------------------------------------------------------
    | Memory Limit (MB)
    |--------------------------------------------------------------------------
    |
    | This value describes the maximum amount of memory the Horizon master
    | supervisor may consume before it is terminated and restarted. For
    | configuring these limits on your workers, see the next section.
    |
    */

    'memory_limit' => 64,

    /*
    |--------------------------------------------------------------------------
    | Queue Worker Configuration
    |--------------------------------------------------------------------------
    |
    | Here you may define the queue worker settings used by your application
    | in all environments. These supervisors and settings handle all your
    | queued jobs and will be provisioned by Horizon during deployment.
    |
    */

    'defaults' => [
        'supervisor-1' => [
            'connection' => 'redis',
            'queue' => ['default'],
            'balance' => 'auto',
            'maxProcesses' => 1,
            'memory' => 128,
            'tries' => 1,
            'nice' => 0,
        ],
    ],

    'environments' => [
        'production' => [
            'supervisor-1' => [
                'maxProcesses' => 10,
                'balanceMaxShift' => 1,
                'balanceCooldown' => 3,
            ],
            'mailcoach-general' => [
                'connection' => 'mailcoach-redis',
                'queue' => ['mailcoach', 'mailcoach-feedback', 'send-mail', 'send-automation-mail'],
                'balance' => 'auto',
                'processes' => 10,
                'tries' => 2,
                'timeout' => 60 * 60,
            ],
            'mailcoach-heavy' => [
                'connection' => 'mailcoach-redis',
                'queue' => ['send-campaign'],
                'balance' => 'auto',
                'processes' => 3,
                'tries' => 1,
                'timeout' => 60 * 60,
            ],
        ],

        'local' => [
            'supervisor-1' => [
                'maxProcesses' => 3,
            ],
            'mailcoach-general' => [
                'connection' => 'mailcoach-redis',
                'queue' => ['mailcoach', 'mailcoach-feedback', 'send-mail', 'send-automation-mail'],
                'balance' => 'auto',
                'processes' => 10,
                'tries' => 2,
                'timeout' => 60 * 60,
            ],
            'mailcoach-heavy' => [
                'connection' => 'mailcoach-redis',
                'queue' => ['send-campaign'],
                'balance' => 'auto',
                'processes' => 3,
                'tries' => 1,
                'timeout' => 60 * 60,
            ],
        ],
    ],
];

```

dbpolito:
I faced a similar issue you guys are having and sent a fix: https://github.com/laravel/horizon/pull/1566

Not related to this PR but this PR makes this issue more noticeable.

PR #1481, opened 2024-07-25
Extend Slack Notification to work with Slack Apps for Laravel 11 support

travisricks:
Reference: https://github.com/laravel/framework/discussions/51879

Laravel 11 supports Slack notifications with not only webhook urls but also using a Slack app, but this repository is configured only for webhook urls. Added a new condition to check if the supplied destination is a webhook url or channel id and return the appropriate response.

Screenshot-003236-2024-07-25at22 18 01@2x

### DIFF
--- a/src/Notifications/LongWaitDetected.php
+++ b/src/Notifications/LongWaitDetected.php
@@ -94,9 +94,10 @@ class LongWaitDetected extends Notification
     public function toSlack($notifiable)
     {
         $fromName = 'Laravel Horizon';
-        $imageUrl = 'https://laravel.com/assets/img/horizon-48px.png';
-        $text = 'Oh no! Something needs your attention.';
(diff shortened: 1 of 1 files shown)


driesvints:
We'll need to fix the static analysis tests before proceeding.

travisricks:
@driesvints 
I (think) I fixed the php stan issues by using `// @phpstan-ignore-line` on line 114. I only did this because the file was already using that previously for a different type hint issue.

If I understand the issue correctly it's because the type-hinted class only exists in Laravel 11. I tried adding a safety check for that by adding `class_exists()` checks in the if statement.

But please let me know if this is not the correct solution or if other changes are necessary.

siarheipashkevich:
@travisricks what about updating the documentation for this ability?

PR #1482, opened 2024-07-28
Add missing keys that should be cleared when running `horizon:clear-metrics`

atendwa:
This update enhances the `clear()` method to include additional Redis keys, ensuring that the `horizon:clear-metrics` command comprehensively removes all relevant metrics stored in the cache.

#### Dashboard State Comparison
- Before Executing `horizon:clear-metrics`:
Screenshot from 2024-07-28 08-37-56

- After Executing `horizon:clear-metrics` (Current Version):
Screenshot from 2024-07-28 08-38-17

- After Updating the `clear()` Method:
Screenshot from 2024-07-28 08-53-01

## Note on Job Hashes in Cache
Currently, `job hashes` remain in the cache, as illustrated below. Ideally, these should also be cleared. However, removing them without flushing the entire Redis queue connection or using wildcard patterns `('*')` is not feasible unless the specific keys to target are known.

A potential improvement would be to store job hashes under a specific key, such as `jobs`, to facilitate direct targeting.

Screenshot from 2024-07-28 09-10-00

### DIFF
--- a/src/Repositories/RedisMetricsRepository.php
+++ b/src/Repositories/RedisMetricsRepository.php
@@ -384,8 +384,13 @@ class RedisMetricsRepository implements MetricsRepository
         $this->forget('measured_jobs');
         $this->forget('measured_queues');
         $this->forget('metrics:snapshot');
+        $this->forget('completed_jobs');
+        $this->forget('recent_jobs');
+        $this->forget('failed_jobs');
+        $this->forget('recent_failed_jobs');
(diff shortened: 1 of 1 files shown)


taylorotwell:
I don't think these keys are meant to be cleared by this command. Metrics is separate from the dashboard.

atendwa:
Is there a command that will achieve this use case?

PR #1505, opened 2024-10-08
quick fix to sync get rid of the  The "json" option does not exist warning 

webpatser:
After upgrading to the latest Framework 11.27 the queue worker introduced a json option. Horizon does't have that option so the worker won't start and starts complaining about `The "json" option does not exist.  `

This is a quick fix, since I don't know the grans scheme of this new option related to horizon.


taylorotwell:
Should be fixed by https://github.com/laravel/framework/pull/53072

Issue #1475, opened 2024-07-17
Ability to customise notifications


lukemorcom:
We are running multiple test environments for our app, and only one Slack channel, to avoid having tons of channels. It would be great to be able to customise the notifications we receive, so we can include the server name in the message that's sent. I imagine this would be useful for other notification channels too. Anyone know if it's possible? Looking at the source it doesn't seem so

driesvints:
Thanks. Right now we don't have plans to work on this. Feel free to attempt a PR to achieve what you need.

PR #1518, opened 2024-11-29
Add support for notification class overrides

lukemorcom:
Back in July I raised issue #1475, because my team wanted to be able to customise the notification that was sent when a long wait occurs. I was told to attempt a PR and this is certainly... an attempt. Please be nice, this is my first ever PR into an open source project.

**Usage**
```
// Pass a map of $event => $notification into the function
Horizon::overrideNotifications([
    LongWaitDetected::class => MyCustomNotification::class
]);
```

**Considerations**
- Currently Horizon has only one notification class (`LongWaitDetected`), but this implementation is designed with future extensibility in mind. Using a map provides flexibility for additional notifications if they are introduced, instead of being limited to a single notification override.
- Any future Horizon events would also need to include similar logic to `LongWaitDetected::toNotification()` to determine whether to construct a custom notification. I considered alternative, more abstract approaches, but they would involve more significant changes to the codebase.
- For compatibility, custom notification classes must extend the original notification associated with the event.

I am quite expecting this PR to be declined, but if that is so, I would greatly appreciate any feedback - as mentioned above this is my first contribution to an open source project so your feedback/advice is genuinely valuable to me. Thanks.

### DIFF
--- a/src/Contracts/LongWaitDetectedNotification.php
+++ b/src/Contracts/LongWaitDetectedNotification.php
@@ -1,0 +1,8 @@
+<?php
+
+namespace Laravel\Horizon\Contracts;
+
+interface LongWaitDetectedNotification
+{
+    //
+}
--- a/src/Events/LongWaitDetected.php
+++ b/src/Events/LongWaitDetected.php
@@ -2,8 +2,8 @@
 
 namespace Laravel\Horizon\Events;
 
-use Laravel\Horizon\Horizon;
-use Laravel\Horizon\Notifications\LongWaitDetected as LongWaitDetectedNotification;
+use Illuminate\Container\Container;
+use Laravel\Horizon\Contracts\LongWaitDetectedNotification;
 
 class LongWaitDetected
 {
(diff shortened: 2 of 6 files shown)


taylorotwell:
@lukemorcom thanks for the PR! I modified this a bit to just use the container and a new interface that you can override with your own binding.

siarheipashkevich:
@lukemorcom what about the documentation for this feature?

jhm-ciberman:
@siarheipashkevich If you see the PR, Taylor did change it so it uses the container instead of a custom method. So I don't think you need extra documentation besides the existing docs for the IoC container. If someone is looking for something this specific they will see that in the source code Horizon uses the container and they will know that they can rebind the class.

muarachmann:
Hi @lukemorcom @taylorotwell I see this is not in the docs ? Any idea on how to use this ? 
Regards,

lukemorcom:
> Hi @lukemorcom @taylorotwell I see this is not in the docs ? Any idea on how to use this ? Regards,

It is using the container. You just need to make a notification class of your own that implements `Laravel\Horizon\Contracts\LongWaitDetectedNotification` and tell the container to use yours:

``` php
// CustomLongWaitDetectedNotification.php
use Laravel\Horizon\Contracts\LongWaitDetectedNotification;

class CustomLongWaitNotification implements LongWaitDetectedNotification
{
  //
}


// MyServiceProvider.php
use Laravel\Horizon\Contracts\LongWaitDetectedNotification;
use MyApp\CustomLongWaitNotification;

public function register()
{
  //
  $this->app->bind(
    LongWaitDetectedNotification::class,
    CustomLongWaitNotification::class,
  );
  //
}
```

Issue #1536, opened 2025-02-25
Horizon Spawning `maxProcesses` on Startup Instead of Scaling Dynamically


andre-hoffmann:
### Horizon Version

5.29.2

### Laravel Version

11.33.2

### PHP Version

8.3.6

### Redis Driver

PhpRedis

### Redis Version

6.0.2

### Database Driver & Version

MySQL 8

### Description

When Laravel Horizon starts, it immediately spawns the number of workers defined by `maxProcesses`, rather than starting with `minProcesses` and scaling up as needed. This contradicts the expected auto-scaling behavior, where Horizon should dynamically adjust the number of workers based on queue workload.

Horizon should begin with `minProcesses` and increase workers as necessary. However, testing shows that Horizon always launches with `maxProcesses` from the start, leading to unnecessary resource consumption and potential inefficiencies.

Additionally, when using auto compute scaling, this behavior results in unnecessary scaling, increasing infrastructure costs.

### Steps To Reproduce

1. Configure Horizon with the following settings in `horizon.php`:
   ```php
   'environments' => [
       'production' => [
           'supervisor-1' => [
               'connection' => 'redis',
               'queue' => ['default'],
               'balance' => 'auto',
               'minProcesses' => 1,
               'maxProcesses' => 5,
               'tries' => 3,
           ],
       ],
   ],
   ```

2. Start Horizon using:
   ```sh
   php artisan horizon
   ```

3. Observe the number of worker processes started.

### Observed Behavior
- Horizon immediately spawns 5 worker processes (matching `maxProcesses`), even when the queue is empty or under low load.

### Expected Behavior
- Horizon should start with `minProcesses` (1 in this case) and scale up dynamically as the queue workload increases.

### Pull Request
[#1537](https://github.com/laravel/horizon/pull/1537)

andre-hoffmann:
Closing as the PR was merged :)

PR #1537, opened 2025-02-26
Scale workers from min instead of max on startup

andre-hoffmann:
When Laravel Horizon starts, it currently spawns the maximum number of workers defined by `maxProcesses`, rather than starting with `minProcesses` and scaling up as needed.  

With this change, Horizon will now start with the **minimum** number of workers and scale up dynamically as required. This improvement is especially beneficial for users utilizing **auto compute scaling**, as it eliminates unnecessary scaling, leading to **reduced infrastructure costs**.  

This change addresses issue [#1536](https://github.com/laravel/horizon/issues/1536).

### DIFF
--- a/src/Console/SupervisorCommand.php
+++ b/src/Console/SupervisorCommand.php
@@ -97,7 +97,7 @@ class SupervisorCommand extends Command
         $supervisor->working = ! $this->option('paused');
 
         $supervisor->scale(max(
-            0, $this->option('max-processes') - $supervisor->totalSystemProcessCount()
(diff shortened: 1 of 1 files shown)


taylorotwell:
I wonder if the correct behavior is somewhere in the middle?

andre-hoffmann:
@taylorotwell The issue still remains that if `maxProcesses` is set to 100, it will still spawn 50 on startup.  

In some cases, this might still be too high, but it's already a lot better. I’ve updated the PR to use the middle value.

taylorotwell:
Will give it a shot. 🫡

PR #1538, opened 2025-03-04
Add `proxy_path` config for subfolder usage in Horizon's Vue front-end

andre-hoffmann:
**Description**

This pull request adds a new config key `proxy_path` to Laravel Horizon, allowing developers to serve Horizon behind a reverse proxy in a subfolder (e.g. `example.com/some-prefix/horizon`) without impacting existing setups that rely on the default `path`. This new config ensures that Horizon’s Vue front-end correctly references the subfolder path for API calls and other resources.

**Key Changes**:
- Introduces a `proxy_path` configuration to `config/horizon.php`.
- Ensures front-end calls check `proxy_path` so subfolder routing works properly behind a proxy.
- Maintains backward compatibility by leaving the existing `path` configuration untouched if `proxy_path` is empty.

With this change, anyone who needs a subfolder URL behind a proxy (e.g. `example.com/laravel/horizon`) can specify `proxy_path` accordingly, avoiding the need for additional rewrites or patchwork solutions—all while preserving the default Horizon behavior for other installations.

### DIFF
--- a/config/horizon.php
+++ b/config/horizon.php
@@ -32,6 +32,19 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Horizon Proxy Path
+    |--------------------------------------------------------------------------
+    |
+    | This value is the base path where Horizon is accessible when behind a
+    | reverse proxy. For example, when your application is accessed at
+    | example.com/app and Horizon at example.com/app/horizon, set this to '/app'.
+    |
+    */
+
+    'proxy_path' => env('HORIZON_PROXY_PATH', ''),
+
+    /*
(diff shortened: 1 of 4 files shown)


andre-hoffmann:
@taylorotwell

First of all, thank you for approving the PR!

I have a quick question—I noticed that you updated horizon.php to remove the new config variable. How would someone know that this functionality is available?

I just want to ensure that my next PR is as polished as possible.

RikkiMasters:
Thanks, I've needed this since time began. Works great.

I agree that it needs to be in Horizon.php by default

PR #1552, opened 2025-05-08
Set AutoScaler logic to target process scaling based on queue size wh…

nick-potts:
With 'balance' set to false, currently the autoscaler will target maxProcesses when there is >0 jobs in the queue, causing it to almost always ramp upwards.

### DIFF
--- a/src/AutoScaler.php
+++ b/src/AutoScaler.php
@@ -106,6 +106,15 @@ class AutoScaler
         $totalJobs = $queues->sum('size');
 
         return $queues->mapWithKeys(function ($timeToClear, $queue) use ($supervisor, $timeToClearAll, $totalJobs) {
+            if (! $supervisor->options->balancing()) {
+                $targetProcesses = min(
+                    $supervisor->options->maxProcesses,
+                    max($supervisor->options->minProcesses, $timeToClear['size'])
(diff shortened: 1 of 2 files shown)


taylorotwell:
Can you provide a more thorough description?

nick-potts:
When `balance = false`, currently horizon tries to apply autoscaling logic but its quite broken due to the fact the autoscaling logic doesn't really work when there's a single queue (which is what balance = false does).

The `numberOfWorkersPerQueue` tries to compute the recommended amount of processes, but with `balance = false`:
- When `$timeToClearAll > 0`, it always returns `maxProcesses`
- When `$timeToClearAll == 0` it returns `maxProcesses` if there's >0 jobs, and `minProcesses` if there's 0 jobs.

This means if you have an average of 1 job in the queue at all times, it'll still try scale to maxProcesses at all times. It'll only ever scale down if the queue is completely empty. On laravel cloud this just means my replicas autoscale to the max, even if there's only a very very light workload.

The new logic takes into account the amount of jobs in the queue, and returns a reasonable number (the min of maxProcess and the size of the queue, since its a single queue) instead of the min or max, meaning horizon just scales properly. It only affects `balance = false`.

PR #1556, opened 2025-05-18
feat: add Poll component to standardize polling behavior

mathiasgrimm:
- Centralizes interval-based logic in reusable Vue component
- Reduces polling frequency to hourly when window not visible
- Improves code maintainability and reduces redundancy

This significantly reduces the bandwidth usage when the browser tab is not active.


resources/js/components/Poll.vue
```diff
@@ -0,0 +1,111 @@
+<script type="text/ecmascript-6">
```
bert-w: I'd rewrite this as a plain javascript function and not a vue component. Now every polling function needs an associated html element in the DOM.

mathiasgrimm: This is a renderless component, so it does not create any DOM elements.

bert-w: You now specifically need to add `<poll @poll="refreshBatchesPeriodically" />` in the code everywhere instead of a plain javascript solution like `myCustomIntervalFunctionThatPausesInBackground(() => ..., 2000)`.

mathiasgrimm: What about the Lifecycle hooks?

bert-w: yes you'd still need the cleanup process; one such implementation could look like this:

```js
import VisibilityAwareInterval from './VisibilityAwareInterval';

data() {
    return {
        this.my_interval: undefined,
    };
},
mounted() {
    // Create some visibility based interval:
    this.my_interval = new VisibilityAwareInterval(cb, 2000);
},
beforeUnmount() {
    // Call the `clear` fn in the custom class (optionally use `?.` here in 
    // case the `my_interval` is already cleared by other code):
    this.my_interval?.clear();
}
```

The `VisibilityAwareInterval` class will mostly contain the same code you have now, but instead it is a javascript class that exposes a few functions like `start()/stop()/clear()`.

mathiasgrimm: Thanks @bert-w  for the follow-up.

I don't see any advantage to that approach since the existing code before my PR already handles it similarly.

With this PR, the implementation becomes much simpler:
```html
<poll @poll="myCallback" />
```

The goal is to centralize polling logic and make it effortless to use by abstracting all complexity into this renderless component - similar to how `wire:poll` works in Livewire. This way, developers don't need to worry about cleanup, initialization, or other implementation details.

PR #1562, opened 2025-06-01
feat: Add `horizon_stats()` helper function

saaiful:
**Description:**
This PR introduces a `horizon_stats()` helper function to programmatically retrieve Laravel Horizon's status and statistics. This is useful for displaying Horizon data in custom application dashboards or for other monitoring purposes, especially when `php artisan horizon:status` is not directly usable (e.g., within web requests or when integrating with tools like Totem).
The function internally uses Horizon's `DashboardStatsController` to fetch the data, providing a simple way to access metrics like job counts, process status, and queue performance.
**Example Usage:**
```php
$stats = horizon_stats();
// $stats now holds an array of Horizon statistics.
{
    "failedJobs": 10,
    "jobsPerMinute": 1,
    "status": "running",
    // ... and other stats
}
```
This helper simplifies integrating Horizon monitoring into Laravel applications.

### DIFF
--- a/src/helpers.php
+++ b/src/helpers.php
@@ -1,6 +1,6 @@
 <?php
 
-if (function_exists('horizon_stats')) {
+if (!function_exists('horizon_stats')) {
     function horizon_stats()
     {
         $dashboard = new \Laravel\Horizon\Http\Controllers\DashboardStatsController();
--- a/src/helpers.php
+++ b/src/helpers.php
@@ -1,0 +1,9 @@
+<?php
+
+if (function_exists('horizon_stats')) {
+    function horizon_stats()
+    {
+        $dashboard = new \Laravel\Horizon\Http\Controllers\DashboardStatsController();
+        return $dashboard->index();
+    }
+}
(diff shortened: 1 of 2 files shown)


src/helpers.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+if (function_exists('horizon_stats')) {
```
devajmeireles: This is wrong. The correct here is to check if this function was not declared, so you need to use a `!` here.

src/helpers.php
```diff
@@ -0,0 +1,9 @@
...
+    function horizon_stats()
+    {
+        $dashboard = new \Laravel\Horizon\Http\Controllers\DashboardStatsController();
```
devajmeireles, line 6: While I think this is a nice addition, I strongly believe that the correct way to do this is the other way around: move the current logic from the `index` method to a place where it can be reused in the controller method as well as here.

devajmeireles (commented):
I don't know if it will be accepted by Taylor, but I think this is cool. Also, this leads us to a path where we can add a new command that visualizes the data via terminal, maybe?

taylorotwell:
Overall idea is fine but not sure this is how I would implement it.

saaiful:
> Overall idea is fine but not sure this is how I would implement it.

Made this pull request just to request this feature, implement it any way you see fit.

PR #1563, opened 2025-06-02
[5.x] Allow to specify job namespace in horizon:clear command

amirhshokri:
This pull request updates the `horizon:clear` command to allow specifying the job namespace as an argument.

Example usage:
`php artisan horizon:clear --queue=emails --job=App\\Jobs\\TestJob`

Why is it useful?
This enhancement provides more granular control when clearing jobs from specific queue. By allowing the `--job` option, developers can target and clear jobs of a particular class within a given queue, without affecting other jobs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /phpunit.xml
 composer.lock
 .phpunit.result.cache
+.idea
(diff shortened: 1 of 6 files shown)


taylorotwell:

PR #1564, opened 2025-06-05
Allowed Horizon Status & Supervisors commands to run outside console

abisheknaagendra:
This helps if you are checking Horizon status & Supervisors' status using HTTP requests, for example using: https://crondog.io/


devajmeireles:
> This helps if you are checking Horizon status & Supervisors' status using HTTP requests, for example using: https://crondog.io/

What is `crondog.io`? I tried to hit on it, but it seems it's unavailable?

PR #1565, opened 2025-06-06
add support for pushing to multiple connections with different redis …

henzeb:
### Problem

In setups with multiple Horizon instances using the same codebase but different queue connections (each pointing to different Redis instances), jobs dispatched to different connections are incorrectly tracked. 
When a job is pushed to a connection, it appears as pending on the wrong Horizon dashboard because job monitoring data is always stored in the Redis connection it was pushed from, rather than the Redis connection of the queue connection the job was pushed to. This causes other Horizon instances monitoring different connections to be unable to show the completed or failed jobs for jobs that were dispatched to their connections.

 ### Solution

This PR fixes the issue by ensuring that job monitoring data (tags, job references) is stored in the correct Redis instance based on the connection the job was pushed to.

 ### Changes

  - **Enhanced Tag Repository**: Modified `RedisTagRepository` to accept connection parameter and resolve the correct Redis connection for storing monitored tags
  - **Updated Job Repository**: Modified `RedisJobRepository::pushed()` to use the job's target connection for storing job references
  - **Fixed Event Handling**: Updated `StoreMonitoredTags` listener to pass the target connection name when storing monitored job tags
  - **Connection Resolution**: Added logic to resolve Redis connections from queue connection configuration with fallback to default

  ### Result

  Jobs are now properly tracked on the correct Horizon instance based on their target connection, allowing each instance to accurately display job statuses for jobs dispatched to their respective Redis instances.

## Why The Test Configuration Changes Don't Interfere With Logic Testing

  The test configuration lines (`config(['queue.connections.redis.connection' => null])`) address a test environment limitation where the `getEnvironmentSetUp()` configuration isn't inherited by spawned worker processes.

  **The Issue**: The `getEnvironmentSetUp()` method sets `queue.connections.redis.connection = 'horizon'` for the test process, but spawned worker processes don't inherit this configuration and use Laravel's default setup instead.

  **The Fix**: Setting the connection to `null` in the specific tests disables the multi-connection logic entirely, forcing both the test process and worker processes to use the same fallback behavior (default 'horizon'
  connection).

  **Logic Preserved**: The core multi-connection functionality remains fully intact and tested in dedicated multi-connection tests. These configuration changes only ensure test environment consistency in supervisor tests that
  spawn separate PHP processes - in production, all processes naturally share the same application configuration.

### DIFF
--- a/src/Listeners/StoreMonitoredTags.php
+++ b/src/Listeners/StoreMonitoredTags.php
@@ -33,10 +33,10 @@ class StoreMonitoredTags
      */
     public function handle(JobPushed $event)
     {
-        $monitoring = $this->tags->monitored($event->payload->tags());
+        $monitoring = $this->tags->monitored($event->payload->tags(), $event->connectionName);
 
         if (! empty($monitoring)) {
-            $this->tags->add($event->payload->id(), $monitoring);
+            $this->tags->add($event->payload->id(), $monitoring, $event->connectionName);
         }
     }
 }
--- a/src/Repositories/RedisJobRepository.php
+++ b/src/Repositories/RedisJobRepository.php
@@ -335,7 +335,7 @@ class RedisJobRepository implements JobRepository
      */
     public function pushed($connection, $queue, JobPayload $payload)
     {
-        $this->connection()->pipeline(function ($pipe) use ($connection, $queue, $payload) {
+        $this->connection($connection)->pipeline(function ($pipe) use ($connection, $queue, $payload) {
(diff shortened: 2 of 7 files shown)


henzeb:
I think this one needs a swing to restart the tests, as this seems unrelated to my changes.

taylorotwell:
We didn't really design Horizon to work across multiple Redis instances, etc. so would need to re-think this quite a bit.

henzeb:
It's ok, created a package that does what I need it to do for now: https://packagist.org/packages/henzeb/event-horizon

PR #1568, opened 2025-06-20
Fix empty failed jobs overview that stays loading

markieo1:
### Problem

When there are no failed jobs in Horizon the overview page keeps showing the Loading indicator.

1. Ensure the are no failed jobs by running `php artisan horizon:forget --all`
2. Open Horizon and enable auto refresh.
3. See that the loading doesn't stop until you disable auto refresh.

### Solution

Adjust the code such that when there are no jobs the `ready` variable is set to true. This way the Loading indicator goes away.

This is the same logic as for the Recent Job overview.

### DIFF
--- a/resources/js/screens/failedJobs/index.vue
+++ b/resources/js/screens/failedJobs/index.vue
@@ -62,10 +62,6 @@
 
                 this.$http.get(Horizon.basePath + '/api/jobs/failed?' + tagQuery + 'starting_at=' + starting)
                     .then(response => {
-                        if (!this.$root.autoLoadsNewEntries && refreshing && !response.data.jobs.length) {
-                            return;
-                        }
-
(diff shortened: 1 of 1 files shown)


taylorotwell:
I don't want to just delete code honestly without more information. Why was this code put here? Surely it must have been put here for some reason?

markieo1:
This code was introduced in https://github.com/laravel/horizon/commit/7879209b9eb9728318ec32a5183171584a2b1e37#diff-18b7c0925eb9cff29aacbe5d533f10957dbe0415c43169b12785ac5094a103be six years ago. 

I believe the author made a mistake by not setting the variable `ready` to `true`.

This does not break any functionality for the Failed Jobs overview. The next if checks the same conditions and even if `response.data.jobs.length === 0`. The code: `response.data.jobs[0]?.id !== this.jobs[0]?.id` doesn't break.

markieo1:
@taylorotwell See my response above ;-)

NadsonKT:
I'm facing the same issue. Do you have any idea when a fix will be released? @taylorotwell

PR #1575, opened 2025-08-04
Add notification for excessive pending jobs in Horizon

SanderMuller:
This PR introduces a new `ManyPendingJobsDetected` event and corresponding notification when the total number of pending jobs exceeds a configurable threshold across all queues and connections. This helps users proactively detect queue congestion before jobs begin to time out or fail.

The feature is disabled by default via `pending_jobs_monitor_threshold => null`. It is designed to be lightweight, using Redis locking and per-minute throttling to prevent unnecessary noise or overhead.

**Benefits to end users:**

Early warning when queues are backing up

Works seamlessly alongside existing LongWaitDetected events

Enables alerting based on queue volume, not just wait time

**Non-breaking change:**
This feature is opt-in and does not modify or interfere with existing Horizon functionality.

This PR already includes the fixed `dueToMonitor` logic as fixed in #1574

### DIFF
--- a/config/horizon.php
+++ b/config/horizon.php
@@ -89,6 +89,19 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Queue Pending Jobs Thresholds
+    |--------------------------------------------------------------------------
+    |
+    | This option determines how many pending jobs will trigger an event
+    | across all queues and connections in your application.
+    | Set to null to disable this monitoring entirely.
+    |
+    */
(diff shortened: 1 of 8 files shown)


taylorotwell:
I personally think it's better to just use the `LongWaitDetected` event? That tells you if a configurable queue latency threshold has been exceeded, which feels like a better metric to me than the raw number of jobs on the queue.

PR #1599, opened 2025-09-01
[5.x] Add `$tags` property support

amirhshokri:
## Description

This PR adds support for a public `$tags` property in Laravel Horizon.

Currently, Horizon determines tags primarily via the `tags()` method or by inspecting models referenced in the job. With this change, developers can now declare tags using a simple `$tags` property.

## Benefits

- Similar to how `$queue`, `$connection`, `$uniqueFor`, etc., can be defined either as a property or via a method in Laravel.
- Easier declaration of tags.
- Fully backward compatible with `tags()` method.
- Works for both jobs and queued event listeners.

## Usage

```php
class Job implements ShouldQueue
{
    public $tags = ['Critical', 'HighPriority'];

    public function handle()
    {
        //...
    }
}
```

### DIFF
--- a/src/Tags.php
+++ b/src/Tags.php
@@ -82,7 +82,17 @@ class Tags
     protected static function explicitTags(array $jobs)
     {
         return collect($jobs)
-            ->map(fn ($job) => method_exists($job, 'tags') ? $job->tags(static::$event) : [])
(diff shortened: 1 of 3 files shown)


taylorotwell:
A method is more flexible.

PR #1606, opened 2025-09-18
[5.x] Horizon proxy_path configuration description.

Daniel-Monroy:
Improves the description of the `proxy_path` option in the Horizon configuration file, adding clarity about its use behind reverse proxies or when served from a subdirectory.


taylorotwell:
Doesn't match rest of docs style.

PR #1608, opened 2025-09-27
[5.x] Add `--json` option to `horizon:list` command

amirhshokri:
## Description

This PR adds a `--json` option to the `horizon:list` command.

Currently, `horizon:list` only outputs a table in the console, which is great for human readability but not ideal for automated tooling or integrations. With this new flag, the command can now output the migration status in JSON format, making it easier to consume programmatically.

Example output:

```json
[
  {
    "name": "user-abcd",
    "pid": "123",
    "supervisors": "supervisor-1",
    "status": "running"
  }
]
```

## Tests

The `horizon:list` command does not currently have any tests in the Laravel Horizon codebase. If preferred, I can add test coverage for the new `--json` option as part of this PR.

### DIFF
--- a/src/Console/ListCommand.php
+++ b/src/Console/ListCommand.php
@@ -14,7 +14,8 @@ class ListCommand extends Command
      *
      * @var string
      */
-    protected $signature = 'horizon:list';
+    protected $signature = 'horizon:list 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I don't want to piece meal add --json to every command we have.

PR #1615, opened 2025-10-03
[5.x] Prevent using the reserved `horizon` Redis connection name

amirhshokri:
## Summary
This PR adds the `checkIsReservedConnectionName` method to prevent accidental use of the reserved `horizon` Redis connection. If `horizon` is passed as a connection name, or if a conflicting configuration exists, the method throws an `InvalidArgumentException` during the Laravel app boot process.

## Why?
This change ensures safer configuration and prevents accidental misconfiguration.

image

### DIFF
--- a/src/Horizon.php
+++ b/src/Horizon.php
@@ -6,6 +6,7 @@ use Closure;
 use Exception;
 use Illuminate\Support\HtmlString;
 use Illuminate\Support\Js;
+use InvalidArgumentException;
 use RuntimeException;
 
 class Horizon
@@ -88,6 +89,23 @@ class Horizon
         static::$authUsing = $callback;
 
         return new static;
+    }
+
+    /**
+     * Check if a given Redis connection name is reserved by Horizon.
+     *
+     * @param  string  $connection
+     * @return void
+     *
+     * @throws \Exception
+     */
+    public static function checkReservedConnectionName(string $connection): void
+    {
+        if ($connection === 'horizon' || config()->has('database.redis.horizon')) {
+            throw new InvalidArgumentException(
+                'The Redis connection name [horizon] is reserved for internal use.'
+            );
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


alexmontoanelli:
Currently we've a db redis connection named horizon, and horizon.use pointing to 'horizon'.

As the image bellow, the method is just overriding the current config.

But, if your are using a redis db named horizon for other things, then the prefix key
is just begin override.

I' think this could introduce a infrastructure break for someone if you throw a exception.

Screenshot 2025-10-03 at 13 08 13

taylorotwell:
Can you just check in the service provider only?

amirhshokri:
Done @taylorotwell. 
Please let me know if any other changes are needed.

amirhshokri:
Yes, that’s correct. This configuration is currently overridden by the existing logic. However, it’s better if the validation isn’t dependent on the current logic, so that if it changes in the future for any reason, it won’t cause issues in the application.

@alexmontoanelli

patrickbrouwers:
Hello @amirhshokri this seems to be giving issues when using `config:cache` Any idea?

donnyxray:
This PR should be reversed. There is no real way to fix this without busting or bypassing the cache. 

Perhaps the config:cache code can allow for such checks to be registered somewhere and then executed upon caching.

thetheago:
Hi @amirhshokri. Thanks for this last commit, it was breaking my route:clear command.

agustin-villalba:
@donnyxray is totally right, this should be reverted. Even version 5.35.1 is not fixing the problem when you have the config cached in production. Still the error `The Redis connection name [horizon] is reserved for internal use. `

amirhshokri:
@agustin-villalba 

Do you use `'use' => 'horizon'` in horizon.php?

agustin-villalba:
It was using a hidden env variable, it's fine now. Thank you for the fix!

PR #1621, opened 2025-10-09
Allow naming of horizon instances

timacdonald:
We have several horizon instances running in different regions all within a single application.

When we get notifications from Horzion, we are unable to determine which instance of horizon the issue is occurring in.

> **Long Wait Detected**
> [Nightwatch] The "{queue}" queue on the "{connection}" connection has a wait time of {seconds} seconds.

The above notification could be any region and we have no way of knowing.

With this PR, we are able to name our Horizon instances while maintaining a single shared app name.

> **Long Wait Detected**
> [eu-central-1] The "{queue}" queue on the "{connection}" connection has a wait time of {seconds} seconds.

> **Long Wait Detected**
> [us-east-2] The "{queue}" queue on the "{connection}" connection has a wait time of {seconds} seconds.

Screenshot 2025-10-10 at 11 02 01

The default prefix also uses the APP_NAME, e.g., but I didn't want to break anything by changing this to respect the new var. It is right there in the config file for people to modify anyway.

```
    'prefix' => env(
        'HORIZON_PREFIX',
        Str::slug(env('APP_NAME', 'laravel'), '_').'_horizon:'
    ),
```

### DIFF
--- a/src/HorizonServiceProvider.php
+++ b/src/HorizonServiceProvider.php
@@ -20,12 +20,12 @@ class HorizonServiceProvider extends ServiceProvider
      */
     public function boot()
     {
+        $this->normalizeConfig();
         $this->registerEvents();
         $this->registerRoutes();
         $this->registerResources();
         $this->offerPublishing();
         $this->registerCommands();
-        $this->normalizeConfig();
     }
 
     /**
--- a/src/HorizonServiceProvider.php
+++ b/src/HorizonServiceProvider.php
@@ -129,7 +129,12 @@ class HorizonServiceProvider extends ServiceProvider
         ]);
     }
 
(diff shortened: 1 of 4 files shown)


src/HorizonServiceProvider.php
```diff
@@ -128,6 +129,13 @@ protected function registerCommands()
         ]);
     }
 
+    protected function normalizeConfig(): void
+    {
+        if (! $this->app['config']->get('horizon.name')) {
+            $this->app['config']->set('horizon.name', $this->app['cofig']->get('app.name'));
+        }
+    }
+
```
timacdonald: Backfills the value on existing installs

PR #1624, opened 2025-10-15
[5.x] Add Enum Support to Horizon Job Tags

amirhshokri:
This PR adds enum support to Horizon job tags, allowing to use enums in `tags()` arrays.

### Example:

```php
class ProcessOrder implements ShouldQueue
{
    // ...
    
    public function tags(): array
    {
        return [
            Priority::HIGH,    
            Status::SHIPPED,
            'order:' . $this->order->id
        ];
    }
}
```

PR #1625, opened 2025-10-16
[5.x] Add Enum Support to Horizon Job Tags

amirhshokri:
Resubmittion of #1624

In the previous PR, the tests were failing because the `Illuminate\Support\enum_value()` function was not defined for them. To prevent this, I think the best approach is to include the same function in Horizon’s source code, just as it’s defined in Laravel.

Please let me know if you suggest any other approach so I can make the necessary adjustments.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,6 +36,9 @@
         "predis/predis": "Required when not using the Redis PHP driver (^1.1|^2.0|^3.0)."
     },
     "autoload": {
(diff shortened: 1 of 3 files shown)


taylorotwell:
Why can't we use the one in Illuminate\Support?

taylorotwell:
Just going to skip this feature for now.

PR #1628, opened 2025-10-17
Fix: Filter inactive supervisors to only include expected ones per master

ciungulete:
Inactive supervisors from the plan were being added to all masters, causing wrong data in the Horizon dashboard.

With this fix will display only inactive supervisors that belong to each specific master.


taylorotwell:

PR #1629, opened 2025-10-17
[5.x] Add Enum Support to Horizon Job Tags

amirhshokri:
Instead of relying on the `enum_value` helper (which isn’t available within Horizon), 
this PR introduces a small inline solution to support Enum-based job tags. 
This keeps the feature self-contained until `enum_value` becomes part of the Horizon’s dependencies.


taylorotwell:
I'm getting a bit weary of adding "enum support" piecemeal across the ecosystem honestly.

Issue #1539, opened 2025-03-06
Processes not auto scaled when long running jobs are being processed


markieo1:
### Horizon Version

5.30.3

### Laravel Version

11.42.1

### PHP Version

8.4.3

### Redis Driver

PhpRedis

### Redis Version

7.2.4

### Database Driver & Version

MySQL 8.0.30 

### Description

Hi,

We've recently switched from simple balancing to auto.
We noticed that some processes weren't moving between queues while the other queue was filling up.

This seemed to be happening when there were some long running jobs in the queue. In this same queue some short jobs were being processed that caused the AutoScaler to give more processes to the queue. After processing the short jobs, the long jobs were still running. 

In the meantime another queue would get filled up. We'd expect the AutoScaler to move processes but this isn't happening. Only after the long-running jobs were finished we saw the processes being moved to another queue.

### Steps To Reproduce

In these steps we have set `maxProcesses` to 5. This can also be happening with different amounts.
Create a new queue configuration in `horizon.php`:
```php
'tests' => [
    'connection' => 'horizon',
    'queue' => ['test_long', 'test_short'],
    'autoScalingStrategy' => 'size',
    'balance' => 'auto',
    // The number of seconds to wait in between auto-scaling attempts.
    'balanceCooldown' => 3,
    // The maximum number of processes that can be scaled up or down in a single auto-scaling attempt.
    'balanceMaxShift' => 1,
    'memory' => 256,
    'tries' => 1,
    'timeout' => 3600,
    'minProcesses' => 1,
    'maxProcesses' => 5,
    'maxJobs' => 1000
]
```
Create a new `TestJob` that does nothing more than sleep:
```php
<?php

namespace App\Jobs;

use Carbon\Carbon;
use Illuminate\Bus\Queueable;
use Illuminate\Contracts\Queue\ShouldQueue;
use Illuminate\Support\Facades\Log;

class TestJob implements ShouldQueue
{
    use Queueable;

    public function __construct(private int $sleepingTime)
    {
    }

    public function handle()
    {
        $currentTime = Carbon::now();

        // Check if the seconds have passed
        while ($currentTime->diffInSeconds(Carbon::now()) < $this->sleepingTime) {
            // Do nothing
        }

        Log::info('Job has been processed');
    }
}
```
Call the following code:
```php
public function reproduce(){
    // Create 2 jobs that are very long. This causes the auto scaling to create 2 processes
    for ($long = 0; $long < 2; $long++) {
[...]

PR #1636, opened 2025-10-20
Fix autoscaling deadlock caused by terminating processes

nckrtl:
This PR addresses #1539: Processes not auto scaled when long running jobs are being processed.

I created a reproducible scenario in https://github.com/nckrtl/horizon-i1539, confirming the described issue. By following the README instructions the following scenario is executed:

1. Two queues, 1 for long running jobs, 1 for short running jobs
2. On the first queue 2 long running jobs of 5 minutes are dispatched, shortly after 100 additional shorter running jobs are dispatched (purely for demonstration purposes).
3. The long running queue scales up nicely to 4 workers, after several scaling cycles.
4. Then the short running queue get filled with jobs.
5. The long running queue scales down by 1 worker but the short running queue workers aren't increased.
6. In the following scaling cycles the long running queue stays "stuck" at 3 workers and the short running queue stays at 1 worker.
7. Only after the long running tasks in the long running queue finishes, the scaling is being corrected.

## Root cause
When a pool scales down, processes that should be terminated are marked and moved to `$terminatingProcesses`, see [scaleDown()](https://github.com/laravel/horizon/blob/5.x/src/ProcessPool.php#L105). However, totalProcessCount() in `ProcessPool.php` includes these terminating processes:

```php
public function totalProcessCount()
{
    return count($this->processes()) + count($this->terminatingProcesses);
}
```
This is used in [scalePool()](https://github.com/laravel/horizon/blob/5.x/src/AutoScaler.php#L146) and is used to determine the worker count to scale to. This is passed to [scale()](https://github.com/laravel/horizon/blob/5.x/src/ProcessPool.php#L69) in the ProcessPool, and this is where the mismatch happens

```php
{
    $processes = max(0, (int) $processes);

    if ($processes === count($this->processes)) {
        return;
    }

    if ($processes > count($this->processes)) {
        $this->scaleUp($processes);
    } else {
        $this->scaleDown($processes);
    }
}
```
A count is performed on the processes, but not including the terminating processes. Leading to this flow happening after the first scale down:

1. AutoScaler sees: totalProcessCount() = 4 (3 active + 1 terminating)
2. Calculates the workers it should scale to, in case of the `long_running` queue it should scale down to 3:
```php
$pool->scale(
    max(
        $totalProcessCount - $maxDownShift, // 4 - 1 = 3
        $supervisor->options->minProcesses, // 1
        $desiredProcessCount // 2
    ) // 3 is being passed to scale()
);

```
3. The  [scale()](https://github.com/laravel/horizon/blob/5.x/src/ProcessPool.php#L69) method receives 3, but that method counts 3 as well. Resulting in no down scale happening, and staying stuck at 3 workers. 
4. As there are no additional workers made available, `test_short` cant scale up.
5. After the first process marked for termination finishes its job and terminates, then the scale down can continue for the `long_running` queue making processes available for the `short_queue` to scale up.

## The solution
Exclude terminating processes from totalProcessCount() so AutoScaler and ProcessPool use the same count.

## Test
Added a test to proof the described issue, could be extended with additional tests to cover specific scaling scenarios.

## Cleanup
- Removed count() from ProcessPool, as totalProcessCount() does the same. As far as i could find, it was not being used. Good to double check.
- Make `scale()` in ProcessPool also use `totalProcessCount` for consistency

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -55,7 +55,7 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           persist-credentials: false
 
@@ -88,7 +88,7 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           persist-credentials: false
 
--- a/src/AutoScaler.php
+++ b/src/AutoScaler.php
@@ -151,7 +151,7 @@ class AutoScaler
 
         $desiredProcessCount = ceil($workers);
 
-        if($desiredProcessCount > $totalProcessCount) {
+        if ($desiredProcessCount > $totalProcessCount) {
             $totalProcessCount += $pool->terminatingProcesses()->count();
         }
 
--- a/src/AutoScaler.php
+++ b/src/AutoScaler.php
@@ -147,9 +147,9 @@ class AutoScaler
     {
         $supervisor->pruneTerminatingProcesses();
 
-        $desiredProcessCount = ceil($workers);
-
         $totalProcessCount = $pool->processes()->count();
+
+        $desiredProcessCount = ceil($workers);
 
         if($desiredProcessCount > $totalProcessCount) {
             $totalProcessCount += $pool->terminatingProcesses()->count();
--- a/src/AutoScaler.php
+++ b/src/AutoScaler.php
@@ -149,7 +149,11 @@ class AutoScaler
 
         $desiredProcessCount = ceil($workers);
 
-        $totalProcessCount = $pool->totalProcessCount();
+        $totalProcessCount = $pool->processes()->count();
+
+        if($desiredProcessCount > $totalProcessCount) {
+            $totalProcessCount += $pool->terminatingProcesses()->count();
+        }
 
         if ($desiredProcessCount > $totalProcessCount) {
             $maxUpShift = min(
--- a/src/AutoScaler.php
+++ b/src/AutoScaler.php
@@ -147,9 +147,9 @@ class AutoScaler
     {
         $supervisor->pruneTerminatingProcesses();
 
-        $totalProcessCount = $pool->totalProcessCount();
-
         $desiredProcessCount = ceil($workers);
+
+        $totalProcessCount = $pool->totalProcessCount();
 
         if ($desiredProcessCount > $totalProcessCount) {
             $maxUpShift = min(
--- a/src/AutoScaler.php
+++ b/src/AutoScaler.php
@@ -152,7 +152,7 @@ class AutoScaler
         $desiredProcessCount = ceil($workers);
 
         if ($desiredProcessCount > $totalProcessCount) {
-            $totalProcessCount += $pool->terminatingProcesses()->count();
+            $totalProcessCount = $pool->totalProcessCount();
         }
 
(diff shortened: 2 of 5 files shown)


taylorotwell:
Hey there - why the removal of the Countable interface? 🤔

nckrtl:
As the content of `count()` ended up being the same as the content of `totalProcessCount()` it felt it would be confusing which one should be used or is being used. I also couldn't find where `count()` was being used, so I thought it made sense to remove it. When count is removed, there is no need for the Countable interface anymore.

However, I might have overlooked something making `count()` still necessary.

taylorotwell:
Hey @nckrtl - do you know if there is a way to fix this without changing the behavior of `totalProcessCount`? Just to minimize the possibility of breaking changes elsewhere that actually do depend on it returning the count + terminating process count.

taylorotwell:
I'm also a bit torn on if this is actually a bug, primarily because of this comment:

> After the first process marked for termination finishes its job and terminates, then the scale down can continue for the long_running queue making processes available for the short_queue to scale up.

Your configuration specifies you don't want more than 5 processes (your max process count). It sounds like Horizon is indeed respecting that and not initiating another process until the terminating one finishes its job and is terminated. If we change this I worry we're going to have situations where people are going to have more processes running than are specified in their max process count, possibility causing out of memory exceptions, etc.

nckrtl:
Hey @taylorotwell I did some additional testing and you're right. My changes do indeed result in exceeding the maximum configured processes. Yet the scaling down issue was fixed, with no processes sitting idle. I think we can have best of both worlds.

Assume all my changes are undone and we have a pool that needs to scale down for the second time: 

- Desired worker count: 1
- Processes: 3
- Terminating processes: 1

Currently, when determining the total process count we always include the terminating processes. But when scaling down only the desired worker count and the (non-terminating) processes matter. If the desired worker count is lower than the processes count, we should keep marking them for termination up until the desired count is reached. Yet when we have terminating processes in a pool they block further scaling down for as long as they are processing work (current issue) as they influence the scale down calculations.

When scaling up the terminating processes should be included in the calculations to not exceed the worker/process limit. When scaling down we don't have this issue.

I'm now proposing just this change in [scalePool()](https://github.com/laravel/horizon/blob/5.x/src/AutoScaler.php#L146) in `AutoScale.php`
```php
$totalProcessCount = $pool->processes()->count(); // <-- ignore terminatingProcesses initially

if($desiredProcessCount > $totalProcessCount) {
    $totalProcessCount = $pool->totalProcessCount(); // <-- include the terminatingProcesses to prevent exceeding the worker limit
}
```

I tested this manually and everything works as described. I then tried to create tests to verify specific scenarios but the FakePool needs quite some changes to this properly. I'm more than happy to do so but wanted to get your opinion first and see if you agree with the proposed fix.

nckrtl:
Hey @taylorotwell as some time has passed by I wanted to check in how you feel about my updated proposal in the previous comment. I still think its valid, and am still more than happy to create and extend the tests.

markieo1:
Any updates on this? We'd like this change ;)

zahouanimalek1-eng:
"The dilemma between @taylorotwell’s concern about process limits and @nckrtl’s fix for scaling deadlocks highlights a fundamental architectural gap in Horizon's current lifecycle management.

While the proposed conditional logic in scalePool() attempts to bridge this, it introduces a 'Dual-Count' inconsistency that might be hard to maintain across different Redis drivers. The real issue is that Horizon treats 'Terminating' processes as 'Active' for capacity planning, but 'Invisible' for scaling commands.

I have implemented a Weight-Based Scaling Algorithm in enterprise environments that assigns different priorities to terminating vs. active processes, allowing for 'Graceful Overshooting' only when certain safety buffers are met. This prevents the deadlock without risking a sudden OOM crash on low-memory containers.

I am available to consult on re-engineering this scaling logic for organizations requiring high-precision worker management.

Check my architectural portfolio on Upwork:
https://www.upwork.com/freelancers/~01ee371dd505084ed7"

nckrtl:
@markieo1 depends on @taylorotwell on how to proceed with this. I'm still open to create a proper PR if suggested fix would be the right direction.

PR #1651, opened 2025-11-02
[5.x] Add `Delayed` column to Pending Jobs table

amirhshokri:
This PR adds a new `Delayed` column to the Pending Jobs page in Laravel Horizon.

Previously, delayed jobs were only indicated with a label, and users had to hover over it to see how long the job had been delayed (e.g., `Delayed for 3 minutes`).

Because the Pending Jobs page updates frequently and the hover interaction was inconvenient, the delayed duration is now displayed directly in a dedicated column.

image


taylorotwell:

PR #1657, opened 2025-11-06
[5.x] Make Job IDs Clickable in the `Recent Retries` Table

amirhshokri:
## Description

This pull request updates the `Recent Retries` table in Laravel Horizon so that all job IDs—including `pending` and `completed` jobs—are rendered as clickable links to their corresponding Job Details page.

Previously, only `failed` jobs had their job IDs linked. This created an inconsistent experience and made it harder to quickly inspect the jobs that were retried.

## Before

Screenshot 2025-11-06 192447

Screenshot 2025-11-06 192604

## After

Screenshot 2025-11-06 193005

Screenshot 2025-11-06 193033

### DIFF
--- a/resources/js/screens/failedJobs/job.vue
+++ b/resources/js/screens/failedJobs/job.vue
@@ -238,7 +238,12 @@
                         <a v-if="retry.status == 'failed'" :href="Horizon.basePath + '/failed/'+retry.id">
                             {{ retry.id }}
                         </a>
-                        <span v-else>{{ retry.id }}</span>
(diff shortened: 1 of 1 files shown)


tomswinkels:
@amirhshokri When I click on the ID of a completed job, I get a blank page and the console shows the following error: TypeError: Cannot read properties of undefined (reading 'pushedAt').

Screenshot 2025-11-15 at 00 26 21

amirhshokri:
Right @tomswinkels 
Since completed jobs in Horizon are only available for a limited time before they’re deleted (based on the configuration in `horizon.php`), I think in this case the job has already been removed and is no longer accessible.

PR #1666, opened 2025-11-23
[5.x] Add `Silenced` Badge to Jobs Lists

amirhshokri:
## Summary

This PR adds a `Silenced` badge to job entries displayed in the following Horizon sections:

- Monitoring
- Pending Jobs
- Failed Jobs

The badge is shown in the same style and position as existing badges (e.g., `Delayed`), making silenced jobs easily recognizable at a glance.

## Why?

Horizon already supports silenced jobs via the `silenced` configuration option.  
Silenced jobs do not appear in the *Completed Jobs* list and instead appear in a dedicated *Silenced* page.

However, until such a job actually finishes running (and moves to the Silenced page), there is **no visual indicator** that the job is silenced while it is:

- Pending  
- Being monitored  
- Failed  

This creates a UX gap: developers cannot know that a job is silenced **until after it has executed.**

This PR closes that gap by making the silenced status visible in all job lists where silenced jobs may appear.

## Screenshots

Monitoring:

image

&nbsp;

Pending Jobs:

image

&nbsp;

Failed Jobs:

image

### DIFF
--- a/resources/js/screens/failedJobs/index.vue
+++ b/resources/js/screens/failedJobs/index.vue
@@ -257,6 +257,11 @@
                             Retried
                         </small>
 
+                        <small class="ms-1 badge bg-secondary badge-sm"
+                               v-if="job.payload.silenced">
+                            Silenced
+                        </small>
+
                         <br>
 
                         <small class="text-muted">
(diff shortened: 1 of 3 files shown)


taylorotwell:
I don't know if anyone actually has a problem with this though.

PR #1671, opened 2025-11-28
Add reload command

barryvdh:
Implements `reloads()` from https://github.com/laravel/framework/pull/57923

Added the check in console only, similar to where the commands are located, with a check to avoid bumping the version.


barryvdh:
Note: This replaces the existing 'queue' command. Alternatively we could add it as horizon, but that would maybe restart them in the wrong order?

PR #1677, opened 2025-12-10
Show attempts on completed jobs detail

Button99:
Hello, this PR  shows the attempts on the completed job detail.

image

image

This would be helpful to spot jobs that eventually succeed but required multiple retries, which is useful when diagnosing transient failures. For example, a job showing 3 attempts could indicate API timeouts or race conditions that deserve further investigation.


taylorotwell:

PR #1683, opened 2025-12-20
[5.x] Display queue paused status in Horizon dashboard

amirhshokri:
## Summary

In commit https://github.com/laravel/framework/pull/57800, Laravel introduced the queue paused feature, allowing queues to be temporarily paused. However, Horizon did not provide a way to visualize this state in its dashboard.

This PR adds the paused status under a `Status` column in the `Current Workload` table, making it easy for users to see which queues are currently paused.

### Backward Compatibility

If Laravel does not include the paused feature, or if Horizon is updated before the Laravel upgrade, the paused status defaults to `false`, ensuring that no errors occur and the dashboard remains fully functional.

## Screenshot

image


taylorotwell:
Horizon has its own pause / continue commands.

PR #1686, opened 2025-12-26
[5.x] Add `ParsesQueue` trait

amirhshokri:
This PR introduces a small internal `ParsesQueue` trait for Horizon that mirrors Laravel’s queue parsing behavior without introducing a hard dependency on Laravel 12 (https://github.com/laravel/framework/pull/57863).

All internal `explode(':', ...)` usages related to queue parsing have been replaced with this method to ensure consistent and robust behavior, including safe handling of missing connection or queue names (falling back to `config('queue.default')` and `default`).


taylorotwell:
I think it's already working fine.

PR #1687, opened 2025-12-31
[5.x] Clarify `monitored` trim precedence over `completed`

amirhshokri:
This PR clarifies that for `monitored` jobs, the `monitored` trim value takes precedence over `completed`.

Developers might not expect `monitored` jobs to affect the trimming of `completed` jobs; however, in practice, `monitored` jobs override `completed`.


taylorotwell:
I don't find this super surprising.

PR #1691, opened 2026-01-15
[5.x] Supports restricting private tunnel requests for `local` environment

crynobone:


AbdelElrafa:
Hi @crynobone,

I was looking into why Horizon stopped working in our dev environments and ultimately traced it back to this. We figured out that we can register a custom 'horizon' driver to add our own logic, basically just `return true`, and that got it working for us.

Can you share the reasoning being the logic for this change and the checks for local tunneling which are in Sentinel? I'd like to learn more about the reasoning for that.

Thanks!

PR #1697, opened 2026-02-02
[5.x] Specify npm package name

amirhshokri:
Updates the file watcher error message to reference `paulmillr/chokidar` instead of just `chokidar` for clarity. This helps users install the correct package using npm.


taylorotwell:
It is installed like this: `npm i chokidar`

PR #1733, opened 2026-03-22
Fix 'delayed until' showing wrong time (fixes laravel/horizon#1668)

### LINKED ISSUE #1668: "Delayed until" showing wrong time (identical to Pushed, not taking delay into account)
### Horizon Version

5.40.0

### Laravel Version

12.39.0

### PHP Version

8.3

### Redis Driver

Predis

### Redis Driver Version

2.4.1

### Database Driver & Version

MySQL 8.0.27 via DBNgin 7.9 for macOS 15.7.1 (arm64)

### Description

The "Delay until" time should be different from "Pushed" time and in accordance to the actual time when the job is going to executed (or at least the minimum time after which the job is able to be picked up).

### Steps To Reproduce

I used the following code to dispatch a job and the screenshot is the result as shown in Horizon.
```php
CheckForNewTenantFiles::dispatch($tenant)
            ->delay(CarbonInterval::minutes(10));
```

Image

JoshSalway:
## Summary

Fixes laravel/horizon#1668 -- The "delayed until" timestamp in Horizon's job detail view shows the wrong time because the delay value is not stored in the job payload when dispatching via `later()`. The frontend tries to compute the delayed-until time from `pushedAt + delay`, but `delay` is missing from the payload.

## Root Cause

**The bug exists because** `RedisQueue::later()` calculates the delay for Redis sorted set scheduling but never persists the delay value into the job payload itself. When Horizon's frontend displays a delayed job, it attempts to deserialize the job class to read its `delay` property. If the job was dispatched with `->delay(60)` via the queue method (not as a class property), the deserialized object has no delay information. The frontend falls back to showing nothing or an incorrect timestamp.

## Why This Fix Works

**This fix works because** it stores the delay (in seconds) directly in the job payload via `$preparedPayload->set(['delay' => $delayInSeconds])` inside `RedisQueue::later()`. The `RedisJobRepository::pushed()` method then persists this value alongside the job record. The Vue frontend already had partial support for reading delay from the payload -- this PR adds a fallback that reads `this.job.delay` when the deserialized class property is unavailable, computing the correct "delayed until" timestamp as `pushedAt + delay`.

## Alternatives Considered

**We chose this approach over** always deserializing the job class to read its delay because: (1) deserialization is expensive and fragile (class may not exist on the Horizon dashboard server); (2) storing the delay in the payload is the source of truth -- it captures the actual delay used at dispatch time, not a property that may differ from the runtime value; (3) the payload-based approach works for all job types, including closures and jobs dispatched with dynamic delays via `->delay()`.

## Files Changed

- `src/RedisQueue.php` -- Stores delay in job payload during `later()` dispatch
- `src/Repositories/RedisJobRepository.php` -- Persists delay from payload into job record
- `resources/js/screens/recentJobs/job.vue` -- Frontend fallback to read delay from job record when unserialized property unavailable

### DIFF
--- a/src/RedisQueue.php
+++ b/src/RedisQueue.php
@@ -108,7 +108,13 @@ class RedisQueue extends BaseQueue
     #[\Override]
     public function later($delay, $job, $data = '', $queue = null)
     {
-        $payload = (new JobPayload($this->createPayload($job, $queue, $data)))->prepare($job)->value;
+        $preparedPayload = (new JobPayload($this->createPayload($job, $queue, $data)))->prepare($job);
+
+        $delayInSeconds = $this->secondsUntil($delay);
+
+        $preparedPayload->set(['delay' => $delayInSeconds]);
+
+        $payload = $preparedPayload->value;
 
         if (method_exists($this, 'enqueueUsing')) {
             return $this->enqueueUsing(
--- a/resources/js/screens/recentJobs/job.vue
+++ b/resources/js/screens/recentJobs/job.vue
@@ -130,6 +130,10 @@
                     return this.formatDate(this.job.payload.pushedAt).add(unserialized.delay, 'seconds')
                         .local()
                         .format('YYYY-MM-DD HH:mm:ss');
+                } else if (this.job.delay > 0) {
+                    return this.formatDate(this.job.payload.pushedAt).add(this.job.delay, 'seconds')
+                        .local()
+                        .format('YYYY-MM-DD HH:mm:ss');
                 }
 
                 return null;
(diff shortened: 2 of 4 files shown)


cosmastech:
@JoshSalway is there anything I can help with for us to get this out of draft and into a reviewable state?

I'm building some logging middleware for job latency. I want to include how long the job was delayed in the metrics.  Claude reported that `RedisQueue@later()` is the root cause of the problem, because it was never updated to accommodate for the changes in https://github.com/laravel/framework/pull/55529.

I have no opinion on the UI change. 😅

src/RedisQueue.php
```diff
@@ -108,7 +108,13 @@ protected function createPayloadArray($job, $queue, $data = '')
     #[\Override]
     public function later($delay, $job, $data = '', $queue = null)
     {
-        $payload = (new JobPayload($this->createPayload($job, $queue, $data)))->prepare($job)->value;
+        $preparedPayload = (new JobPayload($this->createPayload($job, $queue, $data)))->prepare($job);
```
cosmastech, line 111: **question:** Couldn't we just pass `$delay` as the 4th parameter to `createPayload()`?

PR #1741, opened 2026-03-24
Fix race condition in RedisHorizonCommandQueue::pending()

### LINKED ISSUE #1739: Race condition in RedisHorizonCommandQueue::pending() can silently drop supervisor commands
## Description

`RedisHorizonCommandQueue::pending()` has a race condition that can silently drop supervisor commands (`Terminate`, `Scale`, `Balance`, `Pause`, `Continue`).

### The code (`src/RedisHorizonCommandQueue.php`)

```php
public function pending($name)
{
    $length = $this->connection()->llen('commands:'.$name);

    if ($length < 1) {
        return [];
    }

    $results = $this->connection()->pipeline(function ($pipe) use ($name, $length) {
        $pipe->lrange('commands:'.$name, 0, $length - 1);
        $pipe->ltrim('commands:'.$name, $length, -1);
    });
    // ...
}
```

### The problem

1. `LLEN` is a separate Redis round-trip that returns a count (e.g. `2`)
2. Between `LLEN` and the `pipeline()`, another process can `RPUSH` a new command
3. The pipeline uses the **stale** `$length` to `LTRIM`, which can discard the newly pushed command
4. `pipeline()` itself provides no atomicity — it batches commands for performance but does not use `MULTI/EXEC`, so commands from other clients can interleave between the `LRANGE` and `LTRIM`

### How this was found

During a codebase review looking for non-atomic Redis patterns similar to the `SETNX`/`EXPIRE` race in `Lock::get()` (#1738). The `pending()` method uses the same class of bug: a multi-step Redis operation without atomicity guarantees.

### Existing pattern in the codebase

`RedisMetricsRepository::baseSnapshotData()` already solves this exact pattern correctly using `->transaction()` (Redis `MULTI/EXEC`):

```php
$responses = $this->connection()->transaction(function ($trans) use ($key) {
    $trans->hmget($key, ['throughput', 'runtime']);
    $trans->del($key);
});
```

### Suggested fix

Replace the `LLEN` + `pipeline()` with a `transaction()` that reads all items and deletes atomically — matching the existing pattern:

```php
public function pending($name)
{
    $results = $this->connection()->transaction(function ($pipe) use ($name) {
        $pipe->lrange('commands:'.$name, 0, -1);
        $pipe->del('commands:'.$name);
    });

    return collect($results[0] ?? [])
        ->map(fn ($result) => (object) json_decode($result, true))
        ->all();
}
```

### Impact

Supervisor commands could be silently dropped under concurrent load. No errors, no logs — the command vanishes. This is particularly dangerous for `Terminate` commands during deployments where timing is tight.

Related symptoms that could be partially explained by this:
- #1535 (silent 60-second delays)
- #242 (Horizon stops consuming jobs)
- #685 (Horizon suddenly stopped)

JoshSalway:
## Summary

Simplify `RedisHorizonCommandQueue::pending()` by replacing `LLEN` + `pipeline(LRANGE/LTRIM)` with `transaction(LRANGE/DEL)`, matching the existing pattern in `RedisMetricsRepository::baseSnapshotData()`.

Fixes #1739

## The change

The current code makes a separate `LLEN` call, then uses that count as the offset for `LRANGE` and `LTRIM` inside a `pipeline()`. This works in practice because each supervisor has its own command queue and reads it synchronously, but the offset from `LLEN` is inherently stale — commands pushed between `LLEN` and the pipeline are picked up on the next loop iteration rather than the current one.

Using `transaction()` with `LRANGE 0 -1` + `DEL` simplifies this to a single operation. No offset calculation needed, no stale count.

### Before

```php
$length = $this->connection()->llen('commands:'.$name);

if ($length < 1) {
    return [];
}

$results = $this->connection()->pipeline(function ($pipe) use ($name, $length) {
    $pipe->lrange('commands:'.$name, 0, $length - 1);
    $pipe->ltrim('commands:'.$name, $length, -1);
});
```

### After

```php
$results = $this->connection()->transaction(function ($pipe) use ($name) {
    $pipe->lrange('commands:'.$name, 0, -1);
    $pipe->del('commands:'.$name);
});
```

## Why transaction() over pipeline()

`pipeline()` batches commands to reduce network round-trips. `transaction()` wraps commands in Redis `MULTI/EXEC`, which the [Redis docs](https://redis.io/docs/latest/develop/using-commands/transactions/) describe as:

> "All the commands in a transaction are serialized and executed sequentially. A request sent by another client will never be served in the middle of the execution of a Redis Transaction."

For a read-then-delete pattern, `transaction()` ensures the `LRANGE` and `DEL` execute as a single isolated unit. This is the same approach already used in `RedisMetricsRepository::baseSnapshotData()` in the same codebase for the same reason.

## Sentinel compatibility note

Issue #193 previously raised a concern about `transaction()` (MULTI/EXEC) not working with Redis Sentinel via Predis. PRs #198 and #199 attempted to remove the existing `transaction()` from `baseSnapshotData()` but were both closed — the team chose to keep `transaction()`. This PR uses the same approach and does not change the Sentinel compatibility status of Horizon.

## Tests

6 new integration tests against real Redis: command retrieval, atomic consumption (second `pending()` returns empty), empty queue, option preservation, flush, and supervisor isolation.

## Impact

The source change is 4 lines added and 11 removed (+86 lines for the new test file). Eliminates the stale `LLEN` offset and uses `MULTI/EXEC` consistent with the existing codebase pattern.

### DIFF
--- a/src/RedisHorizonCommandQueue.php
+++ b/src/RedisHorizonCommandQueue.php
@@ -49,19 +49,12 @@ class RedisHorizonCommandQueue implements HorizonCommandQueue
      */
     public function pending($name)
     {
-        $length = $this->connection()->llen('commands:'.$name);
-
-        if ($length < 1) {
-            return [];
-        }
-
-        $results = $this->connection()->pipeline(function ($pipe) use ($name, $length) {
-            $pipe->lrange('commands:'.$name, 0, $length - 1);
-
-            $pipe->ltrim('commands:'.$name, $length, -1);
+        $results = $this->connection()->transaction(function ($pipe) use ($name) {
+            $pipe->lrange('commands:'.$name, 0, -1);
+            $pipe->del('commands:'.$name);
         });
 
-        return collect($results[0])
+        return collect($results[0] ?? [])
             ->map(fn ($result) => (object) json_decode($result, true))
             ->all();
     }
--- a/tests/Feature/RedisHorizonCommandQueueTest.php
+++ b/tests/Feature/RedisHorizonCommandQueueTest.php
@@ -1,0 +1,86 @@
+<?php
+
+namespace Laravel\Horizon\Tests\Feature;
+
+use Laravel\Horizon\RedisHorizonCommandQueue;
+use Laravel\Horizon\Tests\IntegrationTest;
+
+class RedisHorizonCommandQueueTest extends IntegrationTest
+{
+    public function test_pending_returns_all_pushed_commands()
+    {
+        $queue = $this->app->make(RedisHorizonCommandQueue::class);
+
+        $queue->push('test-supervisor', 'Scale', ['processes' => 5]);
+        $queue->push('test-supervisor', 'Balance', []);
+        $queue->push('test-supervisor', 'Pause', []);
+
+        $commands = $queue->pending('test-supervisor');
+
+        $this->assertCount(3, $commands);
+        $this->assertEquals('Scale', $commands[0]->command);
+        $this->assertEquals('Balance', $commands[1]->command);
+        $this->assertEquals('Pause', $commands[2]->command);
+    }
+
+    public function test_pending_consumes_commands_atomically()
+    {
+        $queue = $this->app->make(RedisHorizonCommandQueue::class);
+
+        $queue->push('test-supervisor', 'Scale', ['processes' => 5]);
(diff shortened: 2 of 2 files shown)


taylorotwell:
These aren't fixing user reported issues.

PR #1754, opened 2026-04-02
[5.x] Replace sleep() with Carbon time travel in ProcessRepositoryTest

JoshSalway:
## Summary

- Replaces `sleep(2)` with `CarbonImmutable::setTestNow()` in `ProcessRepositoryTest::test_expired_orphans_can_be_found`
- The underlying `RedisProcessRepository::orphaned()` and `orphanedFor()` methods use `CarbonImmutable::now()->getTimestamp()`, so Carbon time travel works correctly here

## Problem

The test uses a real `sleep(2)` to wait for orphan expiry time to pass. On slow CI runners, this is both:
1. **Flaky** — if the CI machine is under load, the 2-second margin may not be sufficient for the 1-second expiry window check
2. **Slow** — adds 2 seconds of wall-clock time to every test run unnecessarily

## Solution

Replace `sleep(2)` with `CarbonImmutable::setTestNow(CarbonImmutable::now()->addSeconds(2))` to simulate time passing without actually waiting. Added `tearDown()` to reset Carbon state.

## Test plan

- [x] Both ProcessRepositoryTest tests pass locally
- [x] Confirmed `orphaned()` and `orphanedFor()` both use `CarbonImmutable::now()` (not Redis TTL)
- [x] Carbon state properly cleaned up in tearDown

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tests/Feature/ProcessRepositoryTest.php
+++ b/tests/Feature/ProcessRepositoryTest.php
@@ -2,17 +2,25 @@
 
 namespace Laravel\Horizon\Tests\Feature;
 
+use Carbon\CarbonImmutable;
 use Laravel\Horizon\Contracts\ProcessRepository;
 use Laravel\Horizon\Tests\IntegrationTest;
 
 class ProcessRepositoryTest extends IntegrationTest
 {
+    protected function tearDown(): void
+    {
+        CarbonImmutable::setTestNow();
+
+        parent::tearDown();
+    }
+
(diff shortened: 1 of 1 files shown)


JoshSalway:
Closing — this conflicts with #1751 which touches the same file and also introduces sleep() calls. Will fold the Carbon time travel approach into #1751 instead.

PR #1760, opened 2026-04-14
[5.x] Add first-class Redis Cluster support

timmylindh:
When using AWS ElastiCache Serverless (Valkey) or any Redis Cluster deployment, Horizon fails because `Horizon::use()` extracts a single node from the cluster config and creates a standalone connection. This doesn't work with cluster-only endpoints. Additionally, Horizon's Lua scripts and multi-key operations cause `CROSSSLOT` errors because keys hash to different slots.
                                                                                                                            
  This PR adds first-class Redis Cluster support to Horizon, following laravel/framework#59533 which added cluster support to the Queue driver.

  ### Approach: Hash-Tagged Prefix                                                                                          
   
  When a cluster connection is detected, the Horizon prefix is automatically wrapped in a Redis hash tag (`horizon:` → `{horizon:}`), ensuring all Horizon keys hash to the same slot. This eliminates all `CROSSSLOT` errors for Lua scripts, pipelines, and multi-key operations with zero changes to repository code.                                                 
                                                            
  ### Changes

  **`Horizon::use()` (`Horizon`)**
  - Cluster connections now register as `database.redis.clusters.horizon` (a proper cluster connection) instead of
  extracting a single node                                                                                                  
  - Prefix is automatically wrapped in hash tags on cluster connections so all Horizon keys land on the same slot
  - Falls back to old single-node behavior on framework versions without cluster support (`Connection::hasHashTag`)         
                                                            
  **`readyNow()` (`RedisQueue`)**
  - Calls `getQueueRedisKey()` (from laravel/framework#59533) so it reads from the correct queue key on cluster connections
  - Falls back to `getQueue()` on older framework versions                                                                  
   
  **`ClusterAwarePipeline` trait (`Repositories`)**                                                                         
  - PhpRedis's `RedisCluster` class does not support `pipeline()`. The framework handles this in `RedisQueue` by using `transaction()` for `PhpRedisClusterConnection` — this trait applies the same pattern across all Horizon repositories and the command queue.
  - Used by `RedisJobRepository`, `RedisTagRepository`, `RedisMasterSupervisorRepository`, `RedisSupervisorRepository`,  `RedisProcessRepository`, and `RedisHorizonCommandQueue`.                                                                 
   
  ### Backward Compatibility                                                                                                
                                                            
  - **Non-cluster users**: Zero change. Standalone connection path is identical to the previous behavior.                   
  - **Older framework versions**: `supportsCluster()` checks for `Connection::hasHashTag`. If the framework doesn't have it, falls back to extracting the first cluster node (old behavior).                                                          
  - **Existing `{default}` workaround**: `ensureHashTaggedPrefix()` detects existing hash tags and prevents double-wrapping.
  - **phpredis + predis**: Both drivers tested against standalone and cluster configurations.

### DIFF
--- a/tests/worker.php
+++ b/tests/worker.php
@@ -7,13 +7,15 @@ use Illuminate\Contracts\Events\Dispatcher;
 use Illuminate\Queue\QueueManager;
 use Illuminate\Queue\Worker;
 use Illuminate\Queue\WorkerOptions;
+use Laravel\Horizon\Tests\RedisClusterHelper;
 
 use function Orchestra\Testbench\container;
 
 // Configure the application...
 $app = container()->createApplication();
-$app->register(Laravel\Horizon\HorizonServiceProvider::class);
 $app->make('config')->set('queue.default', 'redis');
+
+RedisClusterHelper::configure($app);
 
 $worker = new Worker(
     $app->make(QueueManager::class),
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -69,3 +69,59 @@ jobs:
 
       - name: Execute tests
         run: vendor/bin/phpunit --fail-on-risky --fail-on-warning --stop-on-error --stop-on-failure ${{ matrix.laravel >= 10 && '--fail-on-deprecation' || '' }}
+
+  cluster-tests:
+    runs-on: ubuntu-24.04
+
+    strategy:
+      fail-fast: true
+      matrix:
+        php: [8.3, 8.4, 8.5]
+        client: ['phpredis', 'predis']
+
+    name: Redis Cluster - PHP ${{ matrix.php }} (${{ matrix.client }})
+
+    steps:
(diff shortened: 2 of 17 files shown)


tests/worker.php
```diff
@@ -7,14 +7,16 @@
 use Illuminate\Queue\QueueManager;
 use Illuminate\Queue\Worker;
 use Illuminate\Queue\WorkerOptions;
+use Laravel\Horizon\Tests\RedisClusterHelper;
 
 use function Orchestra\Testbench\container;
 
 // Configure the application...
 $app = container()->createApplication();
-$app->register(Laravel\Horizon\HorizonServiceProvider::class);
```
timmylindh, line 15: When container()->createApplication() runs, it reads testbench.yaml and registers + boots all listed providers automatically. The explicit $app->register() call was a no-op since the provider was already registered.

PR #1762, opened 2026-04-27
[5.x] Database Driver

stevebauman:
**Description**:

This PR adds a database driver for Horizon as an alternative to Redis.

The driver is selected via a new `horizon.driver` config key (default: `redis`). When set to `database`, every Horizon repository, the command queue, the lock, and the queue connector are wired against database-backed implementations. The dashboard, CLI, supervisor lifecycle, metrics, monitoring, and notification behaviour are unchanged.

**Notes**:

- While this is 100% backwards compatible (existing Redis-based installs require no changes) -- there is a new interface method `TagRepository::trimExpired()`, which I believe introduces a major breaking change.
- `DatabaseWorkloadRepository` is nearly identical to the `RedisWorkloadRepository`. I was thinking about making an abstract class to contain the duplicate logic, but unsure what to name it without conflicting with the `WorkloadRepository` interface. If you have suggestions let me know!
- All Horizon timestamps remain microtime-precise strings/floats so we maintain operational uniformity with Redis and don't have to touch the frontend to handle different formats.
- Targets `5.x`; supporting Laravel `9.x` through `13.x`.
- Support for SQLServer is not included in this PR ([removed here](https://github.com/laravel/horizon/pull/1762/changes/7ecb1aefecbad0fa39c858b48eeba30fc5be1779)) due to query compatibility issues

**Testing**:

I manually verified dashboard behaviour end-to-end against the database driver with MySQL and SQLite:

- Job + queue metrics pages
- Batches view (list + detail)
- Failed job retry flow + retry chain rendering
- Supervisor + master supervisor + workload widgets
- Monitoring (add/remove tag, monitored job listings)
- Dashboard home + stats, pending/completed/silenced/failed job views
- Manually verified `horizon:install`, `horizon:work`, supervisor scaling, pause/continue, terminate, snapshot, clear, purge, and clear-metrics commands

**Demos (MySQL 8.0)**:

<details>
<summary><code>TestJob.php</code></summary>

```php
namespace App\Jobs;

use Illuminate\Bus\Batchable;
use Illuminate\Bus\Queueable;
use Illuminate\Contracts\Queue\ShouldQueue;
use Illuminate\Foundation\Bus\Dispatchable;
use Illuminate\Queue\InteractsWithQueue;
use Illuminate\Queue\SerializesModels;

class TestJob implements ShouldQueue
{
    use Batchable, Dispatchable, InteractsWithQueue, Queueable, SerializesModels;

    public function __construct(
        public int $sleep = 2,
        public bool $fail = false,
    ) {
    }

    public function handle(): void
    {
        sleep($this->sleep);

        if ($this->fail) {
            throw new \RuntimeException('Intentional failure for testing.');
        }
    }

    public function tags(): array
    {
        return ['test', $this->fail ? 'fail' : 'ok'];
    }
}
```
</details>

<details>
<summary>Mixed Traffic Across All Queues</summary>

```php
class DashboardOverviewSeeder extends Seeder
{
    /**
     * Mixed traffic across all queues for the main dashboard view.
     *
     * Demonstrates per-queue workload, wait times, throughput, recent jobs panel, current process counts.
     */
    public function run(): void
    {
        $queues = ['default', 'emails', 'notifications'];

        foreach ($queues as $queue) {
            for ($i = 0; $i < 30; $i++) {
                TestJob::dispatch(sleep: random_int(0, 2), fail: $i % 12 === 0)
                    ->onQueue($queue);
            }
        }

        $this->command->info('Dispatched 90 jobs (30 per queue) with ~8% failure rate.');
    }
}
```

https://github.com/user-attachments/assets/0d236009-0849-40cb-bf45-f9984e7de6ea
</details>

<details>
<summary>Auto-Scaling Between Queues</summary>

```php
class AutoScalingSeeder extends Seeder
{
    /**
     * Heavily load one queue and lightly drip another so the "auto" balance strategy shifts processes toward the busy queue.
     *
     * Demonstrates per-queue process count changing in real time, supervisor scaling up to maxProcesses, balanceMaxShift cooldown.
     */
    public function run(): void
    {
        for ($i = 0; $i < 80; $i++) {
            TestJob::dispatch(sleep: 4)->onQueue('emails');
        }

        for ($i = 0; $i < 5; $i++) {
            TestJob::dispatch(sleep: 1)->onQueue('default');
        }

        $this->command->info('Dispatched 80 slow jobs to "emails" and 5 light jobs to "default".');
        $this->command->info('Watch supervisor-1 shift processes toward "emails" on the dashboard.');
    }
}
```

https://github.com/user-attachments/assets/61966980-4b4b-45a4-a62b-fadb4c89a9c4
</details>

<details>
<summary>Monitoring Tags</summary>

```php
class MonitoredTagSeeder extends Seeder
{
    /**
     * Start monitoring a tag, then dispatch matching jobs.
     *
     * Demonstrates: monitored tags view, job count per tag, tag drill-down to see individual jobs.
     */
    public function run(): void
    {
        $tags = app(TagRepository::class);
        $tags->monitor('vip');
        $tags->monitor('newsletter');

        for ($i = 0; $i < 120; $i++) {
            TaggedJob::dispatch(['vip', 'customer:'.($i + 1)])->onQueue('default');
        }

        for ($i = 0; $i < 25; $i++) {
            TaggedJob::dispatch(['newsletter', 'campaign:spring'])->onQueue('emails');
        }

        $this->command->info('Monitoring "vip" + "newsletter". Dispatched 120 vip + 25 newsletter tagged jobs.');
    }
}
```

https://github.com/user-attachments/assets/8836f1ae-8099-46f2-b5e4-be9a7fb3d401
</details>

<details>
<summary>Batched Jobs</summary>

```php
class BatchSeeder extends Seeder
{
    /**
     * Dispatch a batch of mixed-success jobs.
     *
     * Demonstrates batches view, batch progress, pending/processed/failed counts within a batch.
     */
    public function run(): void
    {
        $jobs = collect(range(1, 40))->map(function ($i) {
            return new TestJob(sleep: random_int(1, 3), fail: $i % 8 === 0);
        })->all();

        $batch = Bus::batch($jobs)
            ->name('Demo Batch')
            ->allowFailures()
            ->onQueue('default')
            ->dispatch();

        $this->command->info('Dispatched batch '.$batch->id.' with 40 jobs (~12% failure rate).');
    }
}
```

https://github.com/user-attachments/assets/dcbfbfac-7dcb-4c31-8796-9e31043c9d99
</details>

Let me know your thoughts and if you'd like anything adjusted/changed/refactored! Thanks for your time! ❤️


src/Contracts/TagRepository.php
```diff
@@ -87,4 +87,11 @@ public function paginate($tag, $startingAt = 0, $limit = 25);
      * @return void
      */
     public function forget($tag);
+
+    /**
+     * Trim expired tag entries from storage.
+     *
+     * @return void
+     */
+    public function trimExpired();
```
stevebauman, line 96: Redis benefits from expiring temporary tags natively via TTL. Other drivers must trim them explicitly -- so I've added this method to do so here.

src/Repositories/DatabaseTagRepository.php
```diff
@@ -0,0 +1,261 @@
...
+        $time = str_replace(',', '.', microtime(true));
+
+        $rows = array_map(function ($tag) use ($id, $time, $expiresAt) {
+            return [
+                'tag' => $tag,
+                'job_id' => (string) $id,
+                'created_at' => $time,
+                'expires_at' => $expiresAt,
+            ];
+        }, array_values(array_unique($tags)));
```
stevebauman, line 122: This `array_unique` is required for PostgreSQL. Its `ON CONFLICT DO UPDATE` raises `SQLSTATE[21000] Cardinality violation` if the same conflict target appears twice in one statement. MySQL/SQLite tolerate it silently.

peterjthomson:
Oh man. I've wanted this for so long. Would be amazing for local testing, quick db driver only prototypes etc. But seeing the migrations makes me realise how much stuff Horizon is doing. There's precedent from Telescope and Pulse so maybe it's time?

ipalaus:
@stevebauman we've been running this in production for PostgreSQL with heavy load. We have an internal package that is built on top of Horizon, similar approach as yours. I can potentially open it up, or compare what you built with ours and see the differences and what hiccups we run once we deployed this to production

src/HorizonServiceProvider.php
```diff
@@ -198,6 +205,24 @@ protected function registerServices()
                 ? $this->app->singleton($value)
                 : $this->app->singleton($key, $value);
         }
+
+        foreach (array_keys($this->redisServiceBindings) as $abstract) {
+            $this->app->singleton($abstract, function ($app) use ($abstract) {
+                return $app->make($this->driverServiceBindings()[$abstract]);
+            });
+        }
```
stevebauman, line 213: This approach might need to be tweaked, since we're assuming these bindings will always be the same per driver. Let me know your thoughts!

src/Lock.php
```diff
@@ -2,28 +2,8 @@
 
 namespace Laravel\Horizon;
 
-use Illuminate\Contracts\Redis\Factory as RedisFactory;
-
-class Lock
+abstract class Lock
```
stevebauman, line 5: This is likely a breaking change as well since we're no longer able to create `Lock` instances.

ipalaus:
Here's a comparison by Claude of what we have on our horizon-postgre package vs your current PR. I can also try to do a PR against yours, if you prefer.

The optimisations for the PG driver were shaped around this: https://planetscale.com/blog/keeping-a-postgres-queue-healthy

--- 

We've been running an in-tree Postgres-only Horizon driver in production for a while. The shape ended up very similar to this PR, but we collected a handful of footguns that are worth surfacing — most are driver-agnostic, a few are PG-only.

Sections 6 and 7 are PG-specific but high-impact; the rest applies to MySQL too.

---

## 1. Trim must be batched, not a single `DELETE`

`trimRecentJobs()` / `trimFailedJobs()` / `trimMonitoredJobs()` issue an unbounded `DELETE WHERE created_at < cutoff`. On a busy queue this:

- Holds long row locks and blocks concurrent inserts/updates.
- Creates a massive dead-tuple spike that overwhelms autovacuum on PG (and bloats the InnoDB undo log on MySQL).
- Risks lock-wait timeouts when run alongside `pushed()` / `reserved()` / `completed()`.

Trim in 1k batches:

```php
protected function batchDelete(callable $constraints, int $batchSize = 1000): void
{
    do {
        $query = $this->table();
        $constraints($query);

        $ids = $query->limit($batchSize)->pluck('id');

        $deleted = $ids->isNotEmpty()
            ? $this->table()->whereIn('id', $ids)->delete()
            : 0;
    } while ($deleted >= $batchSize);
}

public function trimRecentJobs(): void
{
    $now = CarbonImmutable::now();

    $this->batchDelete(fn ($q) => $q
        ->whereNotIn('status', ['failed', 'silenced', 'completed'])
        ->where('created_at', '<', $now->subMinutes($this->recentJobExpires)));

    $this->batchDelete(fn ($q) => $q
        ->where('status', 'completed')
        ->where('completed_at', '<', $now->subMinutes($this->completedJobExpires)));

    // ...
}
```

Same applies to `DatabaseTagRepository::trimExpired()`.

---

## 2. Race conditions on retry references and metrics

Two read-modify-write paths in `DatabaseJobRepository` are not transactional:

- `updateRetryInformationOnParent()` reads `retried_by`, mutates JSON, writes back.
- `storeRetryReference()` does the same.

Concurrent retries on the same parent will lose updates. Wrap in a transaction with `lockForUpdate()`:

```php
protected function updateRetryStatusOnParent(JobPayload $payload, bool $failed): void
{
    $parentId = $payload->retryOf();

    $this->db()->transaction(function () use ($parentId, $payload, $failed): void {
        $existing = $this->db()->table('horizon_jobs')
            ->where('id', $parentId)
            ->lockForUpdate()
            ->value('retried_by');

        $retries = collect(json_decode($existing ?? '[]', true))
            ->map(fn ($retry) => $retry['id'] === $payload->id()
                ? array_merge($retry, ['status' => $failed ? 'failed' : 'completed'])
                : $retry)
            ->all();

        $this->db()->table('horizon_jobs')
            ->where('id', $parentId)
            ->update(['retried_by' => json_encode($retries)]);
    });
}
```

`DatabaseMetricsRepository::incrementMetric()` does have a `lockForUpdate`, but it serializes **every** write for `queue:default` through one row lock. Under high throughput this becomes the dominant contention point. Replace with a single atomic upsert that recomputes the running average inside SQL — no PHP-side transaction, no row lock held across round trips:

```php
protected function incrementMetric(string $type, string $slug, float $runtime): void
{
    // Defeat locale-dependent decimal formatting (e.g. "1,5") before
    // interpolating into raw SQL.
    $safeRuntime = number_format($runtime, 6, '.', '');

    $this->db()->table('horizon_metrics')->upsert(
        [[
            'type'       => $type,
            'slug'       => $slug,
            'throughput' => 1,
            'runtime'    => $runtime,
            'updated_at' => CarbonImmutable::now(),
        ]],
        ['type', 'slug'],
        [
            'throughput' => DB::raw('horizon_metrics.throughput + 1'),
            'runtime'    => DB::raw("(horizon_metrics.runtime * horizon_metrics.throughput + {$safeRuntime}) / (horizon_metrics.throughput + 1)"),
            'updated_at' => CarbonImmutable::now(),
        ]
    );
}
```

Works on MySQL and PG. Single statement, no PHP-side serialization.

---

## 3. Monitored-tags N+1 on every push

`DatabaseTagRepository::monitored()` calls `monitoring()`, which selects the full `horizon_monitored_tags` table. That fires on **every** job dispatch. Cache it per repository instance:

```php
class DatabaseTagRepository implements TagRepository
{
    private ?array $monitoredTagsCache = null;

    public function monitoring(): array
    {
        return $this->monitoredTagsCache ??= $this->db()
            ->table('horizon_monitored_tags')
            ->pluck('tag')
            ->all();
    }

    public function monitor($tag): void
    {
        $this->db()->table('horizon_monitored_tags')->upsert(
            [['tag' => $tag]], ['tag'], []
        );
        $this->monitoredTagsCache = null;
    }

    public function stopMonitoring($tag): void
    {
        $this->db()->table('horizon_monitored_tags')->where('tag', $tag)->delete();
        $this->monitoredTagsCache = null;
    }

    public function forget($tag): void
    {
        // ...
        $this->monitoredTagsCache = null;
    }
}
```

The repo is a singleton, so per-request the cache stays hot. This was the single biggest perf regression we measured vs Redis.

---

## 4. `migrated()` should use a bulk update

```php
// PR
foreach ($payloads as $payload) {
    $this->table()->where('id', $payload->id())->update([...]);
}
```

```php
// Bulk
$ids = $payloads->map(fn ($p) => $p->id())->all();

$this->table()->whereIn('id', $ids)->update([
    'status' => 'pending',
    'delay' => 0,
    'updated_at' => $this->now(),
]);
```

One round trip instead of N.

---

## 5. Microsecond precision in timestamps

The PR stores timestamps as `double(16, 6)` Unix floats. That works, but two things bit us:

1. We hit a bug where `reserved_at` round-tripped through PHP as `(float)` and lost sub-second precision, making `runtime` show as full Unix timestamps in the UI ([fix](https://github.com/palauaandsons/horizon-postgresql/commit/574c65f)).
2. It drops time-zone semantics and makes ad-hoc SQL ("show me jobs from yesterday") much harder to read.

We switched to native `timestamp(6)` columns and convert to Unix float only at read time, with a string-formatting helper that doesn't lose precision:

```php
// migration
$table->timestamp('reserved_at', 6)->nullable();
$table->timestamp('failed_at', 6)->nullable();
$table->timestamp('completed_at', 6)->nullable();
$table->timestamp('created_at', 6);
$table->timestamp('updated_at', 6);
```

```php
private function now(): string
{
    return CarbonImmutable::now()->format('Y-m-d H:i:s.u');
}

// IMPORTANT: do not (float)-cast — that's how we lost the microseconds.
private function toUnixFloat(string $timestamp): string
{
    $dt = CarbonImmutable::parse($timestamp);

    return sprintf('%d.%06d', $dt->timestamp, $dt->micro);
}
```

If you stay on `double` columns, at least make sure no code path round-trips through `(float)` — `sprintf('%d.%06d', ...)` against the parsed value is safe.

---

## 6. Postgres-only: partial indexes on hot status filters

The PR's plain `status` index gets bloated fast: completed/deleted tuples stay in the index until vacuum, so a dashboard scan for "pending" walks past hundreds of thousands of dead entries. PG's partial indexes solve this — each index only contains the rows it cares about.

There's no Blueprint helper for partial indexes, so we drop to raw `CREATE INDEX` statements. Do this **only** for `pgsql`; MySQL ignores the `WHERE` clause.

### Migration

```php
<?php

use Illuminate\Database\Migrations\Migration;
use Illuminate\Database\Schema\Blueprint;
use Illuminate\Support\Facades\Schema;

return new class extends Migration
{
    public function getConnection(): ?string
    {
        return config('horizon.database.connection');
    }

    public function up(): void
    {
        Schema::create('horizon_jobs', function (Blueprint $table): void {
            $table->string('id')->primary();
            $table->string('connection');
            $table->string('queue');
            $table->string('name');
            $table->string('status', 20)->default('pending');
            $table->boolean('monitored')->default(false);
            $table->longText('payload');
            $table->longText('exception')->nullable();
            $table->longText('context')->nullable();
            $table->longText('retried_by')->nullable();
            $table->timestamp('reserved_at', 6)->nullable();
            $table->timestamp('failed_at', 6)->nullable();
            $table->timestamp('completed_at', 6)->nullable();
            $table->unsignedInteger('delay')->default(0);
            $table->timestamp('created_at', 6);
            $table->timestamp('updated_at', 6);

            // Cross-driver baseline indexes (work on PG, MySQL, SQLite).
            $table->index(['queue', 'status']);
            $table->index(['status', 'created_at']);
        });

        $connection = Schema::getConnection();

        // PG-only: status-specific partial indexes. These keep dead tuples
        // from other statuses out of each index, so dashboard scans for
        // "pending" or "failed" don't walk past completed-then-deleted rows.
        if ($connection->getDriverName() === 'pgsql') {
            $connection->statement("CREATE INDEX horizon_jobs_pending_created_at_index    ON horizon_jobs (created_at)   WHERE status = 'pending'");
            $connection->statement("CREATE INDEX horizon_jobs_reserved_created_at_index   ON horizon_jobs (created_at)   WHERE status = 'reserved'");
            $connection->statement("CREATE INDEX horizon_jobs_completed_completed_at_idx  ON horizon_jobs (completed_at) WHERE status = 'completed'");
            $connection->statement("CREATE INDEX horizon_jobs_silenced_completed_at_idx   ON horizon_jobs (completed_at) WHERE status = 'silenced'");
            $connection->statement("CREATE INDEX horizon_jobs_failed_failed_at_index      ON horizon_jobs (failed_at)    WHERE status = 'failed'");
            $connection->statement('CREATE INDEX horizon_jobs_monitored_updated_at_index  ON horizon_jobs (updated_at)   WHERE monitored = true');
        }
    }

    public function down(): void
    {
        Schema::dropIfExists('horizon_jobs');
    }
};
```

MySQL doesn't have partial indexes; the composite `(status, created_at)` index is a reasonable fallback and is already in the baseline above.

---

## 7. Postgres-only: per-table autovacuum overrides

The job/tag/metrics tables are extremely high-churn — one row goes insert → update (reserved) → update (completed) → delete in seconds. PG's defaults (`autovacuum_vacuum_scale_factor = 0.2`) won't vacuum until 20% of the table is dead. On a 1M-row jobs table that's 200k dead tuples — at which point queries are already slow and the index bloat is permanent until a manual `VACUUM FULL`.

The fix is per-table autovacuum tuning, applied once at migration time. These settings are *only* for these tables; they don't affect anything else in your DB.

### Config

`config/horizon.php`:

```php
'autovacuum' => [
    // Trigger vacuum after 50 dead tuples + 1% of table size (down from 20%).
    'vacuum_threshold'      => (int)   env('HORIZON_AUTOVACUUM_VACUUM_THRESHOLD', 50),
    'vacuum_scale_factor'   => (float) env('HORIZON_AUTOVACUUM_VACUUM_SCALE_FACTOR', 0.01),

    // No I/O throttling for these tables — they need to keep up.
    'vacuum_cost_delay'     => (int)   env('HORIZON_AUTOVACUUM_VACUUM_COST_DELAY', 0),
    'vacuum_cost_limit'     => (int)   env('HORIZON_AUTOVACUUM_VACUUM_COST_LIMIT', 10000),

    // Same aggressive trigger for ANALYZE so the planner stays accurate.
    'analyze_scale_factor'  => (float) env('HORIZON_AUTOVACUUM_ANALYZE_SCALE_FACTOR', 0.01),
],
```

### Shared base migration

Drop a small abstract migration in the package so every table can opt in with one call:

```php
<?php

namespace Laravel\Horizon\Database\Migrations;

use Illuminate\Database\Migrations\Migration as BaseMigration;
use Illuminate\Support\Facades\Schema;

abstract class Migration extends BaseMigration
{
    public function getConnection(): ?string
    {
        return config('horizon.database.connection');
    }

    /**
     * Apply aggressive per-table autovacuum settings on PostgreSQL.
     * No-op on every other driver.
     */
    protected function applyAutovacuumSettings(string $table): void
    {
        $connection = Schema::getConnection();

        if ($connection->getDriverName() !== 'pgsql') {
            return;
        }

        $settings = config('horizon.autovacuum');

        $params = collect([
            'autovacuum_vacuum_threshold'     => $settings['vacuum_threshold'],
            'autovacuum_vacuum_scale_factor'  => $settings['vacuum_scale_factor'],
            'autovacuum_vacuum_cost_delay'    => $settings['vacuum_cost_delay'],
            'autovacuum_vacuum_cost_limit'    => $settings['vacuum_cost_limit'],
            'autovacuum_analyze_scale_factor' => $settings['analyze_scale_factor'],
        ])->map(fn ($value, $key) => "{$key} = {$value}")->implode(', ');

        $connection->statement("ALTER TABLE {$table} SET ({$params})");
    }
}
```

### Using it from a table migration

```php
<?php

use Illuminate\Database\Schema\Blueprint;
use Illuminate\Support\Facades\Schema;
use Laravel\Horizon\Database\Migrations\Migration;

return new class extends Migration
{
    public function up(): void
    {
        Schema::create('horizon_jobs', function (Blueprint $table): void {
            // ... columns + cross-driver indexes from §6
        });

        $connection = Schema::getConnection();

        if ($connection->getDriverName() === 'pgsql') {
            $connection->statement("CREATE INDEX horizon_jobs_pending_created_at_index   ON horizon_jobs (created_at)   WHERE status = 'pending'");
            // ... rest of the partial indexes
        }

        $this->applyAutovacuumSettings('horizon_jobs');
    }

    public function down(): void
    {
        Schema::dropIfExists('horizon_jobs');
    }
};
```

Apply `applyAutovacuumSettings()` to at minimum: `horizon_jobs`, `horizon_tags`, `horizon_metrics`, `horizon_command_queue`. Documenting it in the readme as "if you're on PG, this is the recommended config" would already be a big win — the rest of the driver doesn't need to know.

---

## 8. Smaller things

- **`getPending()`** mixes `pending` and `reserved`. Reserved jobs are actively being worked; lumping them in the pending list is misleading on the dashboard. Worth splitting.
- **`baseSnapshotData()`** deletes the metrics row on every snapshot. Functionally fine, but it means every measured queue/job is re-inserted every minute — unnecessary churn. We `UPDATE ... SET throughput = 0, runtime = 0` instead.
- **`Lock::prune()`** runs on every `exists()` and `get()`. Cheap, but on a hot wait-time-monitor lock this is N deletes per worker per loop. Pruning only inside `get()` (before the insert attempt) is enough.

ahsankhatri:
Desperately looking for this to get merged.

georgebohnisch:
Amazing PR, what I've been waiting for! @stevebauman you're so great!

WendellAdriel:
Hey @stevebauman 
I was playing around with this idea last Friday
If you want to take a look, my implementation is on a branch called `feat/db-driver`

I didn't create a PR because I wanted to discuss with the team some things first.
Since you already opened a PR, if you want to take a look and see if something fits in your approach.

Feel free to ping me on X or Telegram as well if you want to chat about it.

indigoram89:
Very cool

adiologydev:
Can't believe it's finally happening! <3

config/horizon.php
```diff
@@ -56,6 +69,21 @@
 
     'use' => 'default',
```
stevebauman, line 70: Might need to be moved into a nested key like `redis` to fall in line with the `database` config -- though of course a breaking change:


```suggestion
    'redis' => [
        'connection' => 'default',
    ],
```

taylorotwell:
While I respect the effort and the idea, I don't personally want to be responsible for maintaining a DB version of Horizon right now. I think it could be a cool community package / fork though!

fabianpnke:
Hi @stevebauman, and everyone else who found their way to this PR,

Taylor's closing note — "I think it could be a cool community package / fork though!" — stuck with me, so I turned Steve's implementation into exactly that: fabianpnke/horizon-db-driver (https://github.com/fabianpnke/horizon-db-driver).

It doesn't fork Horizon — it requires the official laravel/horizon package and layers a second service provider on top that rebinds the container singletons (lock, repositories, command queue) to database-backed implementations after Horizon's own provider has registered them. All the actual driver logic — the repositories, the lock, the command queue, the queue connector, the migration — is Steve's work from this PR; I just adapted the wiring so it runs alongside Horizon instead of inside a fork of it.

- Package: https://packagist.org/packages/fabianpnke/horizon-db-driver
- Source: https://github.com/fabianpnke/horizon-db-driver
- Docs: https://fabianpnke.github.io/horizon-db-driver/
- First tagged release just went out today: v1.0.0

Full transparency, since I think it matters here: I'm new to maintaining an open-source Laravel package, and I used Claude for most of the implementation work — the container binding wiring, tests, CI matrix, and docs. I reviewed and tested everything before shipping it, but I'm still learning how to run an OSS project well, so please bear with me if something doesn't go smoothly early on.

Steve is credited as the original author in the README and composer.json, and I'd love for him to be involved if he's interested, whatever works for him.

But this doesn't have to be a two-person thing. If anyone reading this has been waiting for a database-backed Horizon and wants to help shape it — testing it against real workloads, filing issues, contributing fixes — I'd genuinely welcome it. This started from wanting a solution myself, not from wanting to own it solo.

PR #1765, opened 2026-05-11
Fix horizon:clear-metrics not clearing metric keys

### LINKED ISSUE #1746: horizon:clear-metrics does not properly clear metrics
### Horizon Version

5.45.4

### Laravel Version

12.55.1

### PHP Version

8.4

### Redis Driver

PhpRedis

### Redis Version

6.3.0

### Database Driver & Version

_No response_

### Description

I enabled metrics in Horizon [as per the instructions](https://laravel.com/docs/13.x/horizon#metrics) after several months in production. The first node on the chart showed a count of over 30,000 which obviously distorted the scale of the chart massive.

So I ran `php artisan horizon:clear-metrics` which cleared the contents of the pages `/horizon/metrics/jobs` and `/horizon/metrics/queues` 

But when a new job came in the metrics came back, the new chart starts at 30,000 again.

Also after clearing metrics, and before any new job has run the page at `/horizon/metrics/jobs/App\Path\To\Some\Job` still has a chart with data even though it's no longer listed at `/horizon/metrics/jobs`

### Steps To Reproduce

Run `php artisan horizon:clear-metrics`

Metrics are not properly cleared.

ChihebBENCHEIKH1:
The previous implementation used SCAN with a manually prefixed match pattern. When PhpRedis is configured with Redis::OPT_SCAN = SCAN_PREFIX, the connection auto-prepends the prefix to scan patterns, so manually adding it produced a double-prefix (horizon:horizon:queue:*) that matched nothing and silently left every metric key in place.

Avoid SCAN entirely: Horizon already tracks every metric key in the measured_jobs and measured_queues Redis sets. Enumerate them directly and delete each metric hash and its snapshot zset by name. The result is deterministic, prefix-agnostic, and also clears the per-job snapshot charts (/horizon/metrics/jobs/<JobName>) that previously lingered after clear-metrics.

Fixes #1746

### DIFF
--- a/src/Repositories/RedisMetricsRepository.php
+++ b/src/Repositories/RedisMetricsRepository.php
@@ -4,7 +4,6 @@ namespace Laravel\Horizon\Repositories;
 
 use Carbon\CarbonImmutable;
 use Illuminate\Contracts\Redis\Factory as RedisFactory;
-use Illuminate\Support\Str;
 use Laravel\Horizon\Contracts\MetricsRepository;
 use Laravel\Horizon\Lock;
 use Laravel\Horizon\LuaScripts;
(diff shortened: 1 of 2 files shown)


taylorotwell:

PR #1774, opened 2026-06-03
Drain Horizon commands atomically

### LINKED ISSUE #1739: Race condition in RedisHorizonCommandQueue::pending() can silently drop supervisor commands

ibobgunardi:
Fixes #1739.

`RedisHorizonCommandQueue::pending()` previously used `LLEN` followed by a pipelined `LRANGE` and `LTRIM`. A command pushed after the range was read but before the trim could be removed without ever being returned.

This drains the command list with a single Lua script so Redis reads and clears the current commands atomically. Commands pushed after the script runs remain in the list for the next loop.

Tests run:

- `php -l src/LuaScripts.php`
- `php -l src/RedisHorizonCommandQueue.php`
- `php -l tests/Unit/RedisHorizonCommandQueueTest.php`
- `php vendor/bin/phpunit tests/Unit/RedisHorizonCommandQueueTest.php`
- `php vendor/bin/phpstan analyse src/LuaScripts.php src/RedisHorizonCommandQueue.php tests/Unit/RedisHorizonCommandQueueTest.php --memory-limit=1G`

Not run: the existing supervisor feature command-loop tests require a Redis server on `127.0.0.1:6379`; my local environment does not have Redis running there.

### DIFF
--- a/src/LuaScripts.php
+++ b/src/LuaScripts.php
@@ -5,6 +5,26 @@ namespace Laravel\Horizon;
 class LuaScripts
 {
     /**
+     * Get the Lua script for atomically draining pending commands.
+     *
+     * KEYS[1] - The name of the command queue
+     *
+     * @return string
+     */
+    public static function pendingCommands()
+    {
+        return <<<'LUA'
+            local commands = redis.call('lrange', KEYS[1], 0, -1)
+
+            if #commands > 0 then
+                redis.call('del', KEYS[1])
+            end
+
+            return commands
+LUA;
+    }
+
+    /**
      * Update the metrics for a job.
      *
      * KEYS[1] - The name of the key being updated
--- a/src/RedisHorizonCommandQueue.php
+++ b/src/RedisHorizonCommandQueue.php
@@ -4,12 +4,9 @@ namespace Laravel\Horizon;
 
 use Illuminate\Contracts\Redis\Factory as RedisFactory;
 use Laravel\Horizon\Contracts\HorizonCommandQueue;
-use Laravel\Horizon\Repositories\UsesClusterAwarePipeline;
 
 class RedisHorizonCommandQueue implements HorizonCommandQueue
 {
-    use UsesClusterAwarePipeline;
-
     /**
      * The Redis connection instance.
      *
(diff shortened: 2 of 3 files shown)


taylorotwell:

I appreciate you taking the time to submit this; however, it appears this contribution may have been primarily AI-generated without careful human review and consideration.

We've found that AI-generated code often doesn't align well with Laravel's conventions, architectural decisions, and the specific context of what we're trying to accomplish with the framework. Quality contributions require thoughtful human insight into the codebase.

If you're interested in contributing to Laravel, I'd encourage you to familiarize yourself with the existing codebase, engage with the community, and submit PRs that reflect your own understanding and careful consideration of the problem you're solving.

PR #1793, opened 2026-07-16
[5.x] Add job lookup by ID

frndchagas:
## Summary

This adds a `Find Job by ID` field to the pending, completed, and silenced job screens.

Submitting an ID uses Horizon's existing `/api/jobs/{id}` endpoint and opens the matching job preview. Failed jobs are directed to the failed job preview, pending or reserved jobs use the pending route, and completed jobs use their stored silenced state to select the completed or silenced route.

If the job is no longer retained, Horizon explains that it may have expired according to the configured trim period.

## Why

Finding a specific retained job currently requires paging through the job tables or manually constructing its detail URL. This becomes impractical when an application processes a large number of jobs.

The lookup is deliberately limited to an exact job ID. Horizon already stores job hashes by ID, so this exposes the existing constant-time lookup without adding Redis indexes, background maintenance, storage overhead, configuration, or dependencies.

## Trimmed jobs

#1669 reverted direct links to completed job details because a job may be trimmed before the link is followed. This lookup checks the existing job endpoint before navigating. If the record has already been trimmed, the user stays on the list and sees a trim-aware message instead of opening an empty preview.

## Manual browser validation

- Pending ID opens the pending job preview.
- Completed ID opens the completed job preview.
- Silenced ID searched from the completed screen opens the silenced job preview.
- Completed ID searched from the silenced screen opens the completed job preview.
- Failed ID opens the failed job preview.
- Missing or trimmed ID stays on the current list and shows the trim-aware message.
- A fresh valid lookup completed without console errors or uncaught page errors.
- The search form, icon button, and decorative SVG expose the expected accessibility semantics.

### Job ID lookup

Find Job by ID field on the pending jobs screen

### Missing or trimmed job

Trim-aware message for a missing job

## Validation

- `npm run build -- --outDir /tmp/laravel-horizon-job-search-dist-20260716-routing-fix`
- `composer test` — 186 tests, 418 assertions
- `composer lint`
- `git diff --check`

### DIFF
--- a/resources/js/screens/recentJobs/index.vue
+++ b/resources/js/screens/recentJobs/index.vue
@@ -95,7 +95,7 @@
                                 jobId: response.data.id,
                                 type: ['pending', 'reserved'].includes(response.data.status)
                                     ? 'pending'
-                                    : (this.$route.params.type === 'silenced' ? 'silenced' : 'completed'),
+                                    : (response.data.payload?.silenced ? 'silenced' : 'completed'),
                             },
                         });
                     })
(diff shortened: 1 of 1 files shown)


taylorotwell:

PR #1800, opened 2026-07-25
[5.x] Fix `horizon:clear-metrics` deleting nothing on phpredis 6.x

### LINKED ISSUE #1799: horizon:clear-metrics` silently clears nothing on phpredis 6.x (SCAN starts with integer cursor `0`)
Horizon 5.48.1
Laravel 13.21.1
PHP 8.4.22
Redis Driver: PhpRedis
Redis 7.x
DB: MySQL 8.0.34

`php artisan horizon:clear-metrics` prints `Metrics cleared successfully.` but does **not** delete any metric keys. As a result `measured_jobs`, `snapshot:job:*`, `job:*`, `queue:*` accumulate without ever being cleared. On a busy app this grows unbounded (in our case ~570k keys / ~12 GB on the queue Redis) with no working way to clear them.

The cause is in `RedisMetricsRepository::clear()`:

```php
foreach (['queue:*', 'job:*', 'snapshot:*'] as $pattern) {
    $cursor = null;

    do {
        $scanResult = $this->connection()->scan(
            $cursor ?? 0, ['match' => $this->snapshotPatternToMatch($pattern)]
        );

        if (! is_array($scanResult)) {
            break;
        }

        [$cursor, $keys] = $scanResult;
        // ...
    } while ($cursor > 0);
}
```

On the first iteration `$cursor` is `null`, so `$cursor ?? 0` passes the **integer `0`** to `Illuminate\Redis\Connections\PhpRedisConnection::scan()`. With **phpredis 6.x the SCAN iterator must start as `null`** — passing integer `0` makes `\Redis::scan()` return `false` (it treats `0` as "iteration finished"). `PhpRedisConnection::scan()` then returns `false`, `if (! is_array($scanResult)) break;` fires on iteration #1, and the loop exits having deleted nothing. The `forget()` calls for `measured_jobs`/`measured_queues` do run, but the per-job/queue and snapshot keys are never scanned, so metrics are never actually cleared.

### Steps To Reproduce

1. `'client' => 'phpredis'`, phpredis 6.3.0.
2. Process a handful of jobs so metrics exist (`measured_jobs` + `snapshot:job:*` populated), or seed a key manually.
3. `php artisan horizon:clear-metrics` → prints `Metrics cleared successfully.`
4. The keys are still there — e.g. `redis-cli -n <db> --scan --pattern '<horizon-prefix>snapshot:job:*' | wc -l` is unchanged.

Minimal proof of the underlying SCAN behavior (phpredis 6.3.0), reproducing exactly what `clear()` does on its first iteration:

```php
use Illuminate\Support\Facades\Redis;

Redis::connection()->set('scan-demo:1', 'x');

$cursor = 0;                                            // == $cursor ?? 0 on the first clear() iteration
var_dump(Redis::connection()->scan($cursor, ['match' => 'scan-demo:*']));
// bool(false)  ← aborts clear()'s do/while on iteration #1

$cursor = null;                                         // works
var_dump(Redis::connection()->scan($cursor, ['match' => 'scan-demo:*']));
// array(2) { [0] => <next cursor>, [1] => ['scan-demo:1'] }
```

### Expected behaviour

`horizon:clear-metrics` deletes all metric keys.

### Suggested fix

Initialize/pass the SCAN cursor as `null` on the first call instead of coalescing to `0` (keep passing the returned cursor on subsequent iterations). Passing `null` returns matches correctly even in the default (non-`SCAN_RETRY`) option mode; the integer-`0` seed is the only thing that breaks it on phpredis 6.x.

lazerg:
On phpredis 6.x, `clear()` seeds the SCAN cursor with integer `0` (`$cursor ?? 0`), which phpredis treats as a finished iteration and returns `false`. The `do/while` then breaks on the first pass, so `horizon:clear-metrics` reports success but never deletes the per-job/queue or snapshot keys, letting them grow unbounded. Passing the cursor as `null` on the first call scans and clears the keys correctly.

Fixes #1799

### DIFF
--- a/src/Repositories/RedisMetricsRepository.php
+++ b/src/Repositories/RedisMetricsRepository.php
@@ -402,7 +402,7 @@ class RedisMetricsRepository implements MetricsRepository
 
             do {
                 $scanResult = $this->connection()->scan(
-                    $cursor ?? 0, ['match' => $this->snapshotPatternToMatch($pattern)]
+                    $cursor, ['match' => $this->snapshotPatternToMatch($pattern)]
                 );
 
                 if (! is_array($scanResult)) {
--- a/tests/Unit/RedisMetricsRepositoryTest.php
+++ b/tests/Unit/RedisMetricsRepositoryTest.php
@@ -96,7 +96,9 @@ class RedisMetricsRepositoryTest extends UnitTest
         }
 
         foreach ($patterns as $pattern) {
-            $connection->shouldReceive('scan')->once()->with(0, ['match' => $pattern])->andReturn([0, []]);
+            $connection->shouldReceive('scan')->once()->with(
+                Mockery::on(fn ($cursor) => $cursor === null), ['match' => $pattern]
+            )->andReturn([0, []]);
         }
 
         return $connection;


taylorotwell:
Fixed on another PR.

PR #1806, opened 2026-07-31
Refresh Horizon's user interface

nckrtl:
This PR is a response to [the request](https://x.com/taylorotwell/status/2081020119623344386) to port my Horizon redesign into this project.

I prepared two approaches. This is the merge-oriented option for Horizon 5.x. It keeps Vue, Bootstrap, Sass, the existing API routes, and support for Laravel 9 through 13.

A separate [Inertia PR](https://github.com/laravel/horizon/pull/1807) implements the same redesign with Inertia v3, React, Tailwind CSS, and shadcn. It provides the same Horizon functionality, but changes the frontend stack and raises the minimum supported Laravel and PHP versions. For that reason, it may be a better fit for Horizon 6.

For future UI development, that approach is also worth considering because shadcn provides common interface primitives such as dropdown menus and dialogs. With the existing stack, those would need to be built in Horizon or added through another UI package.

## What this changes

- Refreshes the shell, sidebar, status indicator, and light and dark styles
- Updates dashboard job counts, pending states, batch progress, and workload statistics
- Refreshes monitoring, metrics, batches, job lists, and job details
- Improves charts, stack traces, empty states, search, and status labels
- Adds capability-gated queue pause and resume controls through Laravel's queue APIs. These may be better exposed as opt-in because they give Horizon users more operational control.

## Screenshots

<details>
<summary>View page screenshots</summary>

### Dashboard

Dashboard

### Monitoring

Monitoring

### Metrics

Metrics

### Batches

Batches

### Pending Jobs

Pending Jobs

### Completed Jobs

Completed Jobs

### Silenced Jobs

Silenced Jobs

### Failed Jobs

Failed Jobs

### Batch details

Batch details

### Job details

Job details

</details>

PR #1807, opened 2026-07-31
Rebuild Horizon's user interface with Inertia and React

nckrtl:
This PR is a response to [the request](https://x.com/taylorotwell/status/2081020119623344386) to port my Horizon redesign into this project.

I prepared two approaches. The [classic PR](https://github.com/laravel/horizon/pull/1806) is the merge-oriented option for Horizon 5.x. It keeps Vue, Bootstrap, Sass, the existing API routes, and support for Laravel 9 through 13.

This future-oriented PR implements the same redesign with Inertia v3, React, Tailwind CSS, and shadcn. It provides the same Horizon functionality as the classic PR, but changes the frontend stack and raises the minimum supported Laravel and PHP versions. For that reason, it may be a better fit for Horizon 6.

There is a [live demo](https://horizon.nckrtl.com/horizon) based on my [Horizon New Dawn concept](https://horizon.nckrtl.com). It uses the same stack, but also explores controls beyond this PR, including a [queue detail page](https://horizon.nckrtl.com/horizon/queues/default?tab=pending), terminating Horizon instances, and starting or stopping individual supervisors. Those controls would need separate agreement before becoming part of Horizon.

The new stack also makes it easier to support mobile layouts, which the live demo already demonstrates.

## What this changes

- Raises the minimum requirements to Laravel 11.35 and PHP 8.2 because those are the minimum versions supported by Inertia v3.
- Preserves every existing Horizon API route, including its method, name, controller, constraints, and middleware.
- Adds navigation counts to `/api/stats`, and pause and throughput metadata to `/api/workload`. No existing response field is removed or renamed.
- Adds queue pause and resume through Laravel's queue APIs. This may be better exposed as an opt-in feature because it gives Horizon users more operational control.
- Publishes the frontend assets through Horizon's install and update flow.


ao-jhelmich:
Is this still on the wishlist?

PR #1813, opened 2026-08-27
[5.x] Allow minProcesses to be set to zero

curruwilla:
A supervisor that works many queues currently pays a floor of one worker per queue, even when every one of those queues is idle. On a supervisor with a dozen queues that is a dozen resident PHP processes doing nothing.

Allowing `minProcesses` to be `0` lets the auto-scaler drain a pool all the way down, and start workers again as soon as jobs are pushed:

- `ProvisioningPlan` now rejects only negative values.
- `horizon:supervisor` no longer boots `floor((min + max) / 2)` workers when `minProcesses` is `0`, so idle queues start with no processes at all instead of being scaled down again a few seconds later.

Behaviour is unchanged for any supervisor with `minProcesses >= 1`.

### DIFF
--- a/src/Console/SupervisorCommand.php
+++ b/src/Console/SupervisorCommand.php
@@ -96,7 +96,13 @@ class SupervisorCommand extends Command
 
         $supervisor->working = ! $this->option('paused');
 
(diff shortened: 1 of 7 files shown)


taylorotwell:

PR #1816, opened 2026-09-05
Allow disabling automatic model tagging

laravel-o11y:
Jobs that do not define a tags() method are tagged automatically with every Eloquent model they carry. Those tags feed the monitoring tab and the tag filter on the failed jobs screen, but applications that use neither still pay for them: the tags are stored in every copy of the payload, and each failure writes a failed:{tag} sorted set per tag, which for model tags means a Redis key per model instance held for trim.failed minutes.

The new auto_tags option gates only the model-derived fallback, so a tags() method declared on a job, mailable, notification, event or listener keeps working, as do silenced_tags and monitoring for the tags an application declares itself. The option defaults to true, leaving current behavior unchanged.

### DIFF
--- a/config/horizon.php
+++ b/config/horizon.php
@@ -141,6 +141,23 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Automatic Job Tagging
+    |--------------------------------------------------------------------------
+    |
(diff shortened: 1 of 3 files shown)


taylorotwell:

